PR #27, opened 2015-07-15
Use upstream 'overlay' filesystem

jprvita:
Ubuntu's Linux 4.1 (a requirement for eos 2.4.0) does not provide the
'overlayfs' filesystem we have been using before the upgrade, but a
compatibility layer implemented on top of the upstream 'overlay'
filesystem (endlessm/linux@a61b539). Since we're moving to a new
codebase in any case, let's just use the upstream version directly.

For both 'overlay' and 'overlayfs' filesystems there is a 'workdir'
mount option that needs to be passed to rw mounts.

[endlessm/eos-shell#4518]

### DIFF
--- a/endless-mount-dirs.service
+++ b/endless-mount-dirs.service
@@ -1,5 +1,5 @@
-# Ensure /var/endless and /var/endless-extra exist prior to mounting
-# /var/endless-extra or the overlay at /endless. This is required for
+# Ensure /var/endless, /var/endless-extra and /var/workdir exist prior to
+# mounting /var/endless-extra or the overlay at /endless. This is required for
 # single disk unit upgrades where /var/endless-extra doesn't exist. It's
 # also good practice to ensure the mount points exist like systemd would
 # do if /endless was a non-overlayfs mount.
@@ -25,9 +25,10 @@ RequiresMountsFor=/var
 # Only run if the directories don't exist
 ConditionPathIsDirectory=|!/var/endless
 ConditionPathIsDirectory=|!/var/endless-extra
+ConditionPathIsDirectory=|!/var/workdir
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=/bin/mkdir -p /var/endless /var/endless-extra
+ExecStart=/bin/mkdir -p /var/endless /var/endless-extra /var/workdir
 ExecStart=-/bin/chown app-manager:app-manager /var/endless /var/endless-extra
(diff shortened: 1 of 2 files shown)


endless-mount-dirs.service
```diff
@@ -25,9 +25,10 @@ RequiresMountsFor=/var
...
 ConditionPathIsDirectory=|!/var/endless
 ConditionPathIsDirectory=|!/var/endless-extra
+ConditionPathIsDirectory=|!/var/workdir
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=/bin/mkdir -p /var/endless /var/endless-extra
+ExecStart=/bin/mkdir -p /var/endless /var/endless-extra /var/workdir
 ExecStart=-/bin/chown app-manager:app-manager /var/endless /var/endless-extra
```
dbnicholson: I'm not sure it's really required, but please update the chown command, too.

dbnicholson:
Can we use /var/endless-workdir instead of /var/workdir? /var/workdir is bit generic for me.

endless-mount-dirs.service
jprvita: In my first test I checked if the dir would be automatically created, and it is not. Also from Documentation/filesystems/overlayfs.txt

> The "workdir" needs to be an empty directory on the same filesystem
> as upperdir.

jprvita:
Ok, I'll updated it to /var/endless-workdir

endless-mount-dirs.service
jprvita: Ow, wait, are you referring to manually creating the directory, or to changing it's ownership to "app-manager". Regarding the later, I've asked myself the same question, but since I was able to write in the overlay (/endless) as the app-manager user, I don't think it is necessary.

dbnicholson: Thinking again, I don't think the workdir ownership should be changed. It's entirely for the kernel's internal use, so I don't think we should allow the app-manager user to accidentally write files into it. So, just changing the workdir name should be enough for this.

jprvita: I agree, lets just keep root as the owner.

dbnicholson:
Just FYI, you can keep using the same pull request if you like after revising. No need to open a new PR unless you really want to.

jprvita:
Hum, didn't know that. The only use of GitHub I did in the past was keeping personal repositories, mainly for backup purposes, with all the review process done in mailing lists.

So I guess the PR will keep tracking the branch name, and as long as I keep pushing to the same branch, the PR will be automatically updated. Sounds nice, I'll stop closing my PRs when I'm fixing comments..

dbnicholson:
Yeah, then we can keep viewing the comments from previous reviews to make sure they were addressed.

PR #29, opened 2015-07-18
Migrate to GPT/UEFI

dsd:
https://github.com/endlessm/eos-shell/issues/5331

### DIFF
--- a/boot.mount
+++ b/boot.mount
@@ -1,0 +1,12 @@
+[Unit]
+Before=local-fs.target
+RequiresOverridable=systemd-fsck@dev-disk-by\x2dlabel-ostree\x2dboot.service
+After=systemd-fsck@dev-disk-by\x2dlabel-ostree\x2dboot.service
+
+[Mount]
+What=/dev/disk/by-label/ostree-boot
+Where=/boot
+Type=ext2
+
+[Install]
+WantedBy=dev-disk-by\x2dlabel-ostree\x2dboot.device
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -39,14 +39,28 @@ if [ -z ${root_part} ]; then
   exit 0
 fi
 
+get_last_char() {
+	ret=$1
+	while [ "${#ret}" != "1" ]; do
+		ret=${ret#?}
+	done
+	echo $ret
+}
+
+# We have to work with a variety of setups here:
+# - Either SCSI or MMC block devices with different naming conventions
+# - A different partition numbering scheme as not all configurations have
+#   a ESP and a BIOS boot partition.
+partno=$(get_last_char ${root_part})
+swap_partno=$((partno + 1))
 case ${root_part} in
-  /dev/mmcblk?p2)
-    root_disk=${root_part%p2}
-    swap_part=${root_part%?}3
+  /dev/mmcblk?p?)
+    root_disk=${root_part%p?}
+    swap_part=${root_disk}p${swap_partno}
     ;;
-  /dev/sd?2)
-    root_disk=${root_part%2}
-    swap_part=${root_part%?}3
+  /dev/sd??)
+    root_disk=${root_part%?}
+    swap_part=${root_disk}${swap_partno}
     ;;
 esac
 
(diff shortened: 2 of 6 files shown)


boot.mount
```diff
@@ -0,0 +1,12 @@
...
+RequiresOverridable=systemd-fsck@dev-disk-by\x2dlabel-ostree\x2dboot.service
+After=systemd-fsck@dev-disk-by\x2dlabel-ostree\x2dboot.service
+
+[Mount]
+What=/dev/disk/by-label/ostree-boot
+Where=/boot
+Type=ext2
+
+[Install]
+WantedBy=dev-disk-by\x2dlabel-ostree\x2dboot.device
```
dbnicholson: I don't actually think this is required. I believe that systemd will take the `What`, convert it to a device unit, and add it `BindsTo` and `WantedBy`. Can you try with and without? You should just be able to do `systemctl daemon-reload && systemctl show boot.mount` to see differences.

dracut/repartition/endless-repartition.sh
```diff
@@ -39,14 +36,28 @@ if [ -z ${root_part} ]; then
...
+		ret=${ret#?}
+	done
+	echo $ret
+}
+
+# We have to work with a variety of setups here:
+# - Either SCSI or MMC block devices with different naming conventions
+# - A different partition numbering scheme as not all configurations have
+#   a ESP and a BIOS boot partition.
+partno=$(get_last_char ${root_part})
```
dbnicholson: If this is bash, you can just use the offset expansion with a negative offset of 1 - `${root_part: -1}`. If this isn't bash, the above function is fine.

dracut/repartition/endless-repartition.sh
```diff
@@ -55,17 +66,9 @@ if [ -z "${root_disk}" ]; then
...
-# This is implicit below: a GPT setup will have a protective MBR with a single
-# partition entry, it will not have the marker partition.
-# There is still a chance that a secondary GPT header can be found on the last
-# sector of the disk, so we must also use "--force" so that sfdisk doesn't
-# bail out in such situations.
-
 # Check for our magic "this is Endless" marker
-marker=$(sfdisk -n --force --print-id $root_disk 4)
-if [ "$marker" != "dd" ]; then
+marker=$(sfdisk --force --part-attrs $root_disk $partno)
```
dbnicholson: Was the dropping of `-n` intentional? I would think writing to the device wouldn't be necessary to read the attrs, but maybe I'm misunderstanding?

dracut/repartition/endless-repartition.sh
dsd: Yes, we're in dash.

dracut/repartition/endless-repartition.sh
dsd: The -n was only added to work around a previous bug where it would open the device for writing when it only needs to read. That has now been fixed, so this change was intentional.

dracut/repartition/endless-repartition.sh
```diff
@@ -95,26 +104,61 @@ if [ $added_space -gt 209715200 ]; then
...
-  udevadm settle
+  parts=$(echo "$parts" | sed -e "$ s/size=[0-9\t ]*,/size=$new_size,/")
 fi
 
 if [ -n "$swap_start" ]; then
   # Create swap partition
   echo "Create swap partition at $swap_start"
-  echo "$swap_start,+,S," | sfdisk --force --no-reread -N3 $root_disk
-  echo "sfdisk returned $?"
+  parts="$parts\nstart=$swap_start, type=0657FD6D-A4AB-43C4-84E5-0933C84B4F4F"
```
dbnicholson: Using `\n` in the assignment seems to work in dash but not in bash and I don't think it's specified behavior. I think it's safer to just continue on the next line:
```
parts="$parts
start=$swap_start, ..."
```
That's supported in any shell.

dbnicholson:
Looks good with a couple minor comments. One thing I will note that will change is with the /boot mount. Previously, it would have been wanted by local-fs.target and blocked there until the device appeared. Now, systemd will continue past local-fs.target if the device hasn't shown up yet, I think.

boot.mount
dsd: There aren't any differences there in the `show` output, but this does seem to be required.
I can't get the mount unit to automatically activate unless it has that [Install] section and is enabled via `systemctl enable`.

PR #30, opened 2015-07-20
Install udev rules for Meson video decoder

dsd:
Chromium looks for the V4L2 video decoder at /dev/mfc-dec. Add
an appropriate symlink for the Meson platform, plus a more appropriate
"meson-vdec" link.

[endlessm/eos-shell#5226]

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -16,6 +16,9 @@ dist_systemdunit_DATA = \
 systemdunit_DATA = \
 	dev-disk-by\\x2dlabel-eos\\x2dswap.swap \
 	var-endless\\x2dextra.mount
+
+udevrulesdir=/lib/udev/rules.d
+dist_udevrules_DATA = 50-meson-vdec.rules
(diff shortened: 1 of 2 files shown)


Makefile.am
```diff
@@ -16,6 +16,9 @@ dist_systemdunit_DATA = \
 systemdunit_DATA = \
 	dev-disk-by\\x2dlabel-eos\\x2dswap.swap \
 	var-endless\\x2dextra.mount
+
+udevrulesdir=/lib/udev/rules.d
```
dbnicholson: Unfortunately, this will break distcheck since it needs the install to all be under `$prefix`. I'll add a commit to handle that.

dsd: I don't think you need to, this is already behind a configure arg which is not enabled for distcheck.

dbnicholson: Oh, it's under the "no systemd" case, right.

PR #32, opened 2015-07-25
Fixes for app mounts

dbnicholson:
Some fixes to make app mounts of /var/endless and /var/endless-extra work better.

[endlessm/eos-shell#5414]


magcius:
Hey man, you're supposed to be on vacation, but I won't turn it down.

dbnicholson:
After today (really the next 30 mins) I will be officially unplugged!

PR #35, opened 2015-12-14
Enable ZRAM for platforms with <1GB RAM

carlocaione:
[endlessm/eos-shell#6162]

### DIFF
--- a/eos-enable-zram
+++ b/eos-enable-zram
@@ -1,0 +1,29 @@
+#!/bin/bash -e
+# Copyright (C) 2015 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+if [ $# -lt 1 ]; then
+	echo "Error: no size supplied" >&2
+	exit 1
+fi
+
+if [ "${1:-0}" -lt 0 ]; then
+	echo "Error: invalid size $1" >&2
+	exit 1
+fi
+
+th_kb=$((1*1024*1024)) # 1GB
+ram_size_kb=$(awk '/MemTotal/{print $2}' /proc/meminfo)
+
+if [ "$ram_size_kb" -le "$th_kb" ]; then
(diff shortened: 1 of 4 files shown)


eos-enable-zram
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env sh
```
dbnicholson: Below you're using some bash features, so I suggest just having this be `#!/bin/bash`. It might also be a good idea to add `-e` to exit on errors. E.g., if you can't modprobe zram, then you probably shouldn't keep going.

eos-enable-zram
```diff
@@ -0,0 +1,22 @@
...
+# Copyright (C) 2015 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+[ $# -eq 1 ] || exit -1
+[ "${1:-0}" -ge 0 ] || exit -1
```
dbnicholson: Can you add error messages for these so we find something in the journal if things get screwed up? E.g.,
```
if [ $# -lt 1 ]; then
  echo "error: no size supplied" >&2
  exit 1
fi
if [ "${1:-0}" -lt 0 ]; then
  echo "error: invalid size $1" >&2
  exit 1
fi

File: eos-enable-zram.service [hunk 3]
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=swap with zram
+DefaultDependencies=no
```
dbnicholson: Since you have `DefaultDependencies=no`, I think you'll need to add a couple more dependencies to ensure systemd does the right thing. For instance, you need to run after `ostree-remount.service` so that `modprobe zram` can find the actual module.

Actually, you may not need `DefaultDependencies=no`. Our other swap and mount units don't have that. If you leave it out, then you don't have to worry about any of this.

PR #37, opened 2016-01-12
Fix extra disk resizing race

dsd:
https://github.com/endlessm/eos-shell/issues/6211


dbnicholson:
Can you also change the `ConditionPathExists` setting in the service file so we don't run at all after successful resizing?

dsd:
That's already done, or did you have something else in mind?

dbnicholson:
Sorry, somehow I missed that chunk in the 2nd commit. Merging now.

PR #39, opened 2016-04-08
T11157: Fix newer systemd un-mounting /boot

jprvita:
https://phabricator.endlessm.com/T11157

### DIFF
--- a/boot.mount
+++ b/boot.mount
@@ -1,12 +1,11 @@
 [Unit]
 Before=local-fs.target
-RequiresOverridable=systemd-fsck@dev-disk-by\x2dlabel-ostree\x2dboot.service
-After=systemd-fsck@dev-disk-by\x2dlabel-ostree\x2dboot.service
 
 [Mount]
-What=/dev/disk/by-label/ostree-boot
+What=/sysroot/boot
 Where=/boot
-Type=ext2
+Type=none
+Options=bind
 
 [Install]
-WantedBy=dev-disk-by\x2dlabel-ostree\x2dboot.device
+WantedBy=local-fs.target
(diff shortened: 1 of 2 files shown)


dbnicholson:
This is kinda scary, but it looks OK. I do feel like these changes are more in the territory of ostree and would be best handled by ostree specific systemd services, but I'm not sure what that would look like.

@dsd Any comments?

jprvita:
@dbnicholson after fiddling with this, I actually think the opposite: whether or not there is a separate boot partition is a configuration decision for the distro (or the user), and it is much easier to configure this through systemd (or fstab) than in a binary that lives in the initramfs. But I agree the whole thing is not exactly friendly.

Additionally, one thing I would like to explore soon (but I want to wait for a more stable master first) is to remove the "ostree" label from the root partition and let systemd automatically detect the root partiton through the gpt-generator. I did a quick experiment and it seems to work fine, so I'm hoping with that we will be able to avoid the conflict when trying to boot a USB drive flashed with EOS on a machine that has EOS flashed on internal storage. The way it works is looking at a gpt-specific label, but on the same device it has been booted from, which is passed by the bootloader if I'm not mistaken.

dbnicholson:
Mostly agree, but the /sysroot/boot to /boot is very much ostree specific.

As for the labels, I would love to see them go away. They cause lots of headaches. However, how do you boot from USB in such a setup? The problem people always hit in the office is they want to boot an endless system with an endless USB stick. Things go haywire because you have no idea which device will get the ostree and ostree-boot labels.

There's an issue open somewhere about that, and I did some experiments with swapping out the labels for UUIDs on first boot, but I never got it working quite right.

PR #41, opened 2016-04-20
endless-repartition: Get root disk from kernel command line

dbnicholson:
Rather than hardcoding the root disk as /dev/disk/by-lable/ostree, read
the root= arg from the kernel command line using dracut's library and
resolve the device from there. This allows us to specify a UUID in the
kernel command line and have the repartitioning work.

https://phabricator.endlessm.com/T11392


dbnicholson:
Closing as #43 was merged instead.

PR #43, opened 2016-04-22
repartition: detect root device from systemd

dsd:
To move away from having the root filesystem always identifed by a fixed
label, update endless-repartition.sh to query sysroot.mount for the root
device.

That introduces a complication in the service file, we can no longer
order after a fixed device such as dev-disk-by\x2dlabel-ostree.device.

To make this unit run after the root device becomes available, but before
fscking it, we would have to likely modify systemd's generators so that
the availability of the root device triggers a new target, which
endless-repartition could be a part of.

But there's no strong requirement for us to run before fsck. We can
instead run after fsck, at which point the root device is guaranteed to
be ready.

https://phabricator.endlessm.com/T11392

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -29,8 +29,15 @@
 #
 # Based on code from dracut-modules-olpc.
 
+root_part=$(systemctl show -p What sysroot.mount)
+root_part=${root_part#What=}
+if [ -z "${root_part}" ]; then
+  echo "repartition: couldn't identify root device"
+  exit 0
+fi
+
 # Identify root partition device node and parent disk
-root_part=$(readlink -f /dev/disk/by-label/ostree)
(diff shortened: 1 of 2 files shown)


dracut/repartition/endless-repartition.sh
```diff
@@ -29,8 +29,15 @@
 #
 # Based on code from dracut-modules-olpc.
 
+root_part=$(systemctl show -p What sysroot.mount)
```
dbnicholson: Clever. Does this work early in the initramfs? Does sysroot.mount get generated with the root= argument? I'm asuming you tested this.

dsd: Yes, it's tested.

sysroot.mount is a core part of the boot process, nothing new, you can see that we already reference it in the service file. It is indeed autogenerated.

PR #45, opened 2016-04-25
T11374: Add conditionals to control boot.mount execution.

jprvita:
Add conditionals to control boot.mount execution.

https://phabricator.endlessm.com/T11374


dbnicholson:
Kinda funky to be bind mounting the real device from /sysroot/boot to /boot, but I guess this is the least painful way to support ostree and converted systems.

Looks OK to me, but I'll wait to see if @dsd has any comments.

PR #46, opened 2016-04-26
T11374: Fix ordering and add conditionals to control boot.mount execution

jprvita:
https://phabricator.endlessm.com/T11374


dbnicholson:
Seems OK to me. @dsd if you don't have time to review, I think it can be merged.

PR #51, opened 2016-05-27
disable eos-swap on eosinstaller images

ramcq:
When installing on systems with EOS already installed, this unit enables the swap on the HDD and makes the install fail.


ramcq:
For https://phabricator.endlessm.com/T11757

PR #53, opened 2016-06-16
Randomize UUIDs on first boot

carlocaione:
https://phabricator.endlessm.com/T12003

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -130,7 +130,9 @@ elif [ $lastpart -gt $swap_partno ]; then
 fi
 
 # remove the last-lba line so that we fill the disk
-parts=$(echo "$parts" | sed -e '/^last-lba:/d')
+# + Randomize PARTUUIDs
+# + Randomize GPT disk UUID
+parts=$(echo "$parts" | sed -e '/^last-lba:/d;s/ uuid=.*//;/^label-id:/d')
 
 if [ $new_size -gt $part_size ]; then
   echo "Try to resize $root_part to fill $new_size sectors"
--- a/eos-firstboot
+++ b/eos-firstboot
@@ -2,7 +2,20 @@
 # Copyright (C) 2014 Endless Mobile, Inc.
 # Licensed under the GPLv2
 
-resize2fs "$(findmnt -rvnf -o SOURCE /)"
+root_part=$(findmnt -rvnf -o SOURCE /)
+
+resize2fs "${root_part}"
+
+# Randomize UUID for the root filesystem
+#   uninit_bg checksums depend on the UUID, this prohibits changes to
+#   the UUID if a checksumming filesystem is mounted. Temporarly disable
+#   the feature before changing the UUID
+tune2fs -O ^uninit_bg ${root_part}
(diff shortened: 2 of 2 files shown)


dracut/repartition/endless-repartition.sh
```diff
@@ -144,12 +144,21 @@ if [ -n "$swap_start" ]; then
 start=$swap_start, type=0657FD6D-A4AB-43C4-84E5-0933C84B4F4F"
 fi
 
+# Randomize UUIDs
+parts=$(echo "$parts" | while read line; do echo $line | sed -e "s/uuid=.*/uuid=$(uuidgen)/"; done)
```
dsd: To avoid adding uuidgen to the initramfs, try simply deleting the uuid= part of the line. Hopefully that will cause sfdisk to generate new ones itself. Same for disk label UUID

dracut/repartition/endless-repartition.sh
```diff
@@ -144,12 +144,21 @@ if [ -n "$swap_start" ]; then
...
+parts=$(echo "$parts" | sed -e "s/label-id:.*/label-id: $(uuidgen)/")
+
 echo "$parts"
 echo "$parts" | sfdisk --force --no-reread $root_disk
 ret=$?
 echo "sfdisk returned $ret"
 udevadm settle
 
+# Randomize PARTUUID for root filesystem
+tune2fs "$root_part" -U $(uuidgen)
```
dsd: lets do this on the real root to avoid adding tune2fs to the initramfs

dsd:
Sorry if my previous feedback was a bit unclear.
For the parts that can be done trivially in the initramfs, I would prefer to have them there, especially if we can avoid another change to the partition table.

We already take and manipulate the sfdisk output when rewriting the partition table there. I checked that if you delete the label-id and uuid parts from the sfdisk output, they will be reset to random values. So I believe the partition UUID and disk label UUID should be done there.

But the fs uuid change should indeed be done in the real root because it truly is an extra step. For that you can do it with `tune2fs -U random`. Please add comments to explain why udevsettle and the fs flags changes are needed.

carlocaione:
> Please add comments to explain why udevsettle and the fs flags changes are needed.

udevsettle was used just in case the change was triggering some udev event. The fs flag changes are needed when we want to change the UUID of a filesystem that is already mounted. Not sure if these are still needed using the `-U random` parameter.

carlocaione:
```
root@endless:/sysroot/home/carlo# tune2fs -U random /dev/sda3
tune2fs 1.42.12 (29-Aug-2014)
The UUID may only be changed when the filesystem is unmounted.
```

dracut/repartition/endless-repartition.sh
```diff
@@ -144,6 +144,12 @@ if [ -n "$swap_start" ]; then
 start=$swap_start, type=0657FD6D-A4AB-43C4-84E5-0933C84B4F4F"
 fi
 
+# Randomize PARTUUIDs
```
dsd: You can pass multiple -e options to sed. So please combine these 2 with the sed that deletes last-lba. That way we run 1 sed process instead of 3.

PR #54, opened 2016-06-17
T12058: Remove support for separate boot partition

jprvita:
Starting with 3.0 we don't need to support legacy setups with a separate
boot partition anymore.

https://phabricator.endlessm.com/T12058


boot.mount
```diff
@@ -1,6 +1,5 @@
 [Unit]
 ConditionKernelCommandLine=|ostree
-ConditionPathIsMountPoint=|/sysroot/boot
```
dsd: Do we still need boot.mount at all? The right mount is set up by ostree in the initramfs, and if there's no boot.mount then systemd will hopefully not touch it beyond that point.

jprvita: We don't really need it as long as the ostree binary keeps doing the bind mount in the initrd. I personally think it is more flexible to have systemd do the bind mount from the rootfs, but I don't see we changing this anytime soon. So I'll add an extra commit removing boot.mount as well.

PR #55, opened 2016-06-28
Don't ship a swap unit anymore

jprvita:
Recent systemd (229+) GPT generator automatically detects and activates
swap partitions on the same device the system has booted from, so we
don't need to ship a swap unit.

https://phabricator.endlessm.com/T11761

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -8,10 +8,3 @@ override_dh_auto_configure:
 
 override_dh_systemd_start:
 	dh_systemd_start --no-start
-
-# Enable all units here, except swap, since dh-systemd doesn't get the
-# escaping right. We enable swap manually in postinst.
-override_dh_systemd_enable:
-	dh_systemd_enable eos-firstboot.service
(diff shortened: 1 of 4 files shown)


debian/rules
```diff
@@ -9,8 +9,6 @@ override_dh_auto_configure:
 override_dh_systemd_start:
 	dh_systemd_start --no-start
 
-# Enable all units here, except swap, since dh-systemd doesn't get the
```
dsd: now that we don't have swap is it possible to remove the override_dh_systemd_enable section altogether?

jprvita: I'm not sure, I'll checit.

jprvita: If we remove the override, `dh_systemd_enable` will try to enable all units. The only units not listed there, `eos-extra-resize.service` and `var-endless\x2dextra.mount`, do not have a `[Install]` section, so I believe `dh_systemd_enable` will do the right thing and not enable them (since they can't be enabled by `systemctl`). I'll send an updated version removing the section altogether.

PR #59, opened 2016-07-22
Support booting from image file on exFAT

dsd:
Add infrastructure needed to support booting from an Endless image file
found on an exFAT partition. This is primarily implemented inside the
new eos-image dracut module.

The endless.image.device parameter should be generated by the bootloader,
either specifying the PARTUUID= of the partition that contains the Endless
image.

The endless.image.path parameter provides the path to the image file on
that filesystem.

A new generator handles these parameters and hooks up the final block
device setup to the normal systemd initramfs fsck/mount infrastructure.

The block device setup script identifies which disk extents are occupied
by the image file and creates a linear device-mapper device that
corresponds. This is easier and much better performing than going through
exfat-fuse.

The repartitioning code now supports working with the resultant
device-mapper setup. We no longer get size and start info from sysfs,
because the sysfs paths to access dm info are a bit harder to deduce,
and the 'start' attribute is not present. We now use blockdev to get
the size, and we read the partition start from the sfdisk dump.

All that should cover the case for the regular booting of an Endless
image stored on a windows filesystem.

The endless.live_boot parameter implementation additionally tends to a
special case for the combined USB live+installer disk. In this case,
the image file used for live boot needs to be left untouched and pristine
so that the installer can work with it too. When this parameter is
specified, all the block devices are created in read-only mode, no
filesystem resizing is attempted, the OS updater is disabled, and
overlayfs is used for (throw-away) writable filesystem backing.

The overlay setup had to be split in 2. Most of them need to be created
quite late during real-root boot to avoid fighting with ostree-remount,
but /etc needs to be done before we switch root.

https://phabricator.endlessm.com/T12481

### DIFF
--- a/dracut/image-boot/Makefile.am
+++ b/dracut/image-boot/Makefile.am
@@ -1,0 +1,3 @@
+dracutmoddir = $(prefix)/lib/dracut/modules.d/50eos-image-boot
+dist_dracutmod_SCRIPTS = module-setup.sh eos-image-boot-setup eos-image-boot-generator
+dist_dracutmod_DATA = eos-image-boot-setup.service eos-live-etc-setup.service
--- a/dracut/image-boot/eos-image-boot-generator
+++ b/dracut/image-boot/eos-image-boot-generator
@@ -1,0 +1,86 @@
+#!/bin/sh
+# Copyright (C) 2016 Endless Mobile, Inc.
+# Licensed under the GPLv2
+#
+# Support booting from an image file hosted on a filesystem.
+# We parse the kernel parameters requesting this functionality and configure
+# other systemd units to go about setting up the block devices and mounting
+# the target filesystem.
+
+type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
+
+GENERATOR_DIR="$1"
+
+image_device=$(getarg endless.image.device=)
+image_path=$(getarg endless.image.path=)
+
+[ -z "${image_device}" -o -z "${image_path}" ] && exit 0
+
+case "${image_device}" in
+PARTUUID=*)
+  image_device=/dev/disk/by-partuuid/${image_device#PARTUUID=}
+  ;;
+UUID=*)
+  image_device=/dev/disk/by-uuid/${image_device#UUID=}
+  ;;
+esac
+
+image_device_escaped=$(systemd-escape -p "${image_device}")
+
+# For convenience so that other scripts don't have to parse the cmdline again
+echo "${image_device}" > /var/tmp/endless-image-host
+
+# When the image device appears, run the setup
+mkdir -p ${GENERATOR_DIR}/initrd-fs.target.wants
+ln -s /lib/systemd/system/eos-image-boot-setup.service \
+	${GENERATOR_DIR}/initrd-fs.target.wants
+mkdir -p ${GENERATOR_DIR}/eos-image-boot-setup.service.d
+cat <<EOF >${GENERATOR_DIR}/eos-image-boot-setup.service.d/50-device.conf
+[Unit]
+Requires=${image_device_escaped}.device
+After=${image_device_escaped}.device
+EOF
+
+if getargbool 0 endless.live_boot; then
+  rwflag=ro
+else
+  rwflag=rw
+fi
+
+# Once the actual EOS device has appeared, use the normal infrastructure
+# to fsck and mount it.
+cat <<EOF >${GENERATOR_DIR}/sysroot.mount
+[Unit]
+Before=initrd-root-fs.target
+Requires=systemd-fsck-root.service
+After=systemd-fsck-root.service
+[Mount]
+What=/dev/mapper/endless-image3
+Where=/sysroot
+Options=$rwflag
+EOF
+
+cat <<EOF >${GENERATOR_DIR}/systemd-fsck-root.service
+[Unit]
+DefaultDependencies=no
+BindsTo=dev-mapper-endless\x2dimage3.device
+After=initrd-root-device.target local-fs-pre.target
+Before=shutdown.target
+ConditionKernelCommandLine=!endless.live_boot
+[Service]
+Type=oneshot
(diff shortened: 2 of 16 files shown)


dracut/image-boot/image-boot-generator
```diff
@@ -0,0 +1,83 @@
...
+type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
+
+GENERATOR_DIR="$1"
+
+image_device=$(getarg endless.image.device=)
+image_path=$(getarg endless.image.path=)
+
+[ -z "${image_device}" -o -z "${image_path}" ] && exit 0
+
+case "${image_device}" in
```
ramcq: We should support UUID= here as well, because we need to re-use this script for MBR booting in the Windows install case.

dracut/image-boot/eos-image-boot-setup
```diff
@@ -0,0 +1,48 @@
...
+  echo "image-boot: failed to lookup image on device"
+  exit 1
+fi
+
+offset=0
+echo "$extents" | while read extent_offset extent_size; do
+  [ -z "${extent_offset}" -o -z "${extent_size}" ] && continue
+  # Convert bytes to sectors, rounding up
+  extent_offset=$(((extent_offset + 511) / 512))
+  extent_size=$(((extent_size + 511) / 512))
```
wjt: Possibly-stupid question: what if the offset and size are not sector-aligned? If they are not, would the mapped block device not be garbage here? And in that case, would it be better to bail out rather than blindly rounding up?

dsd: The offset is guaranteed to be sector aligned by the filesystem design. Rounding up the size is probably indeed a small mistake, although that will never happen for the real images. So I think I'll take your suggestion for the next version.

dracut/image-boot/Makefile.am
```diff
@@ -0,0 +1,3 @@
+dracutmoddir = $(prefix)/lib/dracut/modules.d/50eos-image-boot
+dist_dracutmod_SCRIPTS = module-setup.sh eos-image-boot-setup image-boot-generator
```
wjt: Nitpick: why **eos-**image-boot-setup but no prefix on image-boot-generator?

dracut/repartition/endless-repartition.sh
```diff
@@ -80,11 +88,30 @@ if [ "$marker" != "GUID:55" ]; then
...
+  # a swap partition but didn't finish. Remove it to try again below.
+  parts=$(echo "$parts" | sed '$d')
+elif [ $lastpart -gt $swap_partno ]; then
+  echo "repartition: found $lastpart partitions?"
+  exit 0
+fi
+
+# calculate new partition sizes
 root_disk_name=${root_disk#/dev/}
 root_part_name=${root_part#/dev/}
```
wjt: `root_disk_name` and `root_part_name` are now unused

dracut/image-boot/image-boot-generator
wjt: I don't think this should block merging this PR: for the “live + installer USB stick” case I think we can depend on GPT – except maybe when creating the USB stick from XP?

wjt:
As far as I feel qualified to review this, it looks fine barring a few nit-picks.

dracut/image-boot/image-boot-generator
ramcq: true - we're not able to support making these sticks from XP unless we have the images totally baked ahead of time and we just DD them - I'd prefer to do that for ISOs rather than img.gz

dsd:
Thanks. I'll work on a new version tomorrow with those tweaks, plus latest round of feedback from exfat tools upstream

dracut/image-boot/eos-image-boot-generator
```diff
@@ -0,0 +1,86 @@
...
+GENERATOR_DIR="$1"
+
+image_device=$(getarg endless.image.device=)
+image_path=$(getarg endless.image.path=)
+
+[ -z "${image_device}" -o -z "${image_path}" ] && exit 0
+
+case "${image_device}" in
+PARTUUID=*)
+  image_device=/dev/disk/by-partuuid/${image_device#PARTUUID=}
```
wjt: Might it be simpler to pass `endless.image.device=/dev/disk/by-partuuid/5fda738f-765f-4c38-a3ab-bba8b62e2b90` directly from Grub, rather than inventing and handling `PARTUUID=`, `UUID=`? I appreciate that passing the full path from Grub would work with this branch as it stands; and I'm happy to be disagreed with on this. Just a thought.

dracut/image-boot/eos-image-boot-setup
```diff
@@ -0,0 +1,56 @@
...
+
+offset=0
+echo "$extents" | while read extent_offset extent_size; do
+  [ -z "${extent_offset}" -o -z "${extent_size}" ] && continue
+  # Convert bytes to sectors, rounding up
+  if [ $(( extent_size % 512 )) != 0 ]; then
+    echo "image-boot: extent size $extent_size not sector-aligned" >&2
+    exit 1
+  fi
+  echo "${offset} ${extent_size} linear ${host_device} $extent_offset"
```
wjt: Since `dumpexfat` produces bytes, and the table format documented in `dmsetup(8)` is in terms of sectors, I think this line is now wrong: you need to divide `extent_size` and `extent_offset` by 512.

dracut/image-boot/eos-image-boot-generator
dsd: PARTUUID= and UUID= is the standard way of passing this info from the bootloader, so I think it makes sense to follow that.

dracut/image-boot/eos-image-boot-setup
dsd: Sorry, I had already caught this in testing but forgot to push the updated commit.

wjt:
Looks good to me now, ship it!

PR #60, opened 2016-08-09
eos-gpt2mbr: Add script to convert GPT to MBR

carlocaione:
https://phabricator.endlessm.com/T12793

### DIFF
--- a/eos-repartition-mbr
+++ b/eos-repartition-mbr
@@ -8,12 +8,15 @@ if [ $# -lt 1 ]; then
 	exit 1
 fi
 
-root_part=$1
-root_partno=${root_part: -1}
-root_disk=${root_part%?}
+root_part="$1"
+root_partno="${root_part: -1}"
+root_part_base="${root_part:: -1}"
+eval $(lsblk -o PKNAME,PARTTYPE "${root_part}" --paths --pairs --nodeps)
+root_disk="$PKNAME"
+root_guid="$PARTTYPE"
 
 swap_partno=$((root_partno + 1))
-swap_part=${root_disk}${swap_partno}
+swap_part="${root_part_base}${swap_partno}"
 
 # check if we passed the full device file path
 if [ "${root_part//dev/}" = "${root_part}" ]; then
@@ -41,13 +44,9 @@ udevadm settle
 # take current partition table
 parts=$(sfdisk -d $root_disk)
 
-# take the root partition type GUID
-root_part_dev=${root_part#/dev/}
-root_guid=$(echo "$parts" | sed -n -e "s/.*$root_part_dev.*type=\([A-Z0-9-]*\).*/\1/p")
-
 # check against the DiscoverablePartitionsSpec partition type UUIDs
 dps_root_guid="4F68BCE3-E8CD-4DB1-96E7-FBCAF984B709"
-if [ "$root_guid" != "$dps_root_guid" ]; then
+if [ "${root_guid^^}" != "${dps_root_guid^^}" ]; then
 	echo "$0: $root_part doesn't have the correct GUID (read $root_guid)"
 	exit 1
 fi
@@ -68,17 +67,12 @@ find_and_remove_partition() {
 	local name=${1:?No name supplied to ${FUNCNAME}}
 	local guid=${2:?No GUID supplied to ${FUNCNAME}}
 
-	local partno=$(echo "$parts" | sed -n -e "s/[^:]*\([0-9]\) :.*$guid.*/\1/p")
-
-	if [ -z "${partno}" ]; then
+	if ! echo "$parts" | grep -q "$guid"; then
 		echo "$0: $name not found"
 		exit 1
 	fi
 
-	local part=${root_disk}${partno}
-	local part_dev=${part#/dev/}
-
-	parts=$(echo "$parts" | sed -e "/$part_dev/d")
+	parts=$(echo "$parts" | grep -v "$guid")
 }
 
 find_and_remove_partition "ESP" "C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
@@ -125,23 +119,25 @@ fi
 
 # Reset partition indexes
 root_disk_dev=${root_disk#/dev/}
-parts=$(echo "$parts" | sed -e "s/${root_disk_dev}.//")
+parts=$(echo "$parts" | sed -e "s/${root_disk_dev}[^: ]\+//")
 
 # GPT -> DOS
 parts=$(echo "$parts" | sed -e "s/label: gpt/label: dos/")
 
 # Set MBR partition type
 parts=$(echo "$parts" | sed -e "s/, type=$dps_root_guid/, type=83/")
-# Remove partition UUIDs
-parts=$(echo "$parts" | sed -e "s/, uuid=\S\+//g")
+# Remove partition UUIDs and GPT attributes
+parts=$(echo "$parts" | sed -e "s/, \(uuid\|attrs\)=[^\s,]\+//g")
 
 echo "$parts"
 echo "$parts" | sfdisk --force --no-reread $root_disk
 udevadm settle
+partprobe "${root_disk}"
+udevadm settle
 
 # mkswap with the new partition index
 swap_partno=$((swap_partno - 2))
-swap_part=${root_disk}${swap_partno}
+swap_part=${root_part_base}${swap_partno}
 [ -e "$swap_part" ] && mkswap -L eos-swap $swap_part
 
 # Remove GPT headers. It's a little tricky to work out where the backup one is
@@ -163,9 +159,10 @@ udevadm settle
 
 # Restore bootloader
 root_mountpoint=$(mktemp -d)
-mount ${root_disk}$((root_partno - 2)) ${root_mountpoint}
+mount "${root_part_base}$((root_partno - 2))" ${root_mountpoint}
 dd if=${root_mountpoint}/boot/grub/i386-pc/boot.img of=${root_disk} bs=446 count=1 conv=fsync
 udevadm settle
 dd if=${root_mountpoint}/boot/grub/i386-pc/core.img of=${root_disk} bs=512 seek=1 conv=fsync
 udevadm settle
 umount ${root_mountpoint}
+rmdir ${root_mountpoint}
--- a/eos-repartition-mbr
+++ b/eos-repartition-mbr
@@ -1,0 +1,171 @@
+#!/bin/bash -e
+# Copyright (C) 2016 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+# arguments checking
+if [ $# -lt 1 ]; then
+	echo "Usage: $0 <EOS partition>"
+	exit 1
+fi
+
+root_part=$1
+root_partno=${root_part: -1}
+root_disk=${root_part%?}
+
+swap_partno=$((root_partno + 1))
+swap_part=${root_disk}${swap_partno}
+
+# check if we passed the full device file path
+if [ "${root_part//dev/}" = "${root_part}" ]; then
+	echo "$0: Specify the full path for the root device file (i.e. /dev/sda3)"
+	exit 1
+fi
+
+# check if we passed the device file for the partition and not for the disk
+case ${root_partno} in
+	*[!0-9]* )
+		echo "$0: Specify the root partition (i.e. /dev/sda3)"
+		exit 1
+		;;
+esac
+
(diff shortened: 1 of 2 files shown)


eos-gpt2mbr
```diff
@@ -0,0 +1,158 @@
+#!/bin/bash -e
+# Copyright (C) 2015 Endless Mobile, Inc.
```
wjt: 2016

eos-gpt2mbr
```diff
@@ -0,0 +1,158 @@
...
+	while [ "${#ret}" != "1" ]; do
+		ret=${ret#?}
+	done
+	echo $ret
+}
+
+# arguments checking
+if [ $# -lt 1 ]; then
+	echo "Usage: $0 <EOS partition>"
+       	exit 0
```
wjt: Mismatched indentation!

eos-gpt2mbr
```diff
@@ -0,0 +1,158 @@
...
+
+# check if we passed the full device file path
+if [ "${root_part//dev/}" = "${root_part}" ]; then
+	echo "eos-gpt2mbr: Specify the full path for the root device file (i.e. /dev/sda3)"
+	exit 0
+fi
+
+# check if we passed the device file for the partition and not for the disk
+case ${root_partno} in
+	*[!0-9]* )
```
wjt: Should this pattern be `*[!0-9]` (ie no trailing wildcard)? I don't think this is a huge problem though.

eos-gpt2mbr
wjt: And shouldn't this (and the other error cases) `exit 1` rather than `exit 0`?

eos-gpt2mbr
```diff
@@ -0,0 +1,158 @@
...
+
+# remove ESP
+parts=$(echo "$parts" | sed -e "/$esp_part_dev/d")
+
+# calculate new partition sizes
+disk_size=$(blockdev --getsz $root_disk)
+part_size=$(blockdev --getsz $root_part)
+part_start=$(echo "$parts" | sed -n -e '$ s/.*start=[ ]\+\([0-9]\+\).*$/\1/p')
+part_end=$(( part_start + part_size ))
+echo "Dsize $disk_size Psize $part_size Pstart $part_start Pend $part_end"
```
wjt: Presumably this is debug output, to be removed?

eos-gpt2mbr
```diff
@@ -0,0 +1,158 @@
...
+# Remove GPT types and UUIDs
+parts=$(echo "$parts" | sed -e "s/, type=.*//")
+
+echo "$parts"
+echo "$parts" | sfdisk --force --no-reread $root_disk
+ret=$?
+echo "sfdisk returned $ret"
+udevadm settle
+
+[ -e "$swap_part" ] && mkswap -L eos-swap $swap_part
```
wjt: If we're removing the ESP, then won't all the partitions to be renumbered down by one? And so the swap partition will not have the device name we're expecting?

eos-gpt2mbr
```diff
@@ -0,0 +1,158 @@
...
+
+# GPT -> DOS
+parts=$(echo "$parts" | sed -e "s/label: gpt/label: dos/")
+
+# Remove GPT types and UUIDs
+parts=$(echo "$parts" | sed -e "s/, type=.*//")
+
+echo "$parts"
+echo "$parts" | sfdisk --force --no-reread $root_disk
+ret=$?
```
wjt: Will this line be reached if `sfdisk` fails, given `#!/bin/bash -e`?

eos-gpt2mbr
carlocaione: ok

eos-gpt2mbr
carlocaione: It's just to account for `partno > 9`

eos-gpt2mbr
carlocaione: why?

eos-gpt2mbr
carlocaione: damn, I moved the code block during the cleanup. I'll move it back.

eos-gpt2mbr
carlocaione: probably not

eos-gpt2mbr
carlocaione: Yes, but I followed the same convention as in `dracut/repartition/endless-repartition.sh`.

eos-gpt2mbr
wjt: Sure, I guess it's fine to leave it in

eos-gpt2mbr
```diff
@@ -0,0 +1,159 @@
+#!/bin/bash -e
+# Copyright (C) 2016 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+get_last_char() {
```
dsd: Now that this code is running under bash we can use bash's built in functionality to read the last char of a variable

eos-gpt2mbr
```diff
@@ -0,0 +1,159 @@
...
+  echo "eos-gpt2mbr: $root_part doesn't exist"
+  exit 1
+fi
+
+# udev might still be busy probing the disk, meaning that it will be in use.
+udevadm settle
+
+# Save boot.img to be restored after we write the new MBR
+# as described in http://pete.akeo.ie/2014/05/compiling-and-installing-grub2-for.html
+dd if=${root_disk} of=boot.img conv=fsync bs=446 count=1
```
dsd: should do this to /tmp or similar in case the current working dir is not writable

eos-gpt2mbr
```diff
@@ -0,0 +1,159 @@
...
+
+# Reset partition indexes
+root_disk_dev=${root_disk#/dev/}
+parts=$(echo "$parts" | sed -e "s/${root_disk_dev}.//")
+
+# GPT -> DOS
+parts=$(echo "$parts" | sed -e "s/label: gpt/label: dos/")
+
+# Remove GPT types and UUIDs
+parts=$(echo "$parts" | sed -e "s/, type=.*//")
```
dsd: Don't we need to specify MBR partition type IDs?

eos-gpt2mbr
```diff
@@ -0,0 +1,159 @@
...
+echo "$parts" | sfdisk --force --no-reread $root_disk
+udevadm settle
+
+# mkswap with the new partition index
+swap_partno=$((swap_partno - 1))
+swap_part=${root_disk}${swap_partno}
+[ -e "$swap_part" ] && mkswap -L eos-swap $swap_part
+
+# Remove GPT primary partition table
+dd if=/dev/zero of=${root_disk} conv=fsync bs=512 count=1 seek=1
```
dsd: need to do a "udevadm settle" after each time the disk is opened for writing then closed, because udev will recreate the devices, there is a moment when they don't exist

eos-gpt2mbr
wjt: The `${var: -1}` syntax is new to me, thanks!

ramcq:
bikeshed, can we rename this like, eos-repartition-mbr or something? :)

eos-gpt2mbr
wjt: I suppose we do. Do you have a suggestion for the [type ID](http://www.win.tue.nl/~aeb/partitions/partition_types-1.html) to use for the partition formerly known as the BIOS-boot partition? The only type that seems vaguely similar to “contains the bootloader” is **f0**: “Linux/PA-RISC boot loader”…

The other option is to re-run `grub-bios-setup` within the image being rewritten (or `dd` `core.img` to the first 31 sectors), and drop the BIOS-boot partition entirely as well as the ESP.

dsd: f0 looks OK, another one to consider is da  Non-FS data

wjt: I decided to get rid of the BIOS-boot partition too and move GRUB to the conventional location on MBR disks. I think this is a bit neater and less likely to be accidentally deleted than a strange partition with a strange ID!

I would have preferred to use `grub-bios-setup` from inside the image, but it is a pain to get a proper chroot set up, so I just `dd` the images to the right place…

eos-repartition-mbr
```diff
@@ -0,0 +1,160 @@
...
+# mkswap with the new partition index
+swap_partno=$((swap_partno - 2))
+swap_part=${root_disk}${swap_partno}
+[ -e "$swap_part" ] && mkswap -L eos-swap $swap_part
+
+# Remove GPT primary partition table
+dd if=/dev/zero of=${root_disk} conv=fsync bs=512 count=1 seek=1
+
+# Remove GPT backup partition table
+dd if=/dev/zero of=${root_disk} conv=fsync bs=512 count=1 seek=$((part_size - 1))
```
wjt: I'm very suspicious about this line. I think the backup table is 1 sector from the end of the *disk*, not the partition; so I think this is actually zeroing 512 bytes of the root filesystem.

wjt:
I think I've fixed all the review feedback (including renaming it, @ramcq!) and I've fixed a bug in zeroing out the GPT header & footer (the fix is a bit clunky though…)

eos-repartition-mbr
```diff
@@ -149,10 +149,12 @@ dd if=/dev/zero of=${root_disk} conv=fsync bs=512 count=1 seek=1
 
 # Remove GPT backup partition table
 dd if=/dev/zero of=${root_disk} conv=fsync bs=512 count=1 seek=$((part_size - 1))
+udevadm settle
 
 # Restore bootloader
 root_mountpoint=$(mktemp -d)
 mount ${root_disk}$((root_partno - 2)) ${root_mountpoint}
 dd if=${root_mountpoint}/boot/grub/i386-pc/boot.img of=${root_disk} bs=446 count=1 conv=fsync
```
dsd: Also need a settle after this one. The reason is that dd does open() on the disk as read-write and then close(). Upon close(), udev (via inotify) will reprobe the whole disk, which means: first delete all the nodes, second recreate them. Unless you wait for that to complete, the following dd command can occur right at the moment when the nodes have been deleted but not recreated.

wjt: Ah – I had assumed that modifying `sdb` would cause the partition nodes to be recreated, but not `sdb` itself. Fixed now.

dsd:
This looks good now, but I'd prefer to see it squashed into a smaller number  of commits, maybe just 1 or 2?

wjt:
Okay, will do.

wjt:
I decided 1 commit was enough, and to attribute it to its primary author ☺

eos-repartition-mbr
```diff
@@ -0,0 +1,171 @@
...
+
+# arguments checking
+if [ $# -lt 1 ]; then
+	echo "Usage: $0 <EOS partition>"
+	exit 1
+fi
+
+root_part=$1
+root_partno=${root_part: -1}
+root_disk=${root_part%?}
```
dsd: This won't work for devices like /dev/mmcblk0p1

eos-repartition-mbr
```diff
@@ -0,0 +1,171 @@
...
+
+# udev might still be busy probing the disk, meaning that it will be in use.
+udevadm settle
+
+# take current partition table
+parts=$(sfdisk -d $root_disk)
+
+# take the root partition type GUID
+root_part_dev=${root_part#/dev/}
+root_guid=$(echo "$parts" | sed -n -e "s/.*$root_part_dev.*type=\([A-Z0-9-]*\).*/\1/p")
```
dsd: could be done cleaner with something like `lsblk -n -o PARTUUID` ?

wjt: yes, and finding `${root_disk}` can be done more cleanly with `lsblk -o PKNAME`, so I can kill at least two birds with one patch

wjt:
@dsd I've kept this as a separate patch so you can see what's changed. I've tested this with a drive at `/dev/sdb` and a loopback device (where partitions are named `/dev/loop0p1`); I've also tested manually adding a swap partition before running this script, both in the case where the script considers there to be enough space for one and when it doesn't.

eos-repartition-mbr
```diff
@@ -8,12 +8,15 @@ if [ $# -lt 1 ]; then
 	exit 1
 fi
 
-root_part=$1
-root_partno=${root_part: -1}
-root_disk=${root_part%?}
+root_part="$1"
+root_partno="${root_part: -1}"
+root_part_base="${root_part:: -1}"
+eval $(sudo lsblk -o PKNAME,PARTTYPE "${root_part}" --paths --pairs --nodeps)
```
dsd: Do we need sudo here? I think the whole script needs root

wjt: True!

PR #62, opened 2016-08-27
Fix build failure around escaped backslashes

wjt:
https://phabricator.endlessm.com/T13145

Only very lightly tested I'm afraid…


wjt:
Submitted this to OBS and the package builds correctly again.

This would have been caught by Jenkins if `--enable-systemd` were the default… plenty of stuff in this package, such as the stuff which enables image boot, assumes we're using systemd, so I don't see much value in the conditional!

dbnicholson:
This is probably due to the make upgrade changing how `\` is handled. I'd like to see what the automake generated Makefile.in looks like compared to the actual Makefile generated by configure.

As for `--enable-systemd` not being used in jenkins, it's due to the paths to the systemd unit directory. `distcheck` checks that all files are installed under `${prefix}`. Since the unit files have a fixed path under `/lib`, this would break.

There is a standard fix for that - make the unit directory configurable and then override in `DISTCHECK_CONFIGURE_FLAGS`. Something like https://github.com/endlessm/gnome-software/commit/b320c259bfe139d94f58d5760341d791182b7bd7.

Makefile.am
```diff
@@ -11,18 +11,28 @@ dist_systemdunit_DATA = \
...
-# split this out into EXTRA_DIST unescaped.
+# Unfortunately, the escaped \x2d is handled poorly, so we distribute and
+# install this unit under another name and rename it by hand.
 systemdunit_DATA = \
-	var-endless\\x2dextra.mount
+	var-endless_x2dextra.mount
+
+install-data-hook:
+	pushd "$(DESTDIR)$(systemdunitdir)"; \
+	mv var-endless_x2dextra.mount var-endless\\x2dextra.mount; \
```
dbnicholson: I would quote the `\` escaped one so that you at least avoid the shell stripping one of the `\`s.

Makefile.am
```diff
@@ -11,18 +11,28 @@ dist_systemdunit_DATA = \
...
+	popd
+
+uninstall-hook:
+	pushd "$(DESTDIR)$(systemdunitdir)"; \
+	rm -f var-endless\\x2dextra.mount; \
```
dbnicholson: Same.

Makefile.am
dbnicholson: Also, it might be better to `install` directly instead of implicit `install` + `mv`. I.e, remove the `systemdunit_DATA` setting above so that all automake does is dist the file from the `EXTRA_DIST` below. Then:
```
install-data-hook:
        $(MKDIR_P) $(DESTDIR)$(systemdunitdir)
        $(INSTALL_DATA) $(srcdir)/var-endless_x2dextra.mount $(DESTDIR)$(systemdunitdir)/'var-endless\x2dextra.mount'
```

dbnicholson:
Couple comments, but nothing critical. It's sad that this keeps breaking. Not sure if you want to try to handle the systemd unit directory thing so that `--enable-systemd` can be always on and the option dropped.

dbnicholson:
I went ahead and applied my suggestions and did the work to make systemd integration unconditional. CR ready.

PR #65, opened 2016-09-27
Provide read-only access to the filesystem hosting endless.img

jprvita:
We are mapping the partition passed in the `endless.image.device=` kernel command line argument to a device mapper device called `host`, and installing the necessary polkit rules to allow for regular users to mount it through the file manager (so a dual-boot user can access its files on the windows partition).

When setting up the dm device, we're mapping the blocks where `endless.img` lives to `error`, so any I/O attempt on them through this dm device fails, adding an additional layer of protection to `endless.img`. This implies that the flasher tool will have to read from dm device which maps `endless.img` and not from the `endless.img` file. Reading from the dm device also should have a better performance than reading from the `endless.img` file, since we will not have to go through FUSE.

Additionally, this also hides the partition passed through `endless.image.device=` from the file manager, since it can't be mounted anyway because it is in use by the device mapper.

https://phabricator.endlessm.com/T13135

### DIFF
--- a/eos-image-boot-dm-setup
+++ b/eos-image-boot-dm-setup
@@ -1,0 +1,96 @@
+#!/bin/bash
+# Copyright (C) 2016 Endless Mobile, Inc.
+# Licensed under the GPLv2
+#
+# When booting from an image file hosted on a filesystem, this program maps the
+# host filesystem to a dm-device with a hole where the image file lives.
+
+for i in $(</proc/cmdline); do
+  case $i in
+  endless.image.device=*)
+    host_device=${i//endless.image.device=}
+    ;;
+  endless.image.path=*)
+    image_path=${i//endless.image.path=}
+    ;;
+  esac
+done
+
+dm_roflag="--readonly"
+
+[ -z "${host_device}" -o -z "${image_path}" ] && exit 1
+
+case "${host_device}" in
+PARTUUID=*)
+  host_device=/dev/disk/by-partuuid/${host_device#PARTUUID=}
+  ;;
+UUID=*)
+  host_device=/dev/disk/by-uuid/${host_device#UUID=}
+  ;;
+esac
+
+host_device_size=$(blockdev --getsz ${host_device})
+if [ $? != 0 ]; then
+  echo "Failed to read ${host_device} size"
+  exit 1
+fi
+
+fstype=$(blkid -o value -s TYPE "${host_device}")
+if [ $? != 0 ]; then
+  echo "Failed to detect filesystem type on ${host_device}"
+  exit 1
+fi
+
+case "${fstype}" in
+exfat)
+  extents=$(dumpexfat -f "${image_path}" "${host_device}")
+  ;;
+ntfs)
+  extents=$(ntfsextents "${host_device}" "${image_path}")
+  ;;
+*)
+  echo "Unsupported filesystem ${fstype} on ${host_device}"
+  exit 1
+esac
+
+if [ $? != 0 ]; then
+  echo "Failed to lookup ${image_path} on ${host_device} (${fstype})"
+  exit 1
+fi
+
+# Sort the file extents so we create the holes in the correct places
+extents=$(echo "$extents" | sort -n)
+
+# Cleanup the device mapper table file from previous runs
+rm -f /tmp/dmtable
+
+i=0
+while read extent_i extent_size; do
+  [ -z "${extent_i}" -o -z "${extent_size}" ] && continue
+
+  if [ $((extent_size % 512)) != 0 ]; then
+    echo "Extent size $extent_size is not sector-aligned"
+    exit 1
+  fi
+
+  extent_size=$((extent_size / 512))
+  extent_i=$((extent_i / 512))
+
+  if [ ${i} -lt ${extent_i} ]; then
+    echo "${i} $((extent_i - i)) linear ${host_device} ${i}" >> /tmp/dmtable
+    echo "${extent_i} ${extent_size} error" >> /tmp/dmtable
+    i=$((extent_i + extent_size))
+  fi
+done <<< "$(echo "$extents")"
+
+if [ ${i} -lt ${host_device_size} ]; then
+  echo "${i} $((host_device_size - i)) linear ${host_device} ${i}" >> /tmp/dmtable
+fi
+
+dmsetup create endless-image-device $dm_roflag < /tmp/dmtable
+if [ $? != 0 ]; then
+  echo "Failed to set up a device map for ${host_device}"
+  exit 1
+fi
+
+exit 0
--- a/10-allow-mounting-endless-image-device.rules
+++ b/10-allow-mounting-endless-image-device.rules
@@ -1,0 +1,7 @@
+polkit.addRule(function(action, subject) {
+  if (action.id == "org.freedesktop.udisks2.filesystem-mount-system" &&
+      action.lookup("device") == "/dev/mapper/endless-image-device" &&
+      subject.local && subject.active) {
+    return polkit.Result.YES;
+  }
+});
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,6 +32,11 @@ uninstall-hook:
 udevrulesdir = $(udevdir)/rules.d
 dist_udevrules_DATA = 50-meson-vdec.rules
 
+polkitrulesdir = $(datadir)/polkit-1/rules.d
+dist_polkitrules_DATA = \
+	10-allow-mounting-endless-image-device.rules \
+	$(NULL)
+
 dist_sbin_SCRIPTS = \
 	eos-firstboot \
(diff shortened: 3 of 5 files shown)


eos-live-boot-dm-setup
```diff
@@ -0,0 +1,94 @@
...
+
+for i in $(cat /proc/cmdline); do
+  case $i in
+  endless.image.device=*)
+    host_device=${i//endless.image.device=}
+    ;;
+  endless.image.path=*)
+    image_path=${i//endless.image.path=}
+    ;;
+  endless.live_boot)
```
dsd: any reason to parse this? it's not used later

eos-live-boot-dm-setup
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+# Copyright (C) 2016 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+for i in $(cat /proc/cmdline); do
```
dsd: slightly better to use $(</proc/cmdline)

eos-live-boot-dm-setup.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Endless live boot device mapper setup
```
dsd: Both in the filenames in the descriptions, can you rename this to something without "live", since it is used for both live boots (meaning the root device is kept read only and we put on a lot of overlayfs on top) and non-live boots (e.g. dual boot with windows). Perhaps eos-image-boot-dm-setup or something like that

eos-live-boot-dm-setup
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+# Copyright (C) 2016 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
```
dsd: please add a 2-3 line comment explaining the purpose of this script

10-allow-mounting-dev-mapper-host.rules
```diff
@@ -0,0 +1,7 @@
+polkit.addRule(function(action, subject) {
```
dsd: do you need another rule for umounts?

eos-live-boot-dm-setup.service
jprvita: Makes sense.

eos-live-boot-dm-setup
jprvita: OK

eos-live-boot-dm-setup
jprvita: OK

10-allow-mounting-dev-mapper-host.rules
jprvita: No, by default the user can un-mount anything he has mounted.

eos-live-boot-dm-setup
jprvita: Actually, it was a feature removal leftover, since I was using it to decide whether or not the dm should be read-only. I'll will remove it for now since it is not needed yet.

eos-image-boot-dm-setup
```diff
@@ -0,0 +1,93 @@
...
+
+  if [ $((extent_size % 512)) != 0 ]; then
+    echo "Extent size $extent_size is not sector-aligned"
+    exit 1
+  fi
+
+  extent_size=$((extent_size / 512))
+  extent_i=$((extent_i / 512))
+
+  if [ ${i} -lt ${extent_i} ]; then
```
wjt: will this do the right thing if a file has two extents, with the second coming physically before the first?

wjt: I guess you can just pipe the extents through `sort -n`?

eos-image-boot-dm-setup
```diff
@@ -0,0 +1,93 @@
...
+while read extent_i extent_size; do
+  [ -z "${extent_i}" -o -z "${extent_size}" ] && continue
+
+  if [ $((extent_size % 512)) != 0 ]; then
+    echo "Extent size $extent_size is not sector-aligned"
+    exit 1
+  fi
+
+  extent_size=$((extent_size / 512))
+  extent_i=$((extent_i / 512))
```
wjt: This and the `case ${fstype}` block is copy-pasta from the other script – maybe we should have a stand-alone script which invokes the right one, and converts from bytes to sectors, called by both?

eos-image-boot-dm-setup
```diff
@@ -0,0 +1,96 @@
...
+# Copyright (C) 2016 Endless Mobile, Inc.
+# Licensed under the GPLv2
+#
+# When booting from an image file hosted on a filesystem, this program maps the
+# host filesystem to a dm-device with a hole where the image file lives.
+
+for i in $(</proc/cmdline); do
+  case $i in
+  endless.image.device=*)
+    host_device=${i//endless.image.device=}
```
wjt: What do you say to getting this from `/var/tmp/endless-image-host`, as saved in `eos-image-boot-generator`? The path could be saved in a similar file too, maybe.

eos-image-boot-dm-setup
```diff
@@ -0,0 +1,96 @@
...
+if [ $? != 0 ]; then
+  echo "Failed to lookup ${image_path} on ${host_device} (${fstype})"
+  exit 1
+fi
+
+# Cleanup the device mapper table file from previous runs
+rm -f /tmp/dmtable
+
+# Sort the file extents we create the holes in the correct places
+extents=$(sort ${extents})
```
wjt: I don't think this is going to work, `${extents}` is a load of newline-separated numbers, not a filename as `sort` expects. You'll also need `-n`:

```
  tensionsheet ~
% (echo 2; echo 12) | sort
12
2
  tensionsheet ~
% (echo 2; echo 12) | sort -n
2
12
```

jprvita: Yeah, you are right, I did some quick testing on files and stopped thinking before writing the actual code.

eos-image-boot-dm-setup
jprvita: The generator runs in the initrd, and the initrd's `/tmp` does not persist through the switchroot. I haven't gone as far as checking why this is the case, but the file just isn't there.

eos-image-boot-dm-setup
jprvita: Just realized I haven't replied to this one -- We could do some de-duplication here, but I don't think this is complicated enough to justify changing it at this point.

eos-image-boot-dm-setup
wjt: Ah! I understand – for some reason I thought this script also runs in the initrd, but of course not.

eos-image-boot-dm-setup
```diff
@@ -0,0 +1,93 @@
...
+  echo "Failed to detect filesystem type on ${host_device}"
+  exit 1
+fi
+
+case "${fstype}" in
+exfat)
+  extents=$(dumpexfat -f "${image_path}" "${host_device}" | sort -n)
+  ;;
+ntfs)
+  extents=$(ntfsextents "${host_device}" "${image_path}" | sort -n)
```
wjt: Unfortunately I think this now masks errors from `ntfsextents` because the return code of a pipeline is the return code of the rightmost command (unless you `set -o pipefail`):

```bash
  tensionsheet ~
% true | false; echo $?
1
  tensionsheet ~
% false | true; echo $?
0
  tensionsheet ~
% set -o pipefail; false | true; echo $?
1
```

eos-image-boot-dm-setup
wjt: Yes, fair enough.

eos-image-boot-dm-setup
```diff
@@ -0,0 +1,93 @@
...
+  if [ $((extent_size % 512)) != 0 ]; then
+    echo "Extent size $extent_size is not sector-aligned"
+    exit 1
+  fi
+
+  extent_size=$((extent_size / 512))
+  extent_i=$((extent_i / 512))
+
+  if [ ${i} -lt ${extent_i} ]; then
+    echo "${i} $((extent_i - i)) linear ${host_device} ${i}" >> /tmp/dmtable
```
wjt: I suppose we can assume that the first extent starts at a non-zero sector. Otherwise the first line of the table will be invalid:

```bash
% echo 0 0 linear /dev/sdb1 | sudo dmsetup create error-device
device-mapper: reload ioctl on error-device failed: Invalid argument
Command failed
```

(I was about to say that the conditional around the tail of the device is unnecessary because surely dm is okay with zero-length chunks, but no!)

eos-image-boot-dm-setup
wjt: I think in practice this is one of those things that will "never happen" – we wouldn't have managed to boot the system at all if this command had failed in the initramfs – but maybe better safe than sorry.

PR #70, opened 2016-10-19
T13741: live boot:  don't overlay /sysroot in its entirety

wjt:
If 'endless.live_boot' is on the command line, we pass '--readonly' to
dmsetup when mapping endless-image, and we mount it on /sysroot with the
'ro' flag. So, we don't need to overlay it with a tmpfs to protect it
from modification.

We need to mask:

* /sysroot/flatpak
* /sysroot/home (symlinked from /var/home and /home)
* /sysroot/ostree

as all these potentially need to be written to.

A side-effect of leaving /sysroot untouched is that we can successfully
read xattrs from it -- specifically, the eos-image-version attribute,
used by eos-phone-home (and other applications).

https://phabricator.endlessm.com/T13741


wjt:
This is incomplete…

PR #72, opened 2016-10-28
Support MBR on S905x

carlocaione:
https://phabricator.endlessm.com/T10685

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -9,13 +9,21 @@
 # the disk, and use it to create a swap partition.
 #
 # It is important to identify that this system is an Endless-flashed device
-# that desires such treatment. We do this by detecting that GPT flag #55 is
-# set on the root partition, which is deliberately set in EOS images. If you
-# need to avoid such repartitioning, just remove that flag before first boot.
+# that desires such treatment.
+
+# For platforms using a GPT partition table, we do this by detecting that GPT
+# flag #55 is set on the root partition, which is deliberately set in EOS
+# images. If you need to avoid such repartitioning, just remove that flag
+# before first boot.
 #
-# The 55 flag is also removed after this script is run, providing a
-# quick/cheap way of knowing that our work is done, avoiding going through
-# all this logic on each boot.
+# When dealing with a DOS partition table we do this by detecting that the type
+# code for unused partition entry 4 has magic number 'dd', which is
+# deliberately set in EOS images. If you need to avoid such repartitioning,
+# just remove that 4th partition before first boot.
+#
+# The 55 flag (or the "dd" marker) is also removed after this script is run,
+# providing a quick/cheap way of knowing that our work is done, avoiding going
+# through all this logic on each boot.
 #
 # This script is written carefully so that it can be interrupted after each
 # operation, on next boot it should effectively continue where it left off.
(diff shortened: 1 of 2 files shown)


eos-extra-resize
```diff
@@ -43,10 +50,13 @@ if [ "$marker" = "GUID:55" ]; then
...
     part_start=$(< /sys/class/block/$extra_part_name/start)
 
-    # The partition table is recreated below, which also includes the
-    # magic 55 attribute being dropped.
+    # The partition table is recreated below
     echo "Try to resize $extra_part at sector $part_start to fill $extra_disk"
-    echo "$part_start" | sfdisk --force --no-reread --label gpt $extra_disk
+    echo "$part_start" | sfdisk --force --no-reread --label $pt_label $extra_disk
+
+    # For MBR the marker is removed in endless-repartition.sh when rewriting
```
dsd: Actually for the extra disk, for both GPT and MBR the marker is removed in this script 2 lines above, we rewrite the whole partition table in that sfdisk command.

eos-extra-resize
```diff
@@ -32,9 +32,16 @@ if [ -z "$extra_disk" ]; then
     exit 0
 fi
 
+pt_label=$(blkid -o value -s PTTYPE $extra_disk)
+
+if [ "$pt_label" = "gpt" ]; then
+  marker=$(sfdisk --force --part-attrs $extra_disk 1)
+else
+  marker=$(sfdisk --force --part-type $extra_disk 4)
```
dbnicholson: Am I wrong, or are you reading the wrong partition in the MBR case? There aren't 4 partitions in the extra disk.

dracut/repartition/endless-repartition.sh
```diff
@@ -80,9 +80,18 @@ if [ -z "${root_disk}" ]; then
   exit 0
 fi
 
+pt_label=$(blkid -o value -s PTTYPE $root_disk)
+
 # Check for our magic "this is Endless" marker
-marker=$(sfdisk --force --part-attrs $root_disk $partno)
-if [ "$marker" != "GUID:55" ]; then
+if [ "$pt_label" = "dos" ]; then
+  marker=$(sfdisk --force --part-type $root_disk 4)
```
dbnicholson: Any reason you're using `4` here instead of `$partno`?

carlocaione: Uhm, `4` is wrong but also `$partno` is not correct either. `$partno` is the partition number for the rootfs where is the marker for the GPT case. For the DOS case we have a new additional partition so it should be `$partno+1` if I'm not mistaken.

dsd: In the previous implementation we always put the marker on partition 4 (which of course is not a real partition).

carlocaione: Right. Working on this after a while made me forget these details. 
We are _still_ using the partition `4` for this, so this line is actually correct. Thank for reminding this.

eos-extra-resize
carlocaione: As before we are using `printf "\xdd" | dd of=${disk1_img} bs=1 count=1 seek=498 conv=notrunc` in `eos-image-builder` to write a fake partition `4`

eos-extra-resize
carlocaione: yeah, right.

eos-extra-resize
dbnicholson: OK.

PR #76, opened 2016-11-10
Fix old version split disk upgrading

dbnicholson:
Really fix upgrading a split disk system from 2.3.

https://phabricator.endlessm.com/T13889

### DIFF
--- a/eos-enable-extra-upgrade
+++ b/eos-enable-extra-upgrade
@@ -48,3 +48,9 @@ ln -sf "$mount_unit_path" "$wants_link"
 
 # Reload systemd so that the new unit is seen to be enabled
 systemctl daemon-reload
+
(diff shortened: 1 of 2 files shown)


dsd:
Where is `/var/eos-extra-resize` created these days? i can only see the code that creates `/var/eos-extra-resize2`

dbnicholson:
That's the old one that was created in the 2.3 days. Which is what we want (I think) because the current resizer that creates /var/eos-extra-resize2 (bb6337c3, landed it 2.6.0) doesn't need to handle the migration of the mount unit. We could check for either to be safe, though.

dsd:
Seems to make sense to use the old marker then, but perhaps you can point this out in a comment? Apart from that this looks good to me.

dbnicholson:
OK. I updated the first commit to add comments about why that stamp file suffices for checking.

PR #79, opened 2016-11-30
repartition: Add support for NVMe block devices

dsd:
Accept all devices ending with "p" then a number.
This will cover existing MMC cases (e.g. mmcblk1p3) and also
NVMe (e.g. nvme0n1p3).

https://phabricator.endlessm.com/T14393

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -59,19 +59,20 @@ get_last_char() {
 }
 
 # We have to work with a variety of setups here:
-# - Either SCSI or MMC block devices with different naming conventions
+# - SCSI/MMC/NVMe block devices with different naming conventions
 # - A different partition numbering scheme as not all configurations have
 #   a ESP and a BIOS boot partition.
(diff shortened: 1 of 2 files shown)


dracut/repartition/endless-repartition.sh
```diff
@@ -59,13 +59,13 @@ get_last_char() {
...
 # We have to work with a variety of setups here:
-# - Either SCSI or MMC block devices with different naming conventions
+# - SCSI/MMC/NVMe block devices with different naming conventions
 # - A different partition numbering scheme as not all configurations have
 #   a ESP and a BIOS boot partition.
 partno=$(get_last_char ${root_part})
 swap_partno=$((partno + 1))
 case ${root_part} in
-  /dev/mmcblk?p?)
+  /dev/*p[0-9])
```
wjt: I suppose *technically* if you had 16 ATA devices, `/dev/sdp3` would match this branch and things would go wrong; so the `sd??` match should come before this more-lenient match. I don't think this is especially likely…

dsd:
updated with that suggestion

PR #80, opened 2016-12-02
 eos-repartition-mbr: leave expanding and swap to first boot 

dsd:
https://phabricator.endlessm.com/T13849

### DIFF
--- a/eos-repartition-mbr
+++ b/eos-repartition-mbr
@@ -82,45 +82,6 @@ fi
 find_and_remove_partition "ESP" "C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
 find_and_remove_partition "BIOS boot" "21686148-6449-6E6F-744E-656564454649"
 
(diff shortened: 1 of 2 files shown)


eos-repartition-mbr
```diff
@@ -140,10 +101,9 @@ udevadm settle
 partprobe "${root_disk}"
 udevadm settle
 
-# mkswap with the new partition index
-swap_partno=$((swap_partno - 2))
-swap_part=${root_part_base}${swap_partno}
-[ -e "$swap_part" ] && mkswap -L eos-swap $swap_part
+# Set marker on 4th partition so that the partition gets enlarged to fill the
+# disk on first boot
+printf "\xdd" | dd of=${root_disk} bs=1 count=1 seek=498 conv=notrunc
```
wjt: Won't you need a `udevadm settle` after this, in case the `${root_disk}` device is not recreated in time for the scary reads below?

PR #81, opened 2016-12-13
T13557: add support for booting from an ISO image

muresandone:

### DIFF
--- a/dracut/endless.conf
+++ b/dracut/endless.conf
@@ -1,3 +1,3 @@
 dracutmodules="dracut-systemd systemd-initrd dash drm eos-repartition eos-image-boot plymouth kernel-modules resume ostree systemd base fs-lib"
 fscks="fsck fsck.ext4"
-filesystems="ext4"
+filesystems="ext4 isofs"
--- a/dracut/image-boot/module-setup.sh
+++ b/dracut/image-boot/module-setup.sh
@@ -10,9 +10,9 @@ depends() {
 }
 
 install() {
-  dracut_install blockdev kpartx dmsetup dumpexfat ntfsextents lsblk
+  dracut_install blockdev kpartx dmsetup dumpexfat ntfsextents lsblk losetup
   instmods overlay
-  inst_rules 55-dm.rules
+  inst_rules 55-dm.rules 60-cdrom_id.rules
   inst_script "$moddir"/eos-image-boot-setup /bin/eos-image-boot-setup
   inst_simple "$moddir"/eos-image-boot-setup.service \
 	"$systemdsystemunitdir"/eos-image-boot-setup.service


dracut/image-boot/eos-image-boot-setup
```diff
@@ -18,12 +18,30 @@ if getargbool 0 endless.live_boot; then
...
+
+    # mount the iso image
+    mkdir -p /cd
+    mount ${host_device} /cd
+
+    # manually create the loop device needed for the device mapper
+    mknod -m660 $loopdev b 7 8
+
+    # setup the image partitions
+    losetup -v -r $loopdev "/cd/${image_path}"
```
wjt: Might be able to pass `--partscan` here and avoid needing a separate `kpartx` call?

dracut/image-boot/eos-image-boot-setup
```diff
@@ -18,12 +18,30 @@ if getargbool 0 endless.live_boot; then
...
+    kpartx -r -a -v $loopdev
+
+    # if not an iso: fallback to normal booting
```
wjt: This comment seems misleading to me, it doesn't describe what `exit 0` does here. Maybe more like “for ISOs, that's it: no extent-map needed.”

dracut/image-boot/eos-image-boot-setup
```diff
@@ -18,12 +18,30 @@ if getargbool 0 endless.live_boot; then
...
 
-fstype=$(lsblk --noheadings -o FSTYPE "${host_device}")
+fstype=$(lsblk -d --noheadings -o FSTYPE "${host_device}")
 if [ $? != 0 ]; then
   echo "image-boot: failed to detect filesystem type"
   exit 1
 fi
 
+if [ "$fstype" = "iso9660" ]; then
+    loopdev="/dev/mapper/endless-image"
```
wjt: Can we explain what we're doing in this block? “For ISO host filesystems, we can use a normal loop device, but its path needs to match the device created by the dm-setup logic on other filesystems.” perhaps?

dracut/image-boot/eos-image-boot-setup
dsd: I recommend avoiding that, at least I found it to be racy a few years ago. https://lkml.org/lkml/2012/9/7/535

dracut/image-boot/eos-image-boot-setup
dsd: I'm also curious why we can't just let losetup pick a loop device of its own name and number

muresandone:
Commit updates, please review.

dsd:
I don't think it would be unreasonable to improve the logic in those places
(if relevant) to allow for ISO support - but I'll leave this decision to
Will.

dsd:
Or alternatively we could create the loopback device then create a
device-mapper linear device that corresponds exactly to it.

dracut/image-boot/eos-image-boot-setup
wjt: Interesting, thanks!

wjt:
@dsd interesting idea. So (pseudocode) we'd have an `iso9660` branch in the fstype switch, which would do something like:

```bash
mount $host_device /cd
host_device=`losetup --find --show /cd/${image_path}`
# so $host_device == /dev/loop0 in practice
size_bytes=`lsblk --output SIZE --nodeps --bytes --noheadings ${host_device}`
extents="0 ${size_bytes}"
```

and then the rest of the script would continue as "normal", setting up a trivial mapped device corresponding to the entire loopback device, etc. I like the sound of this more than creating a loop device with the right maj/min numbers by hand, even if there's an extra level of indirection. If we were to add xz decompression to device mapper then this indirection would be needed anyway, of course!

I hear your suggestion about changing the places where `/dev/mapper/endless-image` is hard-coded. In `eos-installer`, it would be easy enough to look up the parent device for the root filesystem. For the systemd units written by `eos-image-boot-generator`, could one arrange for them to be generated by `eos-image-boot-setup` itself? Otherwise we'd have to duplicate the “is it an ISO?” logic in a number of places, which isn't necessarily an improvement. My feeling is that arranging for that well-known path to be created by dm-setup in the normal way (as above) might be enough.

wjt:
At @carlocaione's suggestion, let's merge the Dracut fix; I'll open another PR for the image-boot patch.

carlocaione:
@wjt trying your idea but device-mapper doesn't deal well with loopback devices. When mapping `loop0` with a table like `0 12202152 /dev/loop0 0`, DM complains: `device-mapper: table: 253:0: /dev/loop0: unknown target type`

dsd:
Try: 0 12202152 linear /dev/loop0 0

carlocaione:
@dsd of course. Sorry. It works. Now we have to decide if we like the extra loop level :)

PR #82, opened 2016-12-19
reparition: Don't leave stale boot entries behind

jprvita:
Changing the ESP's UUID breaks the Endless UEFI boot entry, since it
uses it as part of the loader path. To avoid stale boot entries to
accumulate on the machine, lets remove the Endless boot entry pointing
to the original ESP's UUID.

We could create a new boot entry here, but lets leave this job to
fallback.efi, so we do not need to keep the bootloader's path in sync.

https://phabricator.endlessm.com/T14430

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -132,6 +132,11 @@ elif [ $lastpart -gt $swap_partno ]; then
   exit 0
 fi
 
+# read the current ESP's UUID
+if [ "$pt_label" = "gpt" ]; then
+  orig_esp_uuid=$(sfdisk --force --part-uuid $root_disk 1)
+fi
+
 # calculate new partition sizes
 disk_size=$(blockdev --getsz $root_disk)
 part_size=$(blockdev --getsz $root_part)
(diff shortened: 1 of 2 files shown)


dracut/repartition/endless-repartition.sh
```diff
@@ -191,6 +196,11 @@ udevadm settle
 if [ -x /usr/sbin/amlogic-fix-spl-checksum ]; then
   /usr/sbin/amlogic-fix-spl-checksum $root_disk
   udevadm settle
+elif [ "$pt_label" = "gpt" ]; then
+  # Remove the stale boot entry pointing to the old ESP's UUID.
+  # fallback.efi will take care of creating a new one on the next boot
+  boot_num=$(efibootmgr -v -d $root_disk | grep -i $orig_esp_uuid | grep "Endless" | cut -c 5-8)
```
wjt: If we end up using this approach, I think this will do the wrong thing on dual-boot systems. The ESP in the `endless.img` (the file mapped as $root_disk) is not used; the ESP on the host drive is used. So, this line will assign empty string to `boot_num` and then the following `efibootmgr` invocation will be wrong.

In the dual-boot case, changing the UUIDs will actually have no effect at all on the boot entry so we could just skip this part there.

wjt:
Superseded by https://github.com/endlessm/shim/pull/8.

PR #83, opened 2017-01-09
eos-image-boot-setup: add support for booting from an ISO image

wjt:


wjt:
This PR is the second half of https://github.com/endlessm/eos-boot-helper/pull/81

wjt:
Superseded by #86

PR #84, opened 2017-01-09
repartition: don't delete recovery partition at end of disk

dsd:
Asus will start shipping units in China with a recovery partition
added at the end of the disk before Endless first boot.

We can detect this by its GPT partition type UUID and then avoid
deleting it.

Since swap is no longer (always) the final partition we can't
rely on sfdisk making it fill the remaining space, so we need to
calculate and specify it's final size.

https://phabricator.endlessm.com/T14827

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -120,15 +120,28 @@ if [ "$pt_label" = "dos" ]; then
   parts=$(echo "$parts" | sed -e '$d')
 fi
 
-# check the last partition on the disk
-lastpart=$(echo "$parts" | sed -n -e '$ s/[^:]*\([0-9]\) :.*$/\1/p')
+# If we detect a recovery partition at the end of the disk, we'll be careful
+# not to destroy it.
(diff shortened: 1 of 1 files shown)


dracut/repartition/endless-repartition.sh
```diff
@@ -121,14 +121,31 @@ if [ "$pt_label" = "dos" ]; then
...
+  preserve_start=$(echo "$parts" | sed -n -e '$ s/.*start=[ ]*\([^,]*\).*/\1/p')
+
+  # Save the partition entry to be restored later, and delete it from the
+  # list for now, to simplify the logic that follows.
+  preserve_partition=$(echo "$parts" | sed -n -e '$ s/[^:]*:\(.*\)$/\1/p')
+  parts=$(echo "$parts" | sed '$d')
+
+  # Recalculate lastpartno now that we've temporarily excluded the last
+  # partition
+  lastpartno=$(echo "$parts" | sed -n -e '$ s/[^:]*\([0-9]\) :.*$/\1/p')
```
wjt: You don't use `lastpartno` in this `if` block – could we do this initialization once, just before it's used?

dsd:
@wjt updated the commit

PR #85, opened 2017-01-10
Add GNOME flatpak repositories if needed after an OS upgrade

mariospr:
https://phabricator.endlessm.com/T14665

### DIFF
--- a/eos-add-flatpak-gnome-repos
+++ b/eos-add-flatpak-gnome-repos
@@ -1,0 +1,50 @@
+#!/bin/bash -e
+# Copyright (C) 2016 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+# Add the stable GNOME flatpak repositories after an OS upgrade
+
+TMPDIR=`mktemp -d`
+trap '[ -n "$TMPDIR" ] && rm -rf "$TMPDIR"' EXIT
+
+# Sanity check
+if [ -z ${TMPDIR} ] || [ ! -d ${TMPDIR} ] ; then
+    echo "Error creating temporary directory, exiting"
+    exit 1
+fi
+
+# This information for setting up these repos is stable, so we
+# can just hardcode here the contents of the .flatpakrepo files
+# so that we don't need to rely on a network connection.
+
+# GNOME Runtimes flatpak repositories
+cat <<EOF > ${TMPDIR}/gnome.flatpakrepo
+[Flatpak Repo]
+Title=Gnome Stable Runtimes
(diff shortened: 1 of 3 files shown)


eos-add-flatpak-gnome-repos
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash -e
+# Copyright (C) 2016 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+# Add the stable GNOME flatpak repositories after an OS upgrade
+
+TMPDIR=`mktemp -d`
```
dbnicholson: I think you should remove this in a `trap` so that it gets cleaned up even if there's an error below:
```
trap "rm -rf "$TMPDIR" EXIT
```

eos-add-flatpak-gnome-repos
```diff
@@ -0,0 +1,52 @@
...
+
+# Sanity check
+if [ -z ${TMPDIR} ] || [ ! -d ${TMPDIR} ] ; then
+    echo "Error creating temporary directory, exiting"
+    exit 0
+fi
```
dbnicholson: This block is probably not needed since you're running with `-e` and bash will exit immediately if the `mktemp` fails. However, if you do want to keep the sanity check, I'd change this to `exit 1` so it shows up as a failure if the temporary directory wasn't created.

eos-add-flatpak-gnome-repos
dbnicholson: You can also add your paranoia here: `trap '[ -n "$TMPDIR" ] && rm -rf "$TMPDIR"' EXIT`. I forgot about quoting this correctly, too.

eos-add-flatpak-gnome-repos.service
```diff
@@ -0,0 +1,18 @@
+# Add the stable GNOME flatpak repositories after an OS upgrade
+
+[Unit]
+Description=Configure the stable GNOME flatpak repositories
+DefaultDependencies=no
+Conflicts=shutdown.target
```
dbnicholson: I think you want to add `After=local-fs.target` and `Wants=local-fs.target`. This would fail if systemd decided to run it before all the filesystems were mounted.

eos-add-flatpak-gnome-repos
mariospr: Ok

eos-add-flatpak-gnome-repos
mariospr: Ok

PR #86, opened 2017-01-10
image-boot: Enable iso9660 booting

carlocaione:


carlocaione:
@wjt in the CDROM case `$host_device` is `/dev/sr0`. There are no holder devices or slaves so `-d` is useless. In the HDA case `/dev/sda1` is already mapping the whole device, so also in this case `-d` is redundant.

PR #88, opened 2017-01-14
Provide service to fix broken flatpak desktop files

dbnicholson:
For 3.0.10 and 3.0.11, the image builder was changed so that it
installed from the eos3.0 refs instead of the eos3 refs so that there
wasn't interference from eos3.1. To make it look like eos3 apps were
installed, the app checkout directories were changed to look that way.
Unfortunately, the branch is also embedded in the desktop file when
flatpak exports it, and this was not updated.

Provide a boot service that tries to find and fix these desktop files.
This will run only on upgrade.

https://phabricator.endlessm.com/T15005

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,6 +14,7 @@ dist_systemdunit_DATA = \
 	eos-enable-zram.service \
 	eos-image-boot-dm-setup.service \
 	eos-live-boot-overlayfs-setup.service \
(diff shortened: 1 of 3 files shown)


dbnicholson:
This is almost completely untested, but I did run the script on the one app I had installed to make sure the `sed` was appropriate.

PR #94, opened 2017-01-26
ISO: Do not set UDISKS_IGNORE on iso9660 partitions

carlocaione:
https://phabricator.endlessm.com/T14308

### DIFF
--- a/dracut/image-boot/eos-image-boot-generator
+++ b/dracut/image-boot/eos-image-boot-generator
@@ -19,19 +19,25 @@ image_path=$(getarg endless.image.path=)
 # Make sure runtime udev rules dir exists
 mkdir -p /run/udev/rules.d
 
+# When booting an ISO image we cannot DM the host device to
+# 'endless-image-device' but we still need it to be visible and mountable by
+# udisk and eos-installer so we do not set UDISKS_IGNORE
+echo 'SUBSYSTEM=="block", ENV{ID_FS_TYPE}=="iso9660", GOTO="skip_ignore"' > /run/udev/rules.d/79-endless-image-device.rules
+
 case "${image_device}" in
 PARTUUID=*)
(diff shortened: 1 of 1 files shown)


dracut/image-boot/eos-image-boot-generator
```diff
@@ -19,6 +19,11 @@ image_path=$(getarg endless.image.path=)
 # Make sure runtime udev rules dir exists
 mkdir -p /run/udev/rules.d
 
+# When booting an ISO image we cannot DM the host device to
+# 'endless-image-device' but we still need it to be visible and mountable by
+# udisk and eos-installer so we do not set UDISKS_IGNORE
+echo "SUBSYSTEM==\"block\", ENV{ID_FS_TYPE}==\"iso9660\" GOTO=\"skip_ignore\"" > /run/udev/rules.d/79-endless-image-device.rules
```
wjt: I think you're missing a comma after `\"iso9660\"`.

wjt: I know the other lines in this file don't do this but why not use single quotes around the argument to `echo` to avoid having to escape all the double quotes?

dracut/image-boot/eos-image-boot-generator
```diff
@@ -19,6 +19,11 @@ image_path=$(getarg endless.image.path=)
...
+
 case "${image_device}" in
 PARTUUID=*)
   echo "SUBSYSTEM==\"block\", ENV{ID_PART_ENTRY_UUID}==\"${image_device#PARTUUID=}\", ENV{DM_NAME}!=\"endless-image-device\", ENV{UDISKS_IGNORE}=\"1\"" > /run/udev/rules.d/79-endless-image-device.rules
```
wjt: You are going to need to change these to use `>>` to append rather than overwrite.

dracut/image-boot/eos-image-boot-generator
carlocaione: wow, weird it worked fine

PR #97, opened 2017-02-27
add eos-firewall-localonly service

ramcq:
This implements a simple iptables/ip6tables firewall which allows the
host to communicate only with resources on the local network, plus a few
explicitly-enabled hosts (the Endless metrics / phone-home servers) and
services (DNS and NTP).

There is a script eos-firewall-localonly which sets up the initial
tables with common local network and local multicast addresses,
permitted interfaces (lo) and the permitted services, and a
NetworkManager dispatch script called eos-firewall-localonly-nm which
adds additional rules allowing traffic to the permitted hosts (so that the
DNS requests are not done at boot time before the network is available)
and any other local network routes which have been assigned via DHCP.

Denied TCP connections are reset so that the local application receives
an error immediately, and any blocked packets are logged to the journal
with a relatively low frequency.

For obvious reasons, the systemd service is disabled by default: this
functionality must be explicitly enabled on the hosts where it is
desirable.

https://phabricator.endlessm.com/T15709


Makefile.am
```diff
@@ -9,26 +9,40 @@ AM_DISTCHECK_CONFIGURE_FLAGS = \
 
 dist_systemdunit_DATA = \
 	eos-add-flatpak-gnome-repos.service \
-	eos-firstboot.service \
-	eos-extra-resize.service \
```
pwithnall: Nitpick: Sorting this list alphabetically should be done as a separate commit.

Makefile.am
```diff
@@ -9,26 +9,40 @@ AM_DISTCHECK_CONFIGURE_FLAGS = \
...
-	eos-fix-flatpak-branches.service \
 	$(NULL)
 
+# Network Manager dispatcher script for the firewall - scripts which
+# are a symlink to no-wait.d will be run without blocking/ordering.
+# install the script here and make the symlink in install-data-hook
+# below
+networkscriptdir = $(sysconfdir)/NetworkManager/dispatcher.d
+networkscriptnowaitdir = $(networkscriptdir)/no-wait.d
+dist_networkscriptnowait_DATA = eos-firewall-localonly-nm
```
pwithnall: Shame we can’t drop this into `/usr/local/share/NetworkManager/dispatcher.d/no-wait.d`, but it looks like NM only supports `$sysconfdir`. :-(

Makefile.am
```diff
@@ -41,12 +55,13 @@ dist_polkitrules_DATA = \
...
 dist_sbin_SCRIPTS = \
 	eos-add-flatpak-gnome-repos \
-	eos-firstboot \
-	eos-extra-resize \
 	eos-enable-extra-upgrade \
 	eos-enable-zram \
+	eos-extra-resize \
+	eos-firewall-localonly \
+	eos-firstboot \
+	eos-fix-flatpak-branches \
```
pwithnall: Nitpick: Alphabetising should be in a separate commit.

eos-firewall-localonly
```diff
@@ -0,0 +1,152 @@
+#!/bin/bash -e
```
pwithnall: Missing a copyright header.

eos-firewall-localonly
```diff
@@ -0,0 +1,152 @@
...
+    fi
+  done
+}
+
+# clear and remove any localonly chains
+reset() {
+  # disable first because otherwise we can't erase our chains
+  disable
+
+  for iface in $(ls /sys/class/net); do
```
pwithnall: Do you want to limit that `ls` to directories only? I realise `remove_localonly_chain` will ignore non-present chains, but it seems a bit redundant to call it for things like `/sys/class/net/bonding_masters`.

eos-firewall-localonly-nm
```diff
@@ -0,0 +1,158 @@
+#!/bin/bash -e
```
pwithnall: Missing a copyright header.

eos-firewall-localonly
pwithnall: Nitpick: This would be so much more readable as a Python script. (Not expecting this review comment to get addressed. Just venting some hatred for bash a bit.)

eos-firewall-localonly-nm
```diff
@@ -0,0 +1,158 @@
...
+  dhcp6-change)
+    ensure_localonly_chain "${IFACE_CHAIN}"
+    add_rules_v6 "${IFACE_CHAIN}" "${IP6_NUM_ADDRESSES}" "${IP6_NUM_ROUTES}"
+    update_metrics
+    ;;
+  down)
+    remove_localonly_chain "${IFACE_CHAIN}"
+    ;;
+  *)
+    exit 0
```
pwithnall: Shouldn’t it exit with an error if the `${ACTION}` is unrecognised?

pwithnall:
I’ll test with eos-updater now.

eos-firewall-localonly.service
```diff
@@ -0,0 +1,12 @@
+# Run eos-firewall-localonly before NetworkManager starts
+
+[Unit]
+Description=Endless local-only firewall
+Before=NetworkManager.service
+
+[Service]
+Type=oneshot
```
pwithnall: It’s a bit counterintuitive that `sudo systemctl start eos-firewall-localonly.service` will enable the firewall, but `sudo systemctl stop eos-firewall-localonly.service` will *not* stop it.

pwithnall:
It correctly blocks outgoing HTTP traffic to the main EOS servers from the updater. It allows IPv4 LAN update requests (port 12345), but blocks the IPv6 ones.
```
Feb 27 17:16:44 endless eos-updater[1727]: Changing to state Polling
Feb 27 17:16:44 endless eos-updater[1727]: Using product branch os/eos/amd64/master
Feb 27 17:16:44 endless kernel: eos-firewall-localonly: IN= OUT=enp0s3 SRC=192.168.0.56 DST=72.14.189.30 LEN=60 TOS=0x00 PREC=0x00 TTL=64 ID=19298 DF PROTO=TCP SPT=38610 DPT=443 WINDOW=29200 RES=0x00 SYN URGP=0 
Feb 27 17:16:44 endless eos-updater[1727]: Failed to poll metadata from source main: Could not connect: Connection refused
Feb 27 17:16:44 endless eos-updater[1727]: Found name service EOS update service on endless on the network; type: _eos_updater._tcp, domain: local, protocol: 1, interface: 2
Feb 27 17:16:44 endless eos-updater[1727]: Found name service EOS update service on endless on the network; type: _eos_updater._tcp, domain: local, protocol: 0, interface: 2
Feb 27 17:16:44 endless eos-updater[1727]: Using product branch os/eos/amd64/master
Feb 27 17:16:44 endless kernel: eos-firewall-localonly: IN= OUT=enp0s3 SRC=fd08:5c1d:fae8:0000:0b5d:4953:5b0f:6c2c DST=fd08:5c1d:fae8:0000:a611:d757:3d30:aec8 LEN=80 TC=0 HOPLIMIT=64 FLOWLBL=602191 PROTO=TCP SPT=39088 DPT=12345 WINDOW=28800 RES=0x00 SYN URGP=0 
Feb 27 17:16:44 endless eos-updater[1727]: Failed to fetch latest commit from http://[fd08:5c1d:fae8:0:a611:d757:3d30:aec8]:12345/: Could not connect: Connection refused
Feb 27 17:16:45 endless eos-updater[1727]: Using product branch os/eos/amd64/master
Feb 27 17:16:45 endless eos-updater[1727]: Changing to state UpdateAvailable
```

The updater seems to be happily downloading all its object files over IPv4 now.

eos-firewall-localonly
dsd: What is bonding_masters? Here the directory only has symlinks to network interfaces. If we're OK with traversing everything it would be better to do `for iface in /sys/class/net/*` with nullglob set.

pwithnall: `bonding_masters` is a file I happen to have in my `/sys/class/net` (on Fedora), so it might not be the best example.

Makefile.am
ramcq: We want the script to be updatable via ostree, which will be the case if it lives in /etc and is not modified. We could host it somewhere off-path in /usr and place a symlink in /etc - I'd be happy to do that if it's preferred.

Makefile.am
ramcq: Fixed.

Makefile.am
ramcq: Fixed.

eos-firewall-localonly
ramcq: Fixed.

ramcq: Maybe, but it's really not that complex - just some loops and some variable substitution. There was nothing that made me think twice about how to do it in shell apart from the variable indirect lookup, but that was forced upon me by NetworkManager. :P

If the script had some python oneliners in it, we could very easily reach agreement here. :)

eos-firewall-localonly
ramcq: Replaced with find /sys/class/net -maxdepth 1 -type l -printf "%f\n" - as I run Endless I don't have any of this bonding funny business on my box... :P

eos-firewall-localonly-nm
ramcq: Fixed.

eos-firewall-localonly-nm
ramcq: Intentional - we don't respond to all events, and NetworkManager can (and has) added actions over time, so listing known actions and erroring here could start logging spurious failures in future. Refactored to silently exit 0 on the actions that exist but we don't care about, but to log a message and then exit 0 if new actions appear.

eos-firewall-localonly.service
ramcq: Historical accident as I only added start/stop syntax after I wrote the systemd unit. Fixed with ExecPostStop as discussed on Slack, which also cleans up in the case of a failed start.

ramcq:
Have added fd08:: addresses - Unique Local Addresses - to the whitelisted ranges.

Makefile.am
pwithnall: Excellent. Order is imposed.

pwithnall:
> It would be reassuring to see some integration tests for this, installed as installed-tests.

What was the outcome here?

ramcq:
>> It would be reassuring to see some integration tests for this, installed as installed-tests.
>
> What was the outcome here?

Ahem - forgot. https://phabricator.endlessm.com/T15889

pwithnall:
I don’t have any more review comments.

PR #98, opened 2017-03-01
image-boot-dm-setup: use zero, not error, for holes

wjt:
If there happens to be a hole just before the end of the disk, `mount`
and friends try to read it while probing the filesystem type. This
fails, so the drive can't be mounted. (`mount -t ntfs-3g
/dev/mapper/endless-image-device /mnt` works fine -- it is not ntfs-3g
which is reading inside the holes.)

The 'zero' dm target returns zeros when read and discards all writes,
like /dev/zero.

https://phabricator.endlessm.com/T15818

### DIFF
--- a/eos-image-boot-dm-setup
+++ b/eos-image-boot-dm-setup
@@ -1,9 +1,9 @@
 #!/bin/bash
-# Copyright (C) 2016 Endless Mobile, Inc.
+# Copyright (C) 2016-2017 Endless Mobile, Inc.
 # Licensed under the GPLv2
 #
 # When booting from an image file hosted on a filesystem, this program maps the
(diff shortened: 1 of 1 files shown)


eos-image-boot-dm-setup
```diff
@@ -76,7 +76,7 @@ while read extent_i extent_size; do
 
   if [ ${i} -lt ${extent_i} ]; then
```
wjt: I'm a bit suspicious of this comparison -- shouldn't this condition always be true, and wouldn't the mapped device get messed up if it's not?

jprvita: On a quick re-read, you are right, this should always be true. I can't remember the original reason for having this check here.

PR #100, opened 2017-03-16
dracut: Create customization module

jprvita:
Mechanism to ship image-specific customizations in the initramfs. For
now provides boot splash animation customization support.

https://phabricator.endlessm.com/T15989

### DIFF
--- a/dracut/customization/eos-install-custom-boot-splash
+++ b/dracut/customization/eos-install-custom-boot-splash
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+CUSTOM_SPLASH_DIR="/eos-customization/plymouth"
+CUSTOM_SPLASH_CONFIG="${CUSTOM_SPLASH_DIR}/plymouthd.defaults"
+PLYMOUTH_RUNTIME_DIR="/run/plymouth"
+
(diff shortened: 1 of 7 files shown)


dracut/customization/eos-install-custom-boot-splash
```diff
@@ -0,0 +1,27 @@
...
+	echo "${CUSTOM_SPLASH_DIR} not found" >&2
+	exit 1
+fi
+
+if [ ! -f ${CUSTOM_SPLASH_CONFIG} ] ; then
+	echo "${CUSTOM_SPLASH_CONFIG} not found" >&2
+	exit 1
+fi
+
+theme_name=`grep Theme ${CUSTOM_SPLASH_CONFIG}  | sed 's/Theme=\(.*\)/\1/'`
```
dsd: Is grep a new addition to the initramfs? If so can we avoid adding it? I think you should be able to do all this parsing with sed alone.

jprvita: I knew sed wasn't great matching across multiple lines, but I with some research I found the right spell: `sed -n 's/Theme=\(.*\)/\1/p'`

PR #102, opened 2017-04-05
repartition: Fix detection of basic data partition

dsd:
Shell syntax failure, can't have multiple matches without using |
on the same line.

Move back to the if style and wrap the test naturally onto 2 lines.

https://phabricator.endlessm.com/T16442

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -120,21 +120,19 @@ if [ "$pt_label" = "dos" ]; then
   parts=$(echo "$parts" | sed -e '$d')
 fi
 
-# If we detect a data or recovery partition at the end of the disk, we'll be
-# careful not to destroy it.
+# If we detect a recovery partition at the end of the disk, we'll be careful
+# not to destroy it.
 lastparttype=$(echo "$parts" | sed -n -e '$ s/.*type=\([A-F0-9-]\+\).*/\1/p')
-case ${lastparttype} in
-  BFBFAFE7-A34F-448A-9A5B-6213EB736C22)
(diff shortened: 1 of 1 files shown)


dracut/repartition/endless-repartition.sh
```diff
@@ -120,21 +120,19 @@ if [ "$pt_label" = "dos" ]; then
...
-
-    # Save the partition entry to be restored later, and delete it from the
-    # list for now, to simplify the logic that follows.
-    preserve_partition=$(echo "$parts" | sed -n -e '$ s/[^:]*:\(.*\)$/\1/p')
-    parts=$(echo "$parts" | sed '$d')
-    ;;
-esac
+if [ "$lastparttype" = "BFBFAFE7-A34F-448A-9A5B-6213EB736C22" \
+    -o "$lastparttype" = "EBD0A0A2-B9E5-4433-87C0-68B6B72699C7" ]; then
+  # We'll use the free space up to where this partition starts
```
wjt: Doh! I didn't realise this syntax doesn't work either.

You actually can use a `case`:

```bash
$ cat /tmp/foo.sh
uuid=EBD0A0A2-B9E5-4433-87C0-68B6B72699C7
case $uuid in
  BFBFAFE7-A34F-448A-9A5B-6213EB736C22 | EBD0A0A2-B9E5-4433-87C0-68B6B72699C7)
    echo matches
    ;;
esac
$ dash /tmp/foo.sh
matches
```

FWIW I narrowly prefer this, or `[ c ] || [ d ]]`, to `[ c -o d ]` but don't feel that strongly about it. Up to you if you want to change this to one of those alternatives, or leave this as is.

PR #105, opened 2017-04-20
repartition: add a test

wjt:
This test was useful for me in fixing a bug in this script --
https://phabricator.endlessm.com/T16442 -- but has the major limitation
that it must be run as root (or another user able to run `losetup`). I had
hoped that I could test a subset of the script by running `sfdisk` on a
file, but that would not expose the bug in question (a logic error in later
stages of the script).


wjt:
This is a superset of #104.

PR #106, opened 2017-05-02
Add eos-fix-home-dir-permissions

rshuler:
Starting with 3.1.6, we default to dir mode 700 for any
new user home directories, but previously we followed
the debian default of 755.

This service runs once after an upgrade from any version
prior to 3.1.6, to change the dir mode of any existing home
directories from 755 to 700.  Once run, a stamp file is
created to avoid running again, so that any further
tweaks by an admin to directory permissions will not be undone.

https://phabricator.endlessm.com/T16858

### DIFF
--- a/eos-fix-home-dir-permissions
+++ b/eos-fix-home-dir-permissions
@@ -1,0 +1,28 @@
+#!/bin/bash -e
+
+# Prior to 3.1.6, home directory permissions were created with dir mode 755.
+# This is now fixed, but on upgrade from an earlier version to 3.1.6,
+# let's perform a one-time clean up of any existing home dirs.
+
+stamp_dir=/var/lib/misc
+stamp_file="$stamp_dir"/eos-dir-mode-700
+
+if [ -f "$stamp_file" ]; then
+    echo "Dir mode 700 has already been set for home directories, exiting"
+    exit 0
+fi
+
+for home_dir in /home/*; do
+    [ -d "$home_dir" ] || continue
+
+    perm=$(stat -c %a "$home_dir")
+
+    if [ "$perm" = "755" ]; then
+        echo "Changing $home_dir dir mode from 755 to 700"
+        chmod 700 "$home_dir"
+    fi
+done
+
+# Create the stamp file to avoid ever running this again
+mkdir -p "$stamp_dir"
+touch "$stamp_file"
--- a/eos-fix-home-dir-permissions.service
+++ b/eos-fix-home-dir-permissions.service
@@ -1,0 +1,19 @@
+# Clean up any home directories which were created with
+# default permissions 755 rather than 700
+
+[Unit]
+Description=Fix existing home directory permissions from 755 to 700
+
+# Only run if the home dir permissions have not yet been cleaned up.
+# The default dir mode was changed for 3.1.6,
+# and we only want to perform this cleanup once, in case
+# an admin later intentionally changes home dir permissions.
+ConditionPathExists=!/var/lib/misc/eos-dir-mode-700
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/eos-fix-home-dir-permissions
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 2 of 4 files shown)


eos-fix-home-dir-permissions.service
```diff
@@ -0,0 +1,28 @@
...
+[Unit]
+Description=Fix existing home directory permissions from 755 to 700
+DefaultDependencies=no
+Conflicts=shutdown.target
+Wants=local-fs.target
+After=local-fs.target
+
+# Only run on updates
+Before=multi-user.target systemd-update-done.service
+ConditionNeedsUpdate=/etc
```
dbnicholson: Since this script is so simple and you're using a stamp file, I think I'd drop this and make the service much simpler. Once you drop `ConditionNeedsUpdate`, then you can have a normal service that doesn't need to run before `basic.target`. That simplifies things considerably since all the dependencies you care about will be automatically handled by systemd. I think it would reduce to (without comments):
```
[Unit]
Description=Fix existing home directory permissions from 755 to 700
ConditionPathExists=!/var/lib/eos-dir-mode-700

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=/usr/sbin/eos-fix-home-dir-permissions

[Install]
WantedBy=multi-user.target
```

eos-fix-home-dir-permissions.service
```diff
@@ -0,0 +1,28 @@
...
+# Only run if the home dir permissions have not yet been cleaned up.
+# The default dir mode was changed for 3.1.6,
+# and we only want to perform this cleanup once, in case
+# an admin later intentionally changes home dir permissions.
+ConditionPathExists=!/var/eos-dir-mode-700
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+StandardOutput=journal+console
```
dbnicholson: I don't think this setting is needed. It's added in some of the other services here so that you get logging on the console. That's because some of the other services run early and have real effects on boot. If something goes wrong, then maybe you can see something on the console.

This service runs much later when the journal will be normally accessible and messages can be gathered from there.

eos-fix-home-dir-permissions
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -e
+
+# Prior to 3.1.6, home directory permissions were created with dir mode 755.
+# This is now fixed, but on upgrade from an earlier version to 3.1.6,
+# let's perform a one-time clean up of any existing home dirs.
+
+stamp_file=/var/eos-dir-mode-700
```
ramcq: Hmm... stampy stampy. I'd go as far as to say /var/lib/misc because /var/lib is arranged under a hierachy too.

Not sure if it's a red herring seeing as this is actually quite neat with a ConditionPathExists, but can you not check the permissions of /home/shared as the only state you need to determine whether this job needs to run? I can't immediately see a way to hook that to systemd so it would be "strictly worse" unfortunately, in that the job would need to run to decide if it needed to run or not, so would add to the default boot path.

(I do wonder if we could do better with systemd units which need to be run when upgrading from a version before X, and keep one file with the version in a fixed location and do a monotonic check, trigger an events which other systemd tasks can bind to, then update the file... hmm...)

rshuler: Agreed with /var/lib/misc.  I was following the lead of other files that we had in /var, but yes it makes sense to stop polluting this top level directory.

I actually considered checking the permissions of /home/shared, though in theory an admin could decide to change those after the fact (as maybe they actually want the shared user's documents available to other users).  So, I think a stamp file is better -- plus it makes for a simple check that the service turns into a no-op.  (Is there a way for the service to check based on permissions and turn it into a no-op?)

eos-fix-home-dir-permissions.service
rshuler: Will do

eos-fix-home-dir-permissions.service
rshuler: Okay, thanks.  Yes, I just saw that in another file and thought it might be useful.

eos-fix-home-dir-permissions
rshuler: I see that I have a /var/lib/misc, but I don't know if it is guaranteed to exist.  To be on the safe side, I assume it is best to `mkdir -p` it.

PR #108, opened 2017-05-08
add eos-migrate-image-defaults script & service

ramcq:
Simple systemd unit and script to check if /var/eos-image-defaults
is a directory (and not a symlink), move it to /var/lib, and make
a compatibility symlink.

https://phabricator.endlessm.com/T15704

### DIFF
--- a/eos-migrate-image-defaults
+++ b/eos-migrate-image-defaults
@@ -1,0 +1,60 @@
+#!/bin/bash -ex
+
+migration_stamp=/var/lib/.eos-migrate-image-defaults
+
+old_path=/var/eos-image-defaults
+new_path=/var/lib/eos-image-defaults
+
+dconf_profile_etc=/etc/dconf/profile/user
+dconf_profile_usr=/usr/share/dconf/profile/user
+
+# create the migration stamp if the old path exists (not a symlink)
+# and the new path does not exist. the process can be safely re-run
+# until it's successful, and will be re-tried until all steps have
+# succeeded and the stamp will be removed at the end.
+if [ -d "${old_path}" -a ! -h "${old_path}" -a ! -e "${new_path}" ]; then
+  touch "${migration_stamp}"
+fi
+
+if [ -f "${migration_stamp}" ]; then
+  # for safety, add the new path to the legacy dconf file during the migration,
+  # in case there is any interruption. this will cause errors due to one path
+  # not existing, but ensure the settings are respected until the migration is
+  # fully completed (eg on next reboot).
+  if [ -f "${dconf_profile_etc}" ] &&
+     ! grep -qs "${new_path}" "${dconf_profile_etc}"; then
+    sed -i "s|^file-db:${old_path}/settings$|file-db:${new_path}/settings\nfile-db:${old_path}/settings|" "${dconf_profile_etc}"
+  fi
+
+  # move the directory if it still needs moving (same conditions as above).
+  if [ -d "${old_path}" -a ! -h "${old_path}" -a ! -e "${new_path}" ]; then
+    mv "${old_path}" "${new_path}"
+  fi
+
+  # now the folder is moved, we can remove the old path from the dconf profile.
+  if [ -f "${dconf_profile_etc}" ]; then
+    sed -i "\|^file-db:${old_path}/settings$|d" "${dconf_profile_etc}"
+
+    # and if it's now the same as the system default one, we can safely delete it.
+    # otherwise there is a local admin customisation we should preserve.
+    if diff "${dconf_profile_etc}" "${dconf_profile_usr}"; then
+      rm -f "${dconf_profile_etc}"
+    fi
+  fi
+
+  # the creation of the compatibility symlink completes the migration
+  ln -sfT "${new_path}" "${old_path}"
+
+  # remove the stamp so that the migration is not attempted again
+  rm -f "${migration_stamp}"
+fi
+
+# the settings db has been mandatory since before 3.0, but upgrades from 2.x
+# won't have it, so the folder may not exist. ensure that it does, and that
+# it contains an empty settings override file.
+if [ ! -f "${new_path}/settings" ]; then
+  mkdir -p "${new_path}"
+  tmpdir=$(mktemp -d)
+  dconf compile "${new_path}/settings" "${tmpdir}"
+  rmdir "${tmpdir}"
+fi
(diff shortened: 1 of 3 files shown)


eos-migrate-image-defaults
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -ex
+
+old_path=/var/eos-image-defaults
+new_path=/var/lib/eos-image-defaults
+dconf_profile=/etc/dconf/profile/user
+
+if [ -d ${old_path} -a ! -h ${old_path} -a ! -e ${new_path} ]; then
```
cosimoc: Nitpick: I find the early-return style a lot more readable for these cases.

eos-migrate-image-defaults
```diff
@@ -0,0 +1,17 @@
...
+  mv ${old_path} ${new_path}
+fi
+
+if [ ! -e ${old_path} -a -d ${new_path} ]; then
+  ln -s ${new_path} ${old_path}
+fi
+
+# unfortunately the /etc/dconf/profile/user is not contained in the
+# ostree, so must be manually updated when this migration is made
```
cosimoc: I don't think we use this file anymore since we moved to the new GDM that installs it in `/usr`? The overrides created by the image builder are in a subdirectory of the image-defaults location AFAIR, so what else is in this file?

eos-migrate-image-defaults.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Migrate eos-image-defaults to /var/lib
+ConditionPathIsDirectory=/var/eos-image-defaults
+ConditionPathIsSymbolicLink=!/var/eos-image-defaults
```
cosimoc: You also could add `ConditionNeedsUpdate=/var` here and only run this on upgrades.

eos-migrate-image-defaults
ramcq: Ah - because we're doing 3 operations which aren't atomic, it's intentionally written to not early return so that it's idempotent, so that if it's interrupted at any stage, running it again will complete the migration - so if it gets messed up somehow (admittedly unlikely), you can just run the script again. An early return would prevent the later parts from running.

eos-migrate-image-defaults
ramcq: The image builder still creates this /etc/dconf/profile/user on every image: https://github.com/endlessm/eos-image-builder/blob/master/hooks/image/62-dconf-profile

The contents of the overrides are in the endless-image-defaults location, but this Endless-specific dconf user profile is needed to make dconf look in that path. If this Endless-specific path is included in dconf somewhere else now, then a) I didn't find it, because I searched the whole organisation for endless-image-defaults when preparing these PRs, and b) we should remove that and arrange for the file in /etc to be removed (maybe in fact make the existence of this file the trigger for this script).

eos-migrate-image-defaults.service
ramcq: Hmm - I don't think that's better necessarily? These conditions are pretty specifically-focused to make sure this task runs in the exact condition it needs to, and not otherwise. I think you only need ConditionNeedsUpdate if you can't write more specific conditions and need an external reference to ensure your upgrade job gets run. I don't mind either way, but if we did it using ConditionNeedsUpdate I'd remove the other tasks (and we'd need Before=systemd-updates-done or whatever it's called).

ramcq:
Have had a poke around and I think the best plan is this:
* ship a /usr/share/dconf/profile/user file in the ostree (we could do it from eos-boot-helper, eos-default-settings, or as a debian packaging tweak in d-conf) adding the /var/lib path
* drop the hook in the image builder that writes /etc/dconf/profile/user
* make this job run if /etc/dconf/profile/user and /usr/share/dconf/profile/user both exist, and just remove the former
* update eos-build (the Photos flatpak .json) to map /usr/share/dconf/profile/user as well as /etc/dconf/profile/user (it's the only Flatpak we have which actually seems to be able to access the /var/eos-image-defaults/settings anyway, so I guess it's the only one which we try to override in the image builder)

Thoughts? Relies on dconf silently ignoring paths from the profiles that don't exist.

eos-migrate-image-defaults
cosimoc: Makes sense!

eos-migrate-image-defaults
cosimoc: OK, your explanation here makes perfect sense. I will follow up to the other side of this in the other thread.

eos-migrate-image-defaults.service
cosimoc: I am thinking about how these conditions here interact with the non-atomicity of the operations you describe in the other comment.

From https://www.freedesktop.org/software/systemd/man/systemd.unit.html
```
If multiple conditions are specified, the unit will be executed if all of them apply (i.e. a logical AND is applied).
```

So the script will run whenever `/var/eos-image-defaults` is both a directory and not a symlink. If it gets interrupted after the first `mv`, the next boot it will not be a directory anymore, but it's not yet a symlink, leaving things in a broken state.

The documentation also says
```
Condition checks can be prefixed with a pipe symbol (|) in which case a condition becomes a triggering condition. If at least one triggering condition is defined for a unit, then the unit will be executed if at least one of the triggering conditions apply and all of the non-triggering conditions.
```

So perhaps we can simply make those triggering conditions instead.

That still won't solve the case when the script is interrupted after `ln` but before `sed`. For that reason, I believe that you should run `sed` at the beginning of the script instead (or at least before the last triggering condition).

cosimoc:
Me and @ramcq also discussed how the migration of the dconf profile to `/usr`, while very desirable is not that straight forward, since dconf will constantly try to open file-db profiles when missing (see https://git.gnome.org/browse/dconf/tree/engine/dconf-engine-source-file.c#n44). One option is to always generate the dconf db from the image builder, even when empty, but we can also defer that aspect to a later ticket.

ramcq:
Alright, I think I've sorted it out now... :) The script is pretty well-commented so it should make sense, but basically because the dconf hooks are enabled by default in the image builder, we can rely on the "settings" file-db existing other than when the system is upgraded from 2.x. So I think it makes sense for us to just guarantee /var/lib/eos-image-defaults/settings exists, and then we can add the profile/user into /usr/share/dconf in the ostree - I just need to know which package you think it should live in? Either we go for eos-default-settings, but that's kind of a bonus side-package that comes with eos-theme, or we could add it in the debian/ packaging for d-conf?

The script adds the new path to the profile while it's moving folders around and stuff, then removes the old path and deletes the /etc/dconf/profile/user if it ends up identical to the system one, so that any local modifications are preserved. The script makes the compatibility symlink as it's last action, the absence of which triggers the systemd job.

PR #109, opened 2017-06-12
Add Endless apps flatpak repo after an upgrade if needed

ptomato:
https://phabricator.endlessm.com/T17172


eos-add-flatpak-apps-repos.service
```diff
@@ -0,0 +1,20 @@
...
+[Unit]
+Description=Configure the stable Endless apps flatpak repositories
+DefaultDependencies=no
+Conflicts=shutdown.target
+Wants=local-fs.target
+After=local-fs.target
+
+# Only run on updates
+Before=multi-user.target systemd-update-done.service
+ConditionNeedsUpdate=/etc
```
cosimoc: I think technically this should be `ConditionNeedsUpdate=/var` because that's where the repo config lives. (It does not matter much in practice because right now they're updated at the same time, but would be nice to get it right.)

eos-add-flatpak-apps-repos
```diff
@@ -0,0 +1,35 @@
...
+# can just hardcode here the contents of the .flatpakrepo file
+# so that we don't need to rely on a network connection.
+
+# Endless apps runtime Flatpak repository
+cat <<EOF > ${TMPDIR}/eos-sdk.flatpakrepo
+[Flatpak Repo]
+Title=Endless Apps Runtimes
+Url=FIXME https://ostree.endlessm.com/ostree/eos-sdk
+Homepage=http://endlessm.github.io/eos-knowledge-lib/
+Comment=Flatpak runtime for developing offline content apps for Endless
```
cosimoc: I think I would remove the mention of "offline" here.

ptomato: I thought "offline content" was our buzzword.

cosimoc: "asynchronous internet" and "limited connectivity" too :-)
My point is more that this runtime is not only useful for offline apps, it does support creating apps that go online too.
Either way, let's see if @nedrichards wants to participate in the bikeshedding here ;-)

nedrichards: I'd say that Endless apps should inherently express our values and respect the user with offline support, async internet and all that good stuff. So I'd just say it's a: "Flatpak runtime for developing apps for Endless OS."

eos-add-flatpak-apps-repos
```diff
@@ -0,0 +1,35 @@
...
+# This information for setting up this repo is stable, so we
+# can just hardcode here the contents of the .flatpakrepo file
+# so that we don't need to rely on a network connection.
+
+# Endless apps runtime Flatpak repository
+cat <<EOF > ${TMPDIR}/eos-sdk.flatpakrepo
+[Flatpak Repo]
+Title=Endless Apps Runtimes
+Url=FIXME https://ostree.endlessm.com/ostree/eos-sdk
+Homepage=http://endlessm.github.io/eos-knowledge-lib/
```
nedrichards: HTTPS by default, surely

ebassi:
Looks good to me, and the issues raised from @cosimoc and @nedrichards have been addressed.

@cosimoc: Can I get a :+1:?

cosimoc:
@ebassi Yep, feel free to push this!

PR #110, opened 2017-06-16
image-boot: support squashfs on exFAT

wjt:
I'm working on making the Installer for Windows operate on ISOs rather than on raw disk images, and I'm hoping to improve the USB stick creation process at the same time.

For starters: many users are surprised that they need a 32GiB USB stick for many Full images, despite the (compressed) download size being less than 16G(i)B. Even if we keep the current approach of putting the image file within a big exFAT partition on the USB stick (which is good, because it keeps the free space on the USB stick usable), it'd be nice if we could put the squashfs from inside the ISO onto that exFAT partition.

It'd also be nice to be able to reuse more of Rufus's USB stick creation code. Ours is a bit unreliable on Windows; of course, all bugs can be fixed, but if we can just lean on Rufus' (presumably) better-tested code this would be less work in the long run, and might even allow us to get patches into upstream Rufus so it can operate on our ISOs in non-`dd` mode.

I've tested this against the latest `eos3.2` image and it works, both with and without squashfs, because @carlocaione wisely made the embedded GRUB config search for `/endless/grub/grub.cfg` rather than `/endless/endless.{img,squash}` so the same GRUB blobs work.

https://phabricator.endlessm.com/T15890
https://phabricator.endlessm.com/T14312

### DIFF
--- a/dracut/image-boot/Makefile.am
+++ b/dracut/image-boot/Makefile.am
@@ -1,9 +1,10 @@
 dracutmoddir = $(prefix)/lib/dracut/modules.d/50eos-image-boot
 dist_dracutmod_SCRIPTS = \
-	module-setup.sh \
-	eos-image-boot-setup \
 	eos-image-boot-generator \
+	eos-image-boot-setup \
 	eos-live-early-overlayfs-setup \
+	eos-map-image-file \
+	module-setup.sh \
 	$(NULL)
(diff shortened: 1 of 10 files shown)


dsd:
Looks good, thanks a lot for putting the effort into the unit tests too!

PR #115, opened 2017-07-01
Install com.endlessm.EknServices from eos-sdk

ptomato:
When adding the new eos-sdk repo after an upgrade, make sure the only
installed copy of EknServices is the one published there. (We assume
that there is an internet connection since this script runs when we just
completed an OS upgrade.)

We remove any orphaned refs belonging to the old
com.endlessm.EknServices, though normal users would not have these
anyway.

https://phabricator.endlessm.com/T17863

### DIFF
--- a/eos-add-flatpak-apps-repos
+++ b/eos-add-flatpak-apps-repos
@@ -34,23 +34,6 @@ flatpak remote-add --if-not-exists --from eos-sdk ${TMPDIR}/eos-sdk.flatpakrepo
 
 # Change com.endlessm.EknServices over to the version that uses the new runtime
 # in the eos-sdk repo
-cat <<EOF >${TMPDIR}/switch-eknservices.py
-import gi
-gi.require_version('Flatpak', '1.0')
-from gi.repository import Flatpak
-
-installation = Flatpak.Installation.new_system(cancellable=None)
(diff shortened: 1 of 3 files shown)


eos-add-flatpak-apps-repos
```diff
@@ -32,4 +32,11 @@ EOF
 # Add the repo only if not added yet
 flatpak remote-add --if-not-exists --from eos-sdk ${TMPDIR}/eos-sdk.flatpakrepo
 
+# Change com.endlessm.EknServices over to the version that uses the new runtime
+# in the eos-sdk repo
+flatpak install eos-sdk com.endlessm.EknServices//eos3
```
cosimoc: Doing `flatpak install` at boot time is pretty bad -- we should avoid it at all costs if we can.
Can't you do some surgery and change the origin on the existing one to pull from the new repo instead? Then gnome-software would pull the update normally from there.

dbnicholson:
This already got merged, but I would have preferred to not have the deploy file mangling script be in PATH. That operation can easily break your app and now we just gave you a tool to do it for all your apps apps if you want. This is done with no locking or coordination with other flatpak clients.

That should really be a targeted operation that's not publicly exposed. I honestly would have just rewritten the 5 lines of bash into python and copied the function into it.

PR #116, opened 2017-07-03
Use the right directory for the repo configuration

ebassi:
We need to update /etc, not /var, for the repository configuration.

https://phabricator.endlessm.com/T17172


ramcq:
Probably not the right fix...

PR #117, opened 2017-07-04
Consolidate eos-sdk flatpak remote script

ptomato:
We don't want to leave a tool such as eos-flatpak-change-origin lying
around in the PATH, because you can very easily break your apps. Instead,
put everything in the same Python script and make it specific to the one
app whose origin we want to change.

https://phabricator.endlessm.com/T17863


ramcq:
Huh... I kinda semi-intentionally left it as a separate binary because I thought it was a pretty handy tool. We've now got two copies of this function in the OS, so if we don't export it somewhere we might just end up with a 3rd and a 4th... Of course it's not necessarily the best place/way to export it (like we could put it in libexec or something?), but I wasn't hugely unhappy with it being available as a standalone thing.

ptomato:
I'm fine with it being either way — this is definitely not the only way you can break your system with an `eos-` script — but I'll leave you to sort that out with @dbnicholson since he asked for this :smile:

ramcq:
I entirely appreciate his argument too, so I'm fine with it being either way. More relevantly for this PR I guess: we should make the function return early and do nothing if the ref already matches.

ptomato:
Early return added.

PR #126, opened 2017-08-04
eos-copy-host-mali-driver: Put GL driver in extension

ptomato:
This symlinks the host GL Mali driver in
/usr/lib/arm-linux-gnueabihf/mali400-egl-x11 into a place where Flatpak
will hopefully look for it and mount it as an extension. This makes the
separate org.freedesktop.Platform.GL.mali450-blah-blah extension
unnecessary, as we can simply use the host's driver.

### DIFF
--- a/eos-link-host-mali-driver.service
+++ b/eos-link-host-mali-driver.service
@@ -1,0 +1,22 @@
+# Symlink the host GL driver for Mali to a place where Flatpak will mount it
+
+[Unit]
+Description=Symlink Mali GL driver for Flatpak
+DefaultDependencies=no
+Conflicts=shutdown.target
+Wants=local-fs.target
+After=local-fs.target
+Before=multi-user.target
+
+# Only run on ARM systems with the Mali driver where we haven't run before
+ConditionArchitecture=arm
+ConditionPathExists=/usr/lib/arm-linux-gnueabihf/mali400-egl-x11
+ConditionPathExists=!/var/lib/flatpak/extension/org.freedesktop.Platform.GL.host/arm/1.6
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=/usr/sbin/eos-link-host-mali-driver
+
+[Install]
+WantedBy=multi-user.target
--- a/Makefile.am
+++ b/Makefile.am
@@ -20,6 +20,7 @@ dist_systemdunit_DATA = \
 	eos-fix-flatpak-branches.service \
 	eos-fix-home-dir-permissions.service \
 	eos-image-boot-dm-setup.service \
+	eos-link-host-mali-driver.service \
 	eos-live-boot-overlayfs-setup.service \
 	eos-migrate-image-defaults.service \
 	eos-remove-legacy-external-apps.service \
@@ -75,6 +76,7 @@ dist_sbin_SCRIPTS = \
 	eos-fix-flatpak-branches \
 	eos-fix-home-dir-permissions \
 	eos-image-boot-dm-setup \
+	eos-link-host-mali-driver \
 	eos-live-boot-overlayfs-setup \
 	eos-migrate-image-defaults \
 	eos-remove-legacy-external-apps \
--- a/eos-link-host-mali-driver
+++ b/eos-link-host-mali-driver
@@ -1,0 +1,35 @@
+#!/bin/sh -ex
+#
+# eos-link-host-mali-driver: Symlink the host's GL drivers into a flatpak local
+# extension directory.
+#
+# Copyright (C) 2017 Endless Mobile, Inc.
+# Authors:
+#  Philip Chimento <philip@endlessm.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+EXTENSION_REF=org.freedesktop.Platform.GL.host/arm/1.6
+MALI_DIR=/usr/lib/arm-linux-gnueabihf/mali400-egl-x11
+
+DEST_DIR="/var/lib/flatpak/extension/$EXTENSION_REF"
+mkdir -p $(dirname "$DEST_DIR")
+ln -s "$MALI_DIR" "$DEST_DIR"
+
+if test -d /var/endless-extra/flatpak; then
+    DEST_DIR="/var/endless-extra/flatpak/extension/$EXTENSION_REF"
+    mkdir -p $(dirname "$DEST_DIR")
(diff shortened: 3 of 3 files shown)


eos-copy-host-mali-driver
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
```
ramcq: set -ex is probably wise

eos-copy-host-mali-driver
```diff
@@ -0,0 +1,31 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+EXTENSIONS_DIR=/var/lib/flatpak/extension
+if test -d /var/endless-extra/flatpak; then
+    EXTENSIONS_DIR=/var/endless-extra/flatpak/extension
+fi
+
+DEST_DIR="$EXTENSIONS_DIR/org.freedesktop.Platform.GL.mali450/arm/1.6/lib"
+mkdir -p "$DEST_DIR"
+ln -s /usr/lib/arm-linux-gnueabihf/mali400-egl-x11 "$DEST_DIR"
```
ramcq: I am suspicious whether Flatpak will resolve the symlink here - I think it follows symlinks when the source of a --bind is a symlink, so it gets the real directory underneath, but I doubt it would do any recursing - so the most likely outcome here is that we have an extension containing a dangling symlink to a path that only exists outside the chroot. In the case that the flatpak stuff is in /var, we can cp -rl this (and hence hardlink the driver) but if it's in /var/endless-extra we have to fall back to cp -r.

eos-copy-host-mali-driver.service
```diff
@@ -0,0 +1,21 @@
...
+DefaultDependencies=no
+Conflicts=shutdown.target
+Wants=local-fs.target
+After=local-fs.target
+Before=multi-user.target
+
+# Only run on ARM systems with the Mali driver
+ConditionArchitecture=arm
+ConditionPathExists=/usr/lib/arm-linux-gnueabihf/mali400-egl-x11
+
```
ramcq: We can also add ConditionNeedsUpdate=/var so that this is only re-run after an OS upgrade. In this case I think we'd need to also do the copy in the image builder (or run the script in a chroot hook) as I'm not sure the update condition is true on first boot.

eos-copy-host-mali-driver
ptomato: I was assuming that it would be OK because of https://github.com/flatpak/flatpak/blob/2082b508719999c4c7f2decbfd958943948c0b21/common/flatpak-dir.c#L6432, which eventually gets passed to `flatpak_extension_new()` in https://github.com/flatpak/flatpak/blob/e40c736b3bb0b9d1eff6399642e02bcc27a09a9a/common/flatpak-utils.c#L3975. If a symlink is possible, I think it would be better, since then we really only have to run the service once ever. But, you may be right that the symlink ends up dangling because of being outside the sandbox...

ptomato:
Thanks for the review! I'm out of time for today but I'll take another look on Monday or possibly before.

eos-copy-host-mali-driver.service
ptomato: Could do that, but it seems to make it more complicated than necessary; even if run redundantly, this script should just notice that the files are already there and then exit, right?

ptomato:
Copying code is on https://github.com/endlessm/eos-boot-helper/tree/T18288-try2 — I think the only thing needed is to change `ln -s` to `cp -rl`, right?

eos-copy-host-mali-driver
```diff
@@ -0,0 +1,31 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+EXTENSIONS_DIR=/var/lib/flatpak/extension
+if test -d /var/endless-extra/flatpak; then
+    EXTENSIONS_DIR=/var/endless-extra/flatpak/extension
+fi
+
+DEST_DIR="$EXTENSIONS_DIR/org.freedesktop.Platform.GL.host/arm/1.6/lib"
+mkdir -p "$DEST_DIR"
```
ramcq: if you mkdir including the /lib then the ln -s command puts a mali400 dir _inside_ the lib dir - you should take /lib off the DEST_DIR and ln -s to DEST_DIR/lib

eos-copy-host-mali-driver.service
ramcq: yeah but as you have it now, it will run fork the shell needlessly every boot, adding to boot delay and the boot IO storm... we really want another condition which looks for whether the link exists. maybe if we create the link in both /var/lib and /var/endless-extra we can use the existence of the former to check whether we should be running?

(and I still think creating the link in the image builder is worthwhile.)

eos-copy-host-mali-driver
ptomato: I thought that's what it should be; the driver is in a `mali400-blahblah` subdir on the host system as well. But maybe flatpak is not looking for it there? I'll change it.

eos-copy-host-mali-driver.service
ptomato: OK, will do.

eos-copy-host-mali-driver
ramcq: I doubt whatever we have to add that subdir to the LD_LIBRARY_PATH exists inside the Flatpak chroot.

That said, I can't find what makes it look in /lib/GL/<foo> either, but I didn't look hard.

eos-copy-host-mali-driver
```diff
@@ -0,0 +1,35 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+EXTENSION_REF=org.freedesktop.Platform.GL.host/arm/1.6
+MALI_DIR=/usr/lib/arm-linux-gnueabihf/mali400-egl-x11
+
+DEST_DIR="/var/lib/flatpak/extension/$EXTENSION_REF"
+mkdir -p "$DEST_DIR"
+ln -s "$MALI_DIR" "$DEST_DIR/lib"
```
ramcq: As mentioned on Slack, what happens is flatpak binds DEST_DIR directly to /usr/lib/GL/host inside the flatpak, and sets the path to /usr/lib/GL/host/lib. So we need to create the parent of DEST_DIR, have DEST_DIR itself be the link to MALI_DIR, and the link in the ostree from lib -> . will put the libfoo*.so onto the library path.

As you're revising this PR, can I be a pain and have you rename all of this to eos-link-host-gl-driver because I expect @dsd will add nvidia to this mechanism for amd64 during 3.3. :)

ramcq: Actually, bikeshedding myself, host-mali-driver is OK. just link is better than copy. :)

eos-link-host-mali-driver.service
```diff
@@ -0,0 +1,22 @@
...
+[Unit]
+Description=Symlink Mali GL driver for Flatpak
+DefaultDependencies=no
+Conflicts=shutdown.target
+Wants=local-fs.target
+After=local-fs.target
+Before=multi-user.target
+
+# Only run on ARM systems with the Mali driver where we haven't run before
+ConditionArchitecture=arm
```
rshuler: Not that it matters for our current devices, but looking at the other services we have in eos-boot-helper, I suspect we may want to replace this with:
```
ConditionArchitecture=|arm
ConditionArchitecture=|arm64
```

I'm going to merge as is, but @ramcq please confirm whether or not this would be needed on 64-bit ARM.

PR #127, opened 2017-08-16
Explicitly setup /sysroot/boot -> /boot bind mount

jprvita:
On OSTree systems, this is manually setup in the initrd by
ostree-prepare-root. But since it is not configured anywhere in the
system, systemd-gpt-auto-generator is not aware of it, and ends up
un-mounting it due to its own boot.mount and boot.automount units,
generated to mount the ESP on /boot.

We don't need to mount the ESP on EOS, and making this bind mount
explicit avoids this systemd-gpt-auto-generator's behaviour.

This effectively reverts commit
ccb3e26770ebbc7709a689497e7fbd59175e2638.

https://phabricator.endlessm.com/T18825

### DIFF
--- a/boot.mount
+++ b/boot.mount
@@ -1,0 +1,11 @@
+[Unit]
+ConditionKernelCommandLine=ostree
+
+[Mount]
+What=/sysroot/boot
+Where=/boot
+Type=none
+Options=bind
+
+[Install]
+WantedBy=local-fs.target
--- a/Makefile.am
+++ b/Makefile.am
@@ -9,6 +9,7 @@ AM_DISTCHECK_CONFIGURE_FLAGS = \
 	$(NULL)
 
 dist_systemdunit_DATA = \
+	boot.mount \
 	eos-add-flatpak-apps-repos.service \
 	eos-add-flatpak-gnome-repos.service \
 	eos-enable-extra-upgrade.service \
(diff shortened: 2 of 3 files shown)


boot.mount
```diff
@@ -0,0 +1,11 @@
...
+ConditionKernelCommandLine=ostree
+
+[Mount]
+What=/sysroot/boot
+Where=/boot
+Type=none
+Options=bind
+
+[Install]
+WantedBy=local-fs.target
```
dbnicholson: Doesn't this need the `dh_systemd_enable boot.mount` in `debian/rules`? Or is the presence of `boot.mount` enough for systemd to not generate the mount unit?

jprvita: We don't need to list all units on the `dh_systemd_enable` call after @wjt's ba6e6c98c5be32076ca82880db3f9321478db14c.

jprvita:
Also, please mind the discussion whether this should be done in a initrd unit instead of this unit + ostree's `ostree-prepare-root`: https://phabricator.endlessm.com/T18825#469318

boot.mount
dbnicholson: Oh, nice! I didn't see that.

jprvita:
This was fixed in systemd-gpt-auto-generator.

PR #128, opened 2017-08-16
Add Flathub repo if needed after an OS upgrade

rshuler:
https://phabricator.endlessm.com/T18838


wjt:
Is it necessary to have this as a separate script? We now have three scripts to add three repos if missing, which will all run at boot after every update.

rshuler:
Since it only runs after an update, I'm not terribly concerned with whether it runs as 3 scripts rather than 2 (or even 1).  Is there a reason for concern?  FWIW, we'll eventually be removing the GNOME repo once all the GNOME apps are in Flathub.  Perhaps if we set that up as a forcible remove, it might make sense to merge into a single script to handle all flatpak repo config cleanup.

PR #129, opened 2017-08-30
Clean up the legacy external apps remote and runtimes

joaquimrocha:
Since we no longer use the external apps implementation we did for
pre-3.1 releases, we should make sure that our users are not left
with runtimes occupying space, and the external apps remote just
resulting in potential bugs.

https://phabricator.endlessm.com/T14442


joaquimrocha:
Sorry. Forgot to include it in the commit...
PR updated.

PR #130, opened 2017-08-31
Import eos.preset from systemd package

jprvita:
We are currently wasting resources re-building systemd and
re-distributing all of systemd packages every time we change the preset
file. Let move it to a smaller package.

https://phabricator.endlessm.com/T11144

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -12,6 +12,9 @@ m4_define([no_systemd_error],
 m4_define([no_systemdsystemgeneratordir_error],
           [m4_normalize([Could not get systemdsystemgeneratordir setting from]
                         [systemd pkg-config file])])
+m4_define([no_systemdsystempresetdir_error],
+          [m4_normalize([Could not get systemdsystempresetdir setting from]
+                        [systemd pkg-config file])])
 m4_define([no_udev_error],
           [m4_normalize([Could not get udevdir setting from udev]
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -29,6 +32,13 @@ AC_ARG_WITH([systemdgeneratordir],
             [PKG_CHECK_VAR([systemdgeneratordir], [systemd],
                            [systemdsystemgeneratordir], [],
                            [AC_MSG_ERROR(no_systemd_error)])])
+AC_ARG_WITH([systemdpresetdir],
+            [AS_HELP_STRING([--with-systemdpresetdir],
+                            [Path to the system directory for systemd presets])],
+            [systemdpresetdir="$withval"],
+            [PKG_CHECK_VAR([systemdpresetdir], [systemd],
+                           [systemdsystempresetdir], [],
+                           [AC_MSG_ERROR(no_systemd_error)])])
```
dbnicholson: Shouldn't this be the `no_systemdsystempresetdir_error` you defined above?

jprvita:
Good catch! I had actually based my changes on the configure switch for the generator dir, which also has this problem. I have fixed my commit, and add two other commits to the branch, fixing the problem for the generator dir, and renaming the error message variable for the unit dir, to make it clear that it isn't a general systemd error.

PR #131, opened 2017-09-06
live-boot: allow passing a list of extra overlay mounts

cosimoc:
This will be used by test/demo modes to also overlay /etc.

https://phabricator.endlessm.com/T18999

### DIFF
--- a/eos-live-boot-overlayfs-setup
+++ b/eos-live-boot-overlayfs-setup
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+EXTRA_PATHS=$@
+
 # Allow xattrs in the user namespace on /run, and by extension in the ostree
 # repo, where they are used to record permissions when run as an unprivileged
 # user.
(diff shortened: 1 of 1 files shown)


eos-live-boot-overlayfs-setup
```diff
@@ -47,7 +49,7 @@ setup_ostree_flatpak_overlay() {
 
 # Everything but /sysroot/{ostree,flatpak} is pretty straightforward:
 overlay_dirs="bin boot endless home lib opt ostree root sbin srv sysroot/home var"
-for dir in $overlay_dirs; do
+for dir in $overlay_dirs ${EXTRA_PATHS[@]}; do
```
wjt: I don't think `$EXTRA_PATHS` is actually an array as defined here, so the `${EXTRA_PATHS[@]}` syntax has no effect:

```console
$ x=(a b); for y in ${x[@]}; do echo $y; done
a
b
$ cat x
#!/bin/bash
EXTRA_PATHS=$@
for dir in ${EXTRA_PATHS[@]}; do echo $dir; done
$ ./x "a b"
a
b
```

I think you want either:

```bash
EXTRA_PATHS=( "$@" )
for dir in $overlay_dirs "${EXTRA_PATHS[@]}"; do
```

or in the likely case that we really do not care about paths with spaces in them:

```bash
EXTRA_PATHS=$@
for dir in $overlay_dirs $EXTRA_PATHS
```

PR #137, opened 2017-09-22
eos-link-host-mali-driver: add libMali wrapper to fix missing symbol

xjuan:
Our mali driver is compiled against a patched X11 library that has a
new symbol (XDisplaySetErrorHandler) which will not be present in
flatpak runtimes

Since we can not currently rebuild the driver the only solution is to add
a wrapper library that links with the driver and implements the missing symbol

### DIFF
--- a/eos-link-host-mali-driver
+++ b/eos-link-host-mali-driver
@@ -22,7 +22,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 EXTENSION_REF=org.freedesktop.Platform.GL.host/arm/1.6
-MALI_DIR=/usr/lib/mali400/arm-linux-gnueabihf
+MALI_DIR=/usr/lib/mali400/arm-linux-gnueabihf/flatpak
 
 DEST_DIR="/var/lib/flatpak/extension/$EXTENSION_REF"
 mkdir -p $(dirname "$DEST_DIR")


eos-link-host-mali-driver
```diff
@@ -22,14 +22,75 @@
...
+    for LIB in libEGL.so.1.4 libGLESv2.so.2 libGLESv2.so.2.0; do
+      ln $DEST_DIR/libEGL.so.1 $DEST_DIR/$LIB
+      chmod +x $DEST_DIR/$LIB
+    done
+}
+
+# Lookup mali driver
+if test -f /usr/lib/mali400/arm-linux-gnueabihf/libMali.so; then
+    MALI_DRIVER="/usr/lib/mali400/arm-linux-gnueabihf/libMali.so"
+elif test -f /usr/lib/arm-linux-gnueabihf/mali400-egl-x11/libMali.so; then
```
jprvita: We are assuming the new path on master on the unit file that runs this script, so this will never be true. I believe it is safe and simpler to assume the new location on master, but that means you'll need to adjust the path on the fix for eos3.2. The link should be updated on upgrade (before this change).

eos-link-host-mali-driver
```diff
@@ -22,14 +22,75 @@
...
+
+    # Create lib link for flatpak GL extension
+    ln -sr $DEST_DIR $DEST_DIR/lib
 
-if test -d /var/endless-extra/flatpak; then
-    DEST_DIR="/var/endless-extra/flatpak/extension/$EXTENSION_REF"
-    mkdir -p $(dirname "$DEST_DIR")
-    ln -sf "$MALI_DIR" "$DEST_DIR"
+    # Hardlink mali driver
+    ln -t $DEST_DIR $MALI_DRIVER
```
jprvita: Why are you hardlinking instead of symlinking? I've previously added `-f` to the symlink to make sure the link gets overwritten and points to the new target. I don't know how that works with hardlinks.

eos-link-host-mali-driver
```diff
@@ -22,14 +22,75 @@
...
+        return;
+    fi
+
+    DEST_DIR=$FLATPAK_DIR/extension/$EXTENSION_REF
+
+    # Ensure destination directory
+    mkdir -p $DEST_DIR
+
+    # Create lib link for flatpak GL extension
+    ln -sr $DEST_DIR $DEST_DIR/lib
```
jprvita: This link is already exists in the host ostree -- it is created by the mali package.

eos-link-host-mali-driver
```diff
@@ -22,14 +22,75 @@
...
-if test -d /var/endless-extra/flatpak; then
-    DEST_DIR="/var/endless-extra/flatpak/extension/$EXTENSION_REF"
-    mkdir -p $(dirname "$DEST_DIR")
-    ln -sf "$MALI_DIR" "$DEST_DIR"
+    # Hardlink mali driver
+    ln -t $DEST_DIR $MALI_DRIVER
+
+    # Create mali driver wrapper with missing symbol
+    # (Binary generated from libMaliWrapper.c)
+(   base64 -d <<EOF
```
jprvita: Can't we ship this in the ostree as well, through the mali package?

eos-link-host-mali-driver
```diff
@@ -22,14 +22,75 @@
...
+3gc1cjSTXkn99pDER1IQ29qMqtKRxl62JqhGauvhIvqLwwJ6EBLiL6yi5xqNBq0ljffZKu859NXt
+GomwfrsBrE+9/l0X2RyIIXoQBxC9iB+EpzTjfyxFfDyRtoR/Y3Z6+rVobIaWo21H4/Kk/OLo+Fg5
+XS445UmdHcfTen78GBPCFO0uTdfSOSDH6Qe9MjoFLe2tUxJpE9rFbrGHHCS9oT7STw4Jh0MDZFB4
+noBsG7ZjOVoaZEOzsVKySwV7Kc+oY4Fs6aZcMh2QsV0c91jL5zIgv1tEDise5Ewxn9cLyE2dn5c1
+x7Fyaewt+3/EooPHX+C5oHiT88BzJPJ4d/BzmiMKOhRCPD8iz1U/5109geGG7z0v8keAzdYEsBxS
+xPg9bwbT+TvE77T4crXgs+f5d9ynR+cwxQB5Uo+uMa5DfWT/E/YvCfn06LuT3J5Xe+soiDZ593Wu
+T/XoPH0gsu8IvjvN7Y1xnuqNB+JCMcdslbzzgxiMl5rEz/PNW6dR72ITvX8ASrgeTLAHAAA=
+EOF
+) | gunzip > $DEST_DIR/libEGL.so.1
+
```
jprvita: This is a symlink: `libEGL.so.1 -> libEGL.so.1.4 -> libMali.so`. Aren't overwriting `libMali.so`?

eos-link-host-mali-driver
jprvita: Ok, now I see what you are doing in this part. You are no linking the extension dir to the host dir directly, but creating a separate dir for the extension, then hardlinking `lib/libMali.so` in that this directly to the host's `libMali.so`. Sorry I missed that in the beginning, that threw off most of my other comments.

On that point tho, I believe if we ever update `libMali.so` on the host, the hardlink is broken on its own inode and the extension will always point to the old data. While this may even be desirable from the perspective of guaranteeing a certain runtime version does not change, doesn't this goes against the GL.host extension semantics? FWIW, without this change the GL.host extension will always provide the same GL version from the host.

eos-link-host-mali-driver
jprvita: I think the question is: do we want this change only in the runtime, or is it fine to have it in the host? If the later, then we should probably ship it in the ostree, and in that case I believe you could drop all the other changes in this file.

xjuan:
Updated script, works with https://obs-master.endlessm-sf.com/request/show/15188

eos-link-host-mali-driver
```diff
@@ -22,14 +22,45 @@
...
+    ln -sr $DEST_DIR/libMaliWrapper.so $DEST_DIR/libEGL.so.1.4
+    ln -sr $DEST_DIR/libMaliWrapper.so $DEST_DIR/libGLESv2.so.2.0
+}
+
+MALI_DIR="/usr/lib/mali400/arm-linux-gnueabihf"
+
+# Create GL host extensions
+if test -d $MALI_DIR; then
+    create_flatpak_gl_host_extension $MALI_DIR "/var/lib/flatpak"
+    create_flatpak_gl_host_extension $MALI_DIR "/var/endless-extra/flatpak"
```
jprvita: hardlinks don't cross filesystems boundaries, so this would not work for `/var/endless-extra`.

eos-link-host-mali-driver
```diff
@@ -22,14 +22,45 @@
...
-mkdir -p $(dirname "$DEST_DIR")
-ln -sf "$MALI_DIR" "$DEST_DIR"
+#
+# create_flatpak_gl_host_extension MALI_DIR FLATPAK_DIR
+#
+# Creates a flatpak GL host extension in $FLATPAK_DIR from $MALI_DIR directory
+# MALI_DIR needs to contain libMali.so and libMaliWrapper.so libraries.
+#
+function create_flatpak_gl_host_extension {
+    MALI_DIR=$1
```
jprvita: What you are doing here is overwrite (with the same value, BTW) the variable declared on line 59. Despite not leading to problems here, is better to avoid leaving a trap for future changers of this code. Either use `local` here or choose a different name.

eos-link-host-mali-driver
jprvita: Just realized this will not work for `/var/lib/flatpak` either, as due to ostree's bind-mounting `/usr` and `/var` are seen as separate filesystems:
```
$ sudo ln /usr/bin/true /var
ln: failed to create hard link ‘/var/true’ => ‘/usr/bin/true’: Invalid cross-device link
```

xjuan:
mali400-drivers-x11 (https://obs-master.endlessm-sf.com/request/show/15193) compiles and installs the wrapper library and the necessary links in `/usr/lib/mali400/arm-linux-gnueabihf/wrapper/`
So all it has to do this script is create the actual links for the extensions

xjuan:
Updated PR with new directory `/usr/lib/mali400/arm-linux-gnueabihf/flatpak/`

PR #138, opened 2017-09-26
Make NVIDIA's GL available to flatpak

jprvita:
Make NVIDIA's GL vendor implemention available to flatpak runtimes
through flatpak's org.freedesktop.Platform.GL.host extension.

https://phabricator.endlessm.com/T19169

### DIFF
--- a/nvidia-graphics-setup
+++ b/nvidia-graphics-setup
@@ -13,6 +13,25 @@ fi
 
 echo "Found nvidia device ${NV_DEVICE}"
 
+create_flatpak_extension() {
+	local NVIDIA_FLATPAK_DIR=/usr/lib/nvidia/flatpak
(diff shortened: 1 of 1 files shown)


nvidia-graphics-setup
```diff
@@ -70,6 +89,7 @@ if ! modprobe -c | grep -F -x --quiet "blacklist nvidia" &&
   load_module "${MODULE_DIR}"/nvidia-drm.ko
   if [[ -e "/sys/module/nvidia_drm" ]]; then
     echo "nvidia loaded successfully"
+    create_flatpak_extension
```
ramcq: Hmm - I think we need a codepath to remove the link in the case that the nvidia driver is blacklisted and we fall back to nouveau? Don't want to provide an incorrect driver to Flatpak.

PR #139, opened 2017-09-26
Install udev rules for USB mouse

mschiu77:
Support remote wakeup by USB mouse per the request from Acer Desktop.

https://phabricator.endlessm.com/T19106


dsd:
pushed with minor changes

PR #142, opened 2017-10-03
Remove eos-link-host-mali-driver boot service

jprvita:
Instead of having a boot service setup these symlinks, it is better to
use a tmpfiles entry to have systemd-tmpfiles setup them.

Also, we must design them in a way that supports moving to an ostree
which don't ship the symlink target -- right now if that happens, the
extension will be a dangling symlink, what breaks running any flatpak
that runs on a runtime that uses the extension (aka all apps, in this
case).

We'll ship a tmpfiles.d entry in the Mali package with the following
content, so the links disappear on shutdown:

D /run/flatpak/extension/org.freedesktop.Platform.GL.host/arm 0755 - - -
L /run/flatpak/extension/org.freedesktop.Platform.GL.host/arm/1.6 - - - - /usr/lib/mali400/arm-linux-gnueabihf/flatpak
L /var/lib/flatpak/extension - - - - /run/flatpak/extension
L /var/endless-extra/flatpak/extension - - - - /run/flatpak/extension

This way the only dangling symlink left behind would be
/var/lib/flatpak/extension, which doesn't break running flatpak apps.


eos-remove-old-flatpak-extension.service
```diff
@@ -0,0 +1,20 @@
...
+DefaultDependencies=false
+Conflicts=shutdown.target
+Before=systemd-tmpfiles-setup.service
+After=local-fs.target
+Wants=local-fs.target
+
+# Only run on ARM systems where the old directory exists
+ConditionArchitecture=arm
+ConditionPathExists=/var/lib/flatpak/extension/org.freedesktop.Platform.GL.host
+ConditionPathIsDirectory=/var/lib/flatpak/extension
```
ramcq: "Except for ConditionPathIsSymbolicLink=, all path checks follow symlinks."

I guess this is always before tmpfiles runs, but I'd feel a little more confident if we added a ConditionPathIsSymbolicLink=!/var/lib/flatpak/extension just so these jobs are idempotent and will can never go in circles doing/undoing each other's work. :)

jprvita: Agreed

eos-remove-old-flatpak-extension.service
```diff
@@ -0,0 +1,21 @@
...
+# Only run on ARM systems where the old directory exists
+ConditionArchitecture=arm
+ConditionPathIsSymbolicLink=!/var/lib/flatpak/extension
+ConditionPathIsDirectory=/var/lib/flatpak/extension
+ConditionPathExists=/var/lib/flatpak/extension/org.freedesktop.Platform.GL.host
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=/usr/bin/rm -rf /var/lib/flatpak/extension
```
ramcq: I think we also need to clear up /var/endless-extra/flatpak/extension as well, to undo the effects of https://github.com/endlessm/eos-boot-helper/blob/a7ba47d258420502ba828e9493e01229e5af9e57/eos-link-host-mali-driver#L33-L34 on split systems.

jprvita: Yes, good catch! Fixed now.

PR #143, opened 2017-10-04
T19491

rshuler:
https://phabricator.endlessm.com/T19491


Makefile.am
```diff
@@ -12,22 +12,19 @@ AM_DISTCHECK_CONFIGURE_FLAGS = \
...
-	eos-add-flatpak-apps-repos.service \
-	eos-add-flatpak-flathub-repos.service \
-	eos-add-flatpak-gnome-repos.service \
 	eos-enable-extra-upgrade.service \
 	eos-enable-zram.service \
 	eos-extra-resize.service \
 	eos-firewall-localonly.service \
 	eos-firstboot.service \
 	eos-fix-flatpak-branches.service \
+	eos-fix-flatpak-repos.service \
```
dbnicholson: I think this might be better named as `eos-update-flatpak-repos{,.service}`. Only the legacy external apps removal is about fixing. The rest is about adding remotes as a convenience. Or maybe `eos-setup-flatpak-repos`. What do you think?

rshuler: Yeah, I was struggling with the naming -- I had considered "update" and "configure", then went with "fix" because we already had a "fix-flatpak-branches".  But, I think you are right -- I'll change it to "update".

PR #148, opened 2017-10-19
Don't run eos-setup-flatpak-extension-dir-extra on every boot

jprvita:
If the link already exists we don't need to spawn ln.

https://phabricator.endlessm.com/T19691

### DIFF
--- a/eos-setup-flatpak-extension-dir-extra.service
+++ b/eos-setup-flatpak-extension-dir-extra.service
@@ -7,7 +7,7 @@ ConditionPathIsSymbolicLink=!/var/endless-extra/flatpak/extension
 [Service]
 Type=oneshot
 RemainAfterExit=yes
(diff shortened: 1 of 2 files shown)


eos-setup-flatpak-extension-dir-extra.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Setup flatpak extension directory on extra storage
 ConditionPathExists=/var/endless-extra/flatpak
+ConditionPathExists=!/var/endless-extra/flatpak/extension
```
ramcq: Could we do:
ConditionPathIsMountPoint=/var/endless-extra
ConditionPathExists=/var/endless-extra/flatpak
ConditionPathIsSymbolicLink=!/var/endless-extra/flatpak/extension

jprvita: Updated!

PR #151, opened 2017-11-01
T18497

tchx84:
https://phabricator.endlessm.com/T18497


wjt:
Is this still needed?

PR #152, opened 2017-11-10
Add service to install required flatpak extensions

xjuan:
https://phabricator.endlessm.com/T18342

### DIFF
--- a/eos-install-flatpak-extensions
+++ b/eos-install-flatpak-extensions
@@ -1,0 +1,9 @@
+#!/bin/bash -e
+# Copyright (C) 2017 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+# Install flatpak extensions
+extensions=(org.freedesktop.Platform.Icontheme.EndlessOS)
+for extension in ${extensions[@]}; do
+	flatpak install eos-sdk $extension || true
+done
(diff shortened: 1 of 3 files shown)


eos-install-flatpak-extensions
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
+# Copyright (C) 2017 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+# Install flatpak extensions
+extensions=(org.freedesktop.Platform.Icontheme.EndlessOS)
+for extension in ${extensions[@]}; do
+	flatpak install eos-sdk $extension || true
```
cosimoc: AFAICS this will download flatpaks at boot and block boot on network I/O, so it's a big no-no.
Can we make this part of an update to the Freedesktop runtime (or the Endless runtime)?

ebassi: We can make it autodownload alongside com.endlessm.apps.Platform//2, assuming gnome-software will respect the `autodownload = true` metadata.

xjuan:
This wont be needed anymore since we decided to make the extension download automatically which wont be an issue for non EOS systems since they wont have EndlessOS icontheme in their settings

PR #153, opened 2017-11-14
Remove GNOME remotes

joaquimrocha:
The GNOME Flatpak remotes are being replaced by the Flathub ones, so
these changes remove the GNOME remotes from the system installation
in order for the user not to see/install any repeated apps.

https://phabricator.endlessm.com/T19898


eos-remove-flatpak-gnome-repos
```diff
@@ -0,0 +1,12 @@
...
+# Copyright (C) 2017 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+# The GNOME Flatpak remotes have been replaced by the Flathub ones.
+# So we should remove the GNOME remotes we've previously added.
+
+gnome_apps_remote=gnome-apps
+gnome_runtimes_remote=gnome
+
+flatpak remote-delete --force --system $gnome_apps_remote || true
```
rshuler: I'm curious on the "--force --system" syntax here, when we don't have it in eos-remove-legacy-external-apps.  I'm guessing that they make sense, but perhaps we should add at least "--force" to eos-remove-legacy-external-apps.

joaquimrocha: I don't think force existed at the time. As for system, I think in these cases we should be more explicit than implicit. I can add it now to the //legacy ext apps// one.

PR #156, opened 2017-12-05
service: Add eos-companion-app.service and socket activation

smspillaz:
This will listen on port 1110 to launch the Companion App, if available.

https://phabricator.endlessm.com/T20130

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,6 +12,8 @@ AM_DISTCHECK_CONFIGURE_FLAGS = \
 	$(NULL)
 
 dist_systemdunit_DATA = \
+	eos-companion-app.service \
+	eos-companion-app.socket \
 	eos-enable-extra-upgrade.service \
 	eos-enable-zram.service \
 	eos-extra-resize.service \
--- a/Makefile.am
+++ b/Makefile.am
@@ -81,6 +81,11 @@ dist_polkitrules_DATA = \
 	10-allow-mounting-endless-image-device.rules \
 	$(NULL)
 
+tmpfilesddir = $(prefix)/lib/tmpfiles.d
+dist_tmpfilesd_DATA = \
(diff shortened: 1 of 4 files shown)


smspillaz:
Something I'm not sure about - should we have a wrapper executable that returns 0 in case the flatpak isn't installed?

cosimoc:
If the app is not installed, I'd expect the systemd unit to error out, so I don't think we need that

tmpfiles.d/eos-companion-app-helper.conf
```diff
@@ -0,0 +1,3 @@
+# Type Path    Mode UID  GID  Age Argument
+d /var/lib/companion-app-helper 0755 companion-app-helper companion-app-helper -
+Z /var/lib/companion-app-helper 0755 companion-app-helper companion-app-helper -
```
cosimoc: How is this file installed? Does not seem to get installed from anywhere.

smspillaz: Oops, I must have forgotten to add my changes to `Makefile.am`

smspillaz: Okay, fixed 😅

Makefile.am
```diff
@@ -79,6 +81,11 @@ dist_polkitrules_DATA = \
 	10-allow-mounting-endless-image-device.rules \
 	$(NULL)
 
+tmpfilesddir = $(libdir)/tmpfiles.d
```
cosimoc: We can't use `$(libdir)` here unfortunately. See e.g. this commit: https://github.com/endlessm/eos-metrics-instrumentation/commit/d4953db3bab8ffa55694849acf884c704ae144bb#diff-8d777f385d3dfec8815d20f7496026dc

smspillaz: Ah, frustrating. Thanks for catching!

PR #158, opened 2017-12-07
T20322: migrate Steam from eos-apps to flathub

ramcq:
Introduces a new function into eos-add-flatpak-apps-repos which allows us to
provide an array of flatpak names (either apps or runtimes) which are matched
by branch and/or origin, and moved atomically to the new branch and/or origin
as requested.

Changing the origin is achieved with our existing code for rewriting deploy
files, and changing the branch is done by hardlinking the old deploy folder
to the new branch name, and updating the "current" symlink if it's an app
(runtimes have no current symlink as they are always referenced using the
branch name).

This means we can move Steam from the eos-apps origin to flathub, and update
the branch from eos3 to stable. The user can then upgrade their Steam and get
working 32-bit NVidia drivers from upstream Flathub.

This branch also takes the opportunity to refactor eos-boot-helper's 4 separate
shell and python scripts into one python script which takes care of all boot-time
flatpak migrations/fixups. It can be used in future for migrating other gnome and
eos-apps flatpaks to flathub, removing other remotes, etc.

https://phabricator.endlessm.com/T20322


eos-update-flatpak-repos
```diff
@@ -1,10 +1,379 @@
...
+    refs = inst.list_installed_refs()
+    for migrate in FLATPAKS_TO_MIGRATE:
+        name = migrate['name']
+        kind = migrate['kind']
+        old_branch = migrate.get('old-branch')
+        old_origin = migrate.get('old-origin')
+
+        matching_refs = _filter_refs(refs, name=name, kind=kind,
+                                     branch=old_branch, origin=old_origin)
+        # also search for name.* runtimes to migrate to catch extensions
```
ptomato: Could you just do one filter here, with `prefix=name`?

eos-update-flatpak-repos
```diff
@@ -1,10 +1,379 @@
...
+                try:
+                    os.mkdir(new_branch_dir, mode=0o755)
+                    subprocess.check_call(['cp', '-al', deploy_dir, new_deploy_dir])
+                    os.symlink(ref_dir, os.path.join(new_branch_dir, 'active'))
+                    if kind == Flatpak.RefKind.APP:
+                        os.symlink('{}/{}'.format(arch, new_branch), current_tmp)
+                        os.rename(current_tmp, current)
+                    # this is unsafe whilst flatpaks are running, but: this is a startup job
+                    shutil.rmtree(old_branch_dir, ignore_errors=True)
+                except:
```
ptomato: best-practice nitpick: use `except Exception` or you'll also catch `SyntaxError` and the like.

flatpak-repos/Makefile.am
```diff
@@ -0,0 +1,7 @@
+flatpakrepodir = $(datarootdir)/eos-flatpak-repos
```
ptomato: I would recommend using `$(pkgdatadir)` for this

flatpak-repos/eos-sdk.flatpakrepo
```diff
@@ -0,0 +1,9 @@
+[Flatpak Repo]
+Title=Endless Apps Runtimes
+Url=https://ostree.endlessm.com/ostree/eos-sdk
+Homepage=https://endlessm.github.io/eos-knowledge-lib/
+Comment=Flatpak runtime for developing apps for Endless OS
+Description=Endless apps runtimes are released on a three month schedule and contain the libraries necessary for developing and running offline content apps. They receive minor bug fixing and security updates, and should be considered ABI stable and frozen.
```
rshuler: I know this is just cut and paste from what's already there, but I probably would not state about the three month schedule in the description, as that is an implementation detail that is subject to change.

eos-update-flatpak-repos
ramcq: No, we need to look for apps and runtimes separately - and even in the runtime case, this would inadvertently include com.example.runtime2 when we only wanted com.example.runtime and com.example.runtime.extension.

eos-update-flatpak-repos
ramcq: will fix

flatpak-repos/Makefile.am
ramcq: Alright. I kind of hate that this is all in eos-boot-helper but we can move them around whenever we want as this is an exclusively ostree-controlled path anyway.

flatpak-repos/eos-sdk.flatpakrepo
ramcq: Fixed in new PR.

ramcq: Fixed but can't find where the file was originally defined - I assume it was copied out of somewhere?

ptomato: It's in the root dir of eos-knowledge-lib, I think.

PR #159, opened 2017-12-07
T20322: migrate Steam from eos-apps to flathub

ramcq:
Introduces a new function into eos-add-flatpak-apps-repos which allows us to
provide an array of flatpak names (either apps or runtimes) which are matched
by branch and/or origin, and moved atomically to the new branch and/or origin
as requested.

Changing the origin is achieved with our existing code for rewriting deploy
files, and changing the branch is done by hardlinking the old deploy folder
to the new branch name, and updating the "current" symlink if it's an app
(runtimes have no current symlink as they are always referenced using the
branch name).

This means we can move Steam from the eos-apps origin to flathub, and update
the branch from eos3 to stable. The user can then upgrade their Steam and get
working 32-bit NVidia drivers from upstream Flathub.

This branch picks two commits from the full series so that the necessary fix can
be included in 2017-12 (to allow Steam to work with NVidia graphics) and the
others added soon afterwards.

https://phabricator.endlessm.com/T20322

### DIFF
--- a/eos-add-flatpak-apps-repos
+++ b/eos-add-flatpak-apps-repos
@@ -7,6 +7,7 @@
 # Authors:
 #  Mario Sanchez Prada <mario@endlessm.com>
 #  Philip Chimento <philip@endlessm.com>
+#  Robert McQueen <rob@endlessm.com>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -155,8 +156,14 @@ if __name__ == '__main__':
     logging.basicConfig(level=logging.INFO)
     logging.root.addHandler(journal.JournalHandler())
 
+    # ensure eos-sdk remote exists
     _add_flatpak_repo()
+
+    # switch EknServices to eos-sdk remote, migrating from old eos-apps remote (T17863)
     _update_deploy_file_for_app_and_remote('com.endlessm.EknServices', 'eos-sdk')
+
+    # fix up image-builder bug where com.endlessm.apps.* runtimes were
+    # mistakenly installed with eos-runtimes origin (T18366)
     for runtime in ('com.endlessm.apps.Platform', 'com.endlessm.apps.Sdk',
                     'com.endlessm.apps.Platform.Locale',
                     'com.endlessm.apps.Sdk.Locale',
@@ -165,6 +172,9 @@ if __name__ == '__main__':
             _update_deploy_file_for_app_and_remote(runtime, 'eos-sdk',
                                                    kind='runtime',
                                                    arch_branch=arch_branch)
+
+            # fix up image builder bug on 'sea' (south-east asia) which set
+            # an invalid subpath, causing no Locales to be installed (T19472)
             if runtime.endswith('.Locale'):
                 _replace_runtime_subpaths(runtime, arch_branch,
                                           ['/en,id,th,vi'],
--- a/eos-add-flatpak-apps-repos
+++ b/eos-add-flatpak-apps-repos
@@ -151,6 +151,118 @@ def _update_deploy_file(deploy_file_path, idx, new_val, old_val=None):
         os.chmod(tmp_deploy_file.name, 0o644)
         shutil.copy(tmp_deploy_file.name, deploy_file_path, follow_symlinks=True)
 
+FLATPAKS_TO_MIGRATE = [
+    {
+        'name': 'com.valvesoftware.Steam',
+        'kind': Flatpak.RefKind.APP,
+        'old-branch': 'eos3',
+        'new-branch': 'stable',
+        'old-origin': 'eos-apps',
+        'new-origin': 'flathub'
+    }
+]
+
+def _filter_refs(refs, name=None, prefix=None, kind=None, arch=None,
+                 branch=None, origin=None):
+    ret = []
+    for ref in refs:
+        if kind and kind != ref.get_kind():
+            continue
+        if name and name != ref.get_name():
+            continue
+        if prefix and not ref.get_name().startswith(prefix):
+            continue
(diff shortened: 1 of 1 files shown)


eos-add-flatpak-apps-repos
```diff
@@ -151,6 +151,118 @@ def _update_deploy_file(deploy_file_path, idx, new_val, old_val=None):
...
+
+def _migrate_installed_flatpaks():
+    insts = Flatpak.get_system_installations()
+    inst = insts[0]
+
+    refs = inst.list_installed_refs()
+    for migrate in FLATPAKS_TO_MIGRATE:
+        name = migrate['name']
+        kind = migrate['kind']
+        old_branch = migrate.get('old-branch')
```
rshuler: I don't see any reason why one would work and not the other, but it would be nice to use either `[]` or `.get()` consistently.

ramcq: [] throws an exception if the key doesn't exist, get() returns a default None. It's use here is intentional so that the migration dictionaries can include these keys optionally. In fact, there's no real reason to make name or kind mandatory, so we could just go .get all the way I guess. :)

rshuler: Oh, never mind.  I see.  It's because you are treating the branches as optional.

ramcq: When I come back and do the patch to migrate org.gnome.* then I'll make these optional so I can pass in a prefix match too. :)

eos-add-flatpak-apps-repos
```diff
@@ -151,6 +151,118 @@ def _update_deploy_file(deploy_file_path, idx, new_val, old_val=None):
...
+            continue
+        if prefix and not ref.get_name().startswith(prefix):
+            continue
+        if arch and arch != ref.get_arch():
+            continue
+        if branch and branch != ref.get_branch():
+            continue
+        if origin and origin != ref.get_origin():
+            continue
+        ret += [ref]
```
joaquimrocha: Nitpick: I think `ret.append(ref)` is more readable.

eos-add-flatpak-apps-repos
```diff
@@ -151,6 +151,118 @@ def _update_deploy_file(deploy_file_path, idx, new_val, old_val=None):
...
+        old_branch = migrate.get('old-branch')
+        old_origin = migrate.get('old-origin')
+
+        matching_refs = _filter_refs(refs, name=name, kind=kind,
+                                     branch=old_branch, origin=old_origin)
+        # also search for name.* runtimes to migrate to catch extensions
+        prefix = name + '.'
+        matching_refs += _filter_refs(refs, prefix=prefix,
+                                      kind=Flatpak.RefKind.RUNTIME,
+                                      branch=old_branch, origin=old_origin)
```
joaquimrocha: You could check first if the kind is `RUNTIME`, and in such case you call the function with `prefix=name`. As it will gather the runtime itself + the extensions, and avoids going twice over the list. (Assuming you're not deliberately trying to get runtimes with the same prefix as an app's name)
i.e. only call `_filter_refs` once (though I know we don't have to optimize for speed in such scripts).

eos-add-flatpak-apps-repos
```diff
@@ -151,6 +151,118 @@ def _update_deploy_file(deploy_file_path, idx, new_val, old_val=None):
...
+
+                try:
+                    os.mkdir(new_branch_dir, mode=0o755)
+                    subprocess.check_call(['cp', '-al', deploy_dir, new_deploy_dir])
+                    os.symlink(ref_dir, os.path.join(new_branch_dir, 'active'))
+                    if kind == Flatpak.RefKind.APP:
+                        os.symlink('{}/{}'.format(arch, new_branch), current_tmp)
+                        os.rename(current_tmp, current)
+                    # this is unsafe whilst flatpaks are running, but: this is a startup job
+                    shutil.rmtree(old_branch_dir, ignore_errors=True)
```
joaquimrocha: I'd probably add these two lines above to an `else` block because, even though we tell it to ignore errors, if there's a problem after the `os.rename` call, we'd end up with a broken app AFAICS. i.e. after we have all the links in place, we don't want to risk it removing the dir tree (and a simple unnoticed, last minute addition syntax error above would trigger that).

eos-add-flatpak-apps-repos
```diff
@@ -151,6 +151,118 @@ def _update_deploy_file(deploy_file_path, idx, new_val, old_val=None):
...
+            new_branch = migrate.get('new-branch')
+            if new_branch:
+                clashing_refs = _filter_refs(refs, name=name, kind=kind,
+                                             arch=arch, branch=new_branch)
+                if len(clashing_refs) > 0:
+                    logging.warning('Not migrating {}/{}/{}, new branch {} already exists'
+                                    .format(name, arch, branch, new_branch))
+                    continue
+
+                logging.info('Migrating {}/{}/{} to new branch {}'
```
joaquimrocha: This is a big function. I'd move this logic into its own function (likely `migrate_branch`).

eos-add-flatpak-apps-repos
ramcq: I didn't do that intentionally because this would widen the scope too much - com.example.runtime would match com.example.runtime2 as well for instance.

eos-add-flatpak-apps-repos
ramcq: Fixed.

eos-add-flatpak-apps-repos
ramcq: fixed in new PR (along with a bug about clearing up current_tmp)

eos-add-flatpak-apps-repos
ramcq: done!

PR #165, opened 2017-12-15
Add service to migrate /var/roothome to /sysroot/root

dbnicholson:
The ostree builder was setup so that /root was a symlink to
/var/roothome, but this had no connection to the system wide persistent
/sysroot/root. Now it's been changed so that /root is a symlink to
/sysroot/root, but existing users may have content in /var/roothome.
Migrate it if /var/roothome exists and is not a symlink to
/sysroot/root. A symlink from /var/roothome to /sysroot/root is kept in
case the user rolls back to a previous OS without the /root ->
/sysroot/root symlink.

https://phabricator.endlessm.com/T20475

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -24,6 +24,7 @@ dist_systemdunit_DATA = \
 	eos-image-boot-dm-setup.service \
 	eos-live-boot-overlayfs-setup.service \
 	eos-migrate-image-defaults.service \
+	eos-migrate-root-home.service \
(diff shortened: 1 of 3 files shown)


wjt:
Still desirable?

dbnicholson:
I'd still like to do it since the current setup feels icky to me, but I think the appetite is low. I'll close the PR but leave my branch in the repo should someone want to resurrect it.

PR #168, opened 2018-01-08
Add eos-fix-external-appstream

joaquimrocha:
EOS downloads an external appstream file "eos-extra.xml" every time
there's a refresh and the server's file is younger than the cached one.

We ship a version of eos-extra.xml in the images as a base point but the
name used to store the file no longer corresponds to the one GNOME
Software generates: it should now take an "org.gnome.Software-" prefix.

So this patch creates a service to rename or remove that file.

https://phabricator.endlessm.com/T19319

### DIFF
--- a/eos-fix-external-appstream.service
+++ b/eos-fix-external-appstream.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Rename or remove the outdated external appstream file
+ConditionPathExists=/var/cache/app-info/xmls/eos-extra.xml.gz
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/eos-fix-external-appstream
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


eos-fix-external-appstream.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Rename or remove the outdated external appstream file
+ConditionPathExists=/var/cache/app-info/xmls/eos-extra.xml
```
dbnicholson: The script handles `eos-extra.xml.gz`, but this is only running when `eos-extra.xml` exists.

dbnicholson:
Another way to do this without having to make a service that forks a shell script is to use tmpfiles.d. I believe this would do it:
```
# Copy old appstream file if new one doesn't exist
C /var/cache/app-info/xmls/eos-extra.xml.gz - - - - /var/cache/app-info/xmls/org.gnome.Software-eos-extra.xml.gz
# Remove the old appstream file
r /var/cache/app-info/xmls/eos-extra.xml.gz
```
From my reading of tmpfiles.d(5), these would be execute in order.

dbnicholson:
Forget the tmpfiles.d thing. I tried it out and it didn't work as expected.

PR #174, opened 2018-01-16
T20443: fix race of eos-update-flatpak-repos with mount of extra SD card

ramcq:
Introduce a new eos-extra-settled.target to address race conditions when
running eos-update-flatpak-repos on "split" systems where the default
flatpak system dir is provided by an SD card mounted at /var/endless-extra.
The problem is that (by design) the local-fs.target is not blocked by the
availability of /var/endless-extra so that systems can boot properly when
the SD card is unavailable - whether missing, damaged or corrupted - the
only symptom is the availability of the apps installed there, but the rest
of the system works OK.

If the discovery of the SD card is slow enough, or it's being resized or
fscked, the /var/endless-extra mount job isn't in the queue when the
eos-update-flatpak-repos job is evaluated, so the After= has no effect
and the job runs before /var/endless-extra is mounted, causing flatpak
to operate on /var/lib/flatpak (it selects /var/endless-extra by runtime
detection of /var/endless-extra/flatpak).

The fix is to introduce a new target whose conditions are met either
when /var/endless-extra is mounted, or if the system does not expect
to have an SD card, which is detected by checking for the link in
/etc/systemd/system/local-fs.wants created by the image builder to
enable /var/endless-extra on split systems.

The eos-update-flatpak-repos and eos-fix-flatpak-branches units are
updated to require eos-extra-settled.target. Note that this target
should never be used to do critical work on paths outside of the SD
card, because in the case the SD card is unavailable on a split system,
those would never run.

Also fixes a potential bug where eos-update-flatpak-repos could be
run concurrently with eos-fix-flatpak-branches.

https://phabricator.endlessm.com/T20443

### DIFF
--- a/eos-extra-target-generator
+++ b/eos-extra-target-generator
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+normal_dir="$1"
+target_unit="${1}/eos-extra-settled.target"
+
+unit="[Unit]
+Description=endless-extra SD card mounted (if expected)
+Wants=local-fs.target"
+
+if [ -L "/etc/systemd/system/local-fs.target.wants/var-endless\x2dextra.mount" ]; then
+  unit="${unit}\nRequiresMountsFor=/var/endless-extra"
+fi
+
+echo "${unit}" >"${target_unit}"
--- a/Makefile.am
+++ b/Makefile.am
@@ -31,6 +31,7 @@ dist_systemdunit_DATA = \
 	$(NULL)
 
 dist_systemdgenerator_SCRIPTS = \
+	eos-extra-target-generator \
 	eos-live-boot-generator \
 	$(NULL)
 
--- a/eos-fix-flatpak-branches.service
+++ b/eos-fix-flatpak-branches.service
@@ -10,6 +10,10 @@ After=local-fs.target
 Before=multi-user.target systemd-update-done.service
 ConditionNeedsUpdate=/var
 
+# Ensure that the extra disk for a split-disk system has been mounted
+Requires=eos-extra-settled.target
+After=eos-extra-settled.target
+
 [Service]
 Type=oneshot
 RemainAfterExit=true
--- a/eos-update-flatpak-repos.service
+++ b/eos-update-flatpak-repos.service
@@ -17,7 +17,8 @@ ConditionNeedsUpdate=|/var
 After=eos-ostree-remotes-config.service
 
 # Ensure that the extra disk for a split-disk system has been mounted
-After=var-endless_x2dextra.mount
+Requires=eos-extra-settled.target
+After=eos-extra-settled.target
 
 [Service]
 Type=oneshot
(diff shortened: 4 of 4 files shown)


eos-extra-target-generator
```diff
@@ -0,0 +1,14 @@
...
+normal_dir="$1"
+target_unit="${1}/eos-extra-settled.target"
+
+cat >"${target_unit}" <<EOF
+[Unit]
+Description=endless-extra SD card mounted (if expected)
+Wants=local-fs.target
+EOF
+
+if [ -L "/etc/systemd/system/local-fs.target.wants/var-endless\x2dextra.mount" ]; then
```
dbnicholson: Rather than testing this path, I think it would be better to use `systemctl is-enabled --quiet "var-endless\x2dextra.mount"` so that it doesn't depend on how the image builder enables the unit. The only thing I'd be worried about is somehow getting systemd into a deadlock doing that.

eos-extra-target-generator
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+normal_dir="$1"
+target_unit="${1}/eos-extra-settled.target"
+
+cat >"${target_unit}" <<EOF
```
dbnicholson: Probably overly pedantic, but I'm a little worried about writing this file directly and then editing it below. Both because we might error before it completes and that systemd might act on the unit before it's complete. I think it would be better to use a temp file and then copy/move into place at the end.

eos-extra-target-generator
jprvita: According to the manpage we can't use `systemctl` from generators:
```
Generators are run very early at boot and cannot rely on any external services. They may not talk to any other process. That includes simple things such as logging to syslog(3), or systemd itself (this means: no systemctl(1))!
```

eos-extra-target-generator
jprvita: I don't think systemd acts on the unit before the generator exits. From the manpage:
`If you are careful, you can implement generators in shell scripts. We do recommend C code however, since generators are executed synchronously and hence delay the entire boot if they are slow.
`

I'm not sure what you mean by "we might error before it completes", but it that is a real problem we may need to check whether https://github.com/endlessm/eos-boot-helper/blob/master/dracut/image-boot/eos-image-boot-generator#L79-L98 is also affected.

If we use a temp file which disappears before the generator exits I don't know if systemd would complain when trying to remove it on deamon-reload [1], or if it only keeps track of the files left by the generator after it terminates.

[1] `Units written by generators are removed when the configuration is reloaded. That means the lifetime of the generated units is closely bound to the reload cycles of systemd itself.`

dbnicholson: @jprvita For the "completes before error case", I mean if the script exited because of an error before it could write the `RequiresMountsFor` line. But since this script isn't running with `-e`, I guess that's really unlikely.

Thanks for checking the blocking behavior of the generator. In that case I guess it's fine to edit in place.

If you did want to use a temp file, there are completely standard ways to do it in shell to ensure it gets cleaned up. The temp file would only be used while editing and would have no connection to the actual file copied into /etc. But given how generators work, it's probably fine as is.

eos-extra-target-generator
dbnicholson: @jprvita thanks for checking! Let's just keep the simple check then. That said, it would be more correct to just check for existence of the file rather than whether it's a symlink. There's no reason it has to be a symlink besides convenience. Meh, this is fine as is.

PR #176, opened 2018-01-23
T20391

smspillaz:
https://phabricator.endlessm.com/T20391

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -91,3 +91,8 @@ dist_sbin_SCRIPTS = \
 	eos-repartition-mbr \
 	eos-update-flatpak-repos \
 	$(NULL)
+
+tmpfilesdir = $(prefix)/lib/tmpfiles.d
+dist_tmpfiles_DATA = \
+	tmpfiles.d/avahi-service-writers.conf \
+	$(NULL)
--- a/tmpfiles.d/avahi-service-writers.conf
+++ b/tmpfiles.d/avahi-service-writers.conf
@@ -1,0 +1,2 @@
+# Type Path    Mode UID  GID  Age Argument
+d /etc/avahi/services 0755 root avahi-service-writers


eos-chown-avahi-helper-to-writers.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Change owner of Avahi Services
+Before=multi-user.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=/usr/sbin/eos-chown-vaahi-helper-to-writers
```
rshuler: Typo: vaahi --> avahi

smspillaz: yeeks :(

smspillaz: Fixed.

cosimoc:
I'm not sure this PR is the best approach for this goal. You should be able to add a `tmpfiles.d` entry that ensures the same permissions applied on the directory, without a new boot helper entry.

eos-chown-avahi-helper-to-writers
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -e
```
cosimoc: As mentioned, I'm not sure a boot helper unit is actually the best way to achieve this. But if you do, please add the commands directly as `ExecStart` directives in the unit, instead of adding another binary in `$PATH`

smspillaz: @cosimoc Will `tmpfiles.d` entries change the permissions on an existing path? I suppose I could try that.

smspillaz: It looks like `tmpfiles.d` is only meant to be used with `/var`, `/tmp` and `/run`? https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html

cosimoc: @smspillaz I'm pretty sure it would also work in `/etc` -- we have already examples of that in `/usr/lib/tmpfiles.d`.

cosimoc: And yes, it should also change permissions on an existing path with the right flags.

tmpfiles.d/eos-companion-app-helper.conf
```diff
@@ -1,3 +1,5 @@
 # Type Path    Mode UID  GID  Age Argument
 d /var/lib/companion-app-helper 0755 companion-app-helper companion-app-helper -
 Z /var/lib/companion-app-helper 0755 companion-app-helper companion-app-helper -
+d /etc/avahi/services 0755 companion-app-helper companion-app-helper
+Z /etc/avahi/services 0755 companion-app-helper companion-app-helper
```
cosimoc: Should this not be `root:avahi-service-writers`? (Or only change the group if possible.)

smspillaz:
Agh, I must not have pushed my most recent branch...

smspillaz:
Okay, branch pushed.

Makefile.am
```diff
@@ -91,3 +91,8 @@ dist_sbin_SCRIPTS = \
 	eos-repartition-mbr \
 	eos-update-flatpak-repos \
 	$(NULL)
+
+tmpfilesddir = $(prefix)/lib/tmpfiles.d
```
pwithnall: (Really minor) nitpick: This variable is called `tmpfilesdir` in `systemd.pc` (note `dir` not `ddir`), so we should probably name it consistently with that to ease greppability.

tmpfiles.d/avahi-service-writers.conf
```diff
@@ -0,0 +1,3 @@
+# Type Path    Mode UID  GID  Age Argument
+d /etc/avahi/services 0755 root avahi-service-writers
+Z /etc/avahi/services 0755 root avahi-service-writers
```
pwithnall: Is the `Z` line necessary? The documentation for `d` (in `man tmpfiles.d`) says “The mode and ownership will be adjusted if specified and the directory already exists.”.

Makefile.am
smspillaz: Fair point. Fixed.

tmpfiles.d/avahi-service-writers.conf
smspillaz: So `Z` will `Recursively set the access mode, group and user, and restore the SELinux security context of a file or directory if it exists`. My thinking here was that if someone had already touched that file then the `Z` will at least reset the permissions so that it is writable.

Then again, I don't see that happening on a normal system so I'm happy to drop this line. Drop it?

cosimoc: I don't see it as being harmful, but you're right -- the directory permissions should be enough, so let's remove it for now.

cosimoc:
@pwithnall That's a good point about it conceptually being a better approach to have the avahi daemon do this; on the other hand, the service file writing is not done by avahi itself here, and it sounds like a more flexible approach to ensure the right permissions this way.

PR #178, opened 2018-01-24
Add eos-boot-upgrade and related service file

carlocaione:
https://phabricator.endlessm.com/T14870

### DIFF
--- a/eos-boot-upgrade
+++ b/eos-boot-upgrade
@@ -1,0 +1,278 @@
+#!/usr/bin/python3 -u
+#
+# Copyright (C) 2018 Endless Mobile, Inc.
+# Author:
+#  Carlo Caione <carlo@endlessm.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+# USA.
+
+import contextlib
+import distutils.dir_util
+import filecmp
+import json
+import os
+import re
+import shutil
+import subprocess
+
+
+class BootUpgradeException(Exception):
+    def __init__(self, *args):
+        super(BootUpgradeException, self).__init__(args)
+        self.msg = ' '.join(map(str, args))
+
+    def __str__(self):
+        return str(self.msg)
+
+
+class Partition(object):
+    def __init__(self, part_number, disk, **kwargs):
+        self.disk = disk
+        self.number = part_number
+
+        for config_name in kwargs:
+            setattr(self, config_name, kwargs[config_name])
+
+    @staticmethod
+    def parse(config, disk):
+        '''Parse the partition data passed by sfdisk json data'''
+        if 'node' not in config:
+            raise BootUpgradeException('"node" missing from sfdisk partition info:',
+                                       config)
+
+        part_number = re.match('^.*([0-9]+)$',
+                               config['node']).group(1)
+        part_config = {
+            'device': config['node'],
+            'uuid': config.get('uuid'),
+            'type': config.get('type')
+        }
+        return Partition(part_number, disk, **part_config)
+
+
+class Disk(object):
+    def __init__(self, device):
+        self.label = None
+        self.partitions = {}
+        self.device = device
+        self._read_partition_table()
+
+    def _read_partition_table(self):
+        '''Read partition table using sfdisk json data'''
+        sfdisk_json = subprocess.check_output(['sfdisk',
+                                               '--json',
+                                               self.device]).decode('utf-8')
+        disk_config = json.loads(sfdisk_json)
+
+        self.label = disk_config['partitiontable']['label'] or None
+
+        for part_config in disk_config['partitiontable']['partitions']:
+            part = Partition.parse(part_config, self)
+            self.partitions[part.number] = part
+
+    def get_partition_by_guid(self, guid):
+        '''Get partition by GUID / partition type'''
+        for part_num, conf in self.partitions.items():
+            if conf.type == guid:
+                return conf
+        return None
+
+    @classmethod
+    def get_booting_device(cls):
+        '''Get the booting device'''
+        root_part_device = subprocess.check_output([
+            'systemctl', 'show', '-p', 'What', 'sysroot.mount'
+            ]).decode('utf-8').strip()
+
+        root_part_device = root_part_device[5:]
+        if not root_part_device:
+            raise BootUpgradeException('Could not identify root device')
+
+        boot_disk_device = re.sub(r'p?[0-9]+$', "", root_part_device)
+        return cls(boot_disk_device)
+
+
+@contextlib.contextmanager
+def mount(device, mount_point):
+    '''Mount device on mount_point'''
+    subprocess.check_call(['mount', device, mount_point])
+    try:
+        yield mount_point
+    finally:
+        subprocess.check_call(['umount', mount_point])
+
+
+def parse_cmdline():
+    '''Parse /proc/cmdline'''
+    options = {}
+    with open('/proc/cmdline', 'r') as f:
+        for option in f.readline().strip().split():
+            fields = option.split("=", 1)
+            if len(fields) == 1:
+                options[fields[0]] = True
+            else:
+                options[fields[0]] = fields[1]
+    return options
+
+
+def get_disk():
+    '''Get the booting disk'''
+    disk = Disk.get_booting_device()
+    print('[*] Booting disk: ' + disk.device)
+
+    if disk.label != 'gpt':
+        raise BootUpgradeException('[!] Partition table is not GPT')
+
+    root_partition = disk.get_partition_by_guid('4F68BCE3-E8CD-4DB1-96E7-FBCAF984B709')
+    if not root_partition:
+        raise BootUpgradeException('[!] root partition not found (or wrong GUID)')
+
+    print('[*] Root partition: ' + root_partition.device)
+    return disk
+
+
+def get_esp(disk):
+    '''Get the ESP partition data'''
+    esp_partition = disk.get_partition_by_guid('C12A7328-F81F-11D2-BA4B-00A0C93EC93B')
+    if not esp_partition:
+        raise BootUpgradeException('[!] ESP partition not found (or wrong GUID)')
+    if not esp_partition.device.endswith('1'):
+        raise BootUpgradeException('[!] ESP is not the first partition')
+
+    print('[*] ESP partition: ' + esp_partition.device)
+    return esp_partition
+
+
+def get_dir_size(dir_path):
+    '''Get the directory size'''
+    total_size = 0
+    for (dirpath, _, filenames) in os.walk(dir_path):
+        for f in filenames:
+            fp = os.path.join(dirpath, f)
+            total_size += os.path.getsize(fp)
+    return total_size
+
+
+def check_esp_space(esp_mount, ostree_boot_efi_path):
+    '''Check that we have enough space in the ESP to copy the new data'''
+    ostree_boot_efi_size = get_dir_size(ostree_boot_efi_path)
+    esp_free_space = shutil.disk_usage(esp_mount).free
+    return esp_free_space > ostree_boot_efi_size
+
+
+def dir_cmp(dcmp):
+    '''Helper for directories comparison'''
+    if dcmp.diff_files or dcmp.right_only:
+        return True
+
+    for sub_dcmp in dcmp.subdirs.values():
+        if dir_cmp(sub_dcmp):
+            return True
+    return False
+
+
+def check_for_upgrade(esp_mount, ostree_boot_efi_path):
+    '''Recursively compare the files to upgrade'''
+    dcmp = filecmp.dircmp(esp_mount, ostree_boot_efi_path)
+    return dir_cmp(dcmp)
+
+
+def do_uefi_upgrade(esp_mount, ostree_boot_efi_path):
+    '''Actually start the UEFI upgrade copying the files'''
+    esp_mount_tmp = os.path.join(esp_mount, 'EFI.new')
+
+    print('[*] Updated:')
+    try:
+        file_list = distutils.dir_util.copy_tree(ostree_boot_efi_path,
+                                                 esp_mount_tmp)
+        os.sync()
+
+        for f_source in file_list:
+            f_rp = os.path.relpath(f_source, esp_mount_tmp)
+            f_dest = os.path.join(esp_mount, f_rp)
+
+            if not os.path.isfile(f_dest) or not filecmp.cmp(f_source, f_dest):
+                shutil.move(f_source, f_dest)
+                print(f_rp)
+    finally:
+        if os.path.exists(esp_mount_tmp):
+            distutils.dir_util.remove_tree(esp_mount_tmp)
+
+
+def start_uefi_upgrade():
+    '''Start the UEFI files upgrade'''
+    esp_mount = '/boot/efi'
+    ostree_boot_efi_path = os.path.join(cmdline['ostree'], 'boot/efi')
+
+    if not os.path.isdir(esp_mount):
+        raise BootUpgradeException('[!] ESP mountpoint',
+                                   esp_mount, 'does not exist')
+
+    disk = get_disk()
+    esp_partition = get_esp(disk)
+
+    with mount(esp_partition.device, esp_mount) as esp_mount:
+        if not check_for_upgrade(esp_mount, ostree_boot_efi_path):
+            print('[*] No UEFI files to upgrade')
+            return
+
+        if not check_esp_space(esp_mount, ostree_boot_efi_path):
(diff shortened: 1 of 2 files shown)


eos-grub-upgrade.service
```diff
@@ -0,0 +1,21 @@
+# GRUB/SHIM updater
```
wjt: Seems redundant with `Description=`

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
+#!/bin/bash
+# Copyright (C) 2018 Endless Mobile, Inc.
+# Licensed under the GPLv2
+#
+# Extracted from https://phabricator.endlessm.com/w/software/x86-64-boot-paths/
```
wjt: Maybe "Currently handles the following boot paths (taken from URL):". Also, this has more data than that table, including whether bootia32.efi is present; please could you update the wiki?

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+    /dev/sd??)
+      root_disk=${root_part%?}
+      ;;
+    /dev/*p[0-9])
+      root_disk=${root_part%p?}
+      ;;
+  esac
+
+  if [[ -z ${root_disk} ]]; then
+    echo "No root disk found for ${root_part}"
```
wjt: I ran this on a dual-boot system, and the first line of non-`bash -x` output is:

```
Booting disk: No root disk found for /dev/mapper/endless-image3
```

which shows that the error was not properly handled by the caller. This message, and many others, should go to stderr, and the caller needs to check the exit status (or, better, the whole script should use `set -e`).

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+  else
+    echo "bios"
+  fi
+}
+
+mount_ESP() {
+  local root_disk=$1
+  local esp_guid="C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
+
+  # Check if ESP does exist and retrieve the partition name
```
wjt: "Check if ESP exists ..." is more idiomatic

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+  local esp_part=$(sfdisk -d "${root_disk}" | grep -i type=${esp_guid} | sed 's/ .*//')
```
wjt: `sfdisk --dump "$root_disk" | awk '/type='$esp_guid'/ { print $1 }'` is a little neater than `| grep | sed` if we really must parse the output of `sfdisk` with shell pipelines.

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+
+  if [[ -z ${root_part} ]]; then
+    echo "Couldn't identify root device"
+    exit 1
+  fi
+
+  # Get the root disk device file
+  local root_disk=
+  case ${root_part} in
+    /dev/sd??)
```
wjt: Could do this check the other way around: if the path doesn't match `/dev/*p[0-9]` then just chop off the last character (and don't explicitly match `/dev/sd??`). I don't know the full menagerie of device name formats though.

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+  # Get the installation type
+  local inst_type=$(get_installation_type "uefi" "gpt")
+  if [[ -z ${inst_type} ]]; then
+    echo "Installation type unknown"
+    exit 1
+  fi
+  echo "Installation type: ($inst_type)"
+
+  GRUB_AMD64=${GRUB_AMD64[$inst_type]}
+  SHIM=${SHIM[$inst_type]}
```
wjt: So, we are not migrating from the `\EFI\BOOT\{bootx64.efi,grubx64.efi}` layout to the `\EFI\BOOT\{bootx64.efi,fallback.efi}` + `\EFI\Endless\{shim.efi,MokManager.efi,grubx64.efi,BOOT.CSV}` layout?

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+  if [[ "${#files_to_upgrade[@]}" -eq 0 ]]; then
+    echo "Nothing to upgrade"
+    return
+  fi
+
+  # Let's try to do this atomically using mv because the old and the new
+  # files are now in the same partition
+  local file
+  for file in "${files_to_upgrade[@]}"; do
+    mv -f "${file}.new" "${file}"
```
wjt: What if this fails? Should we try to roll back any files that have already been overwritten?

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+  fi
+
+  # Mount the ESP
+  if ! mount "${esp_part}" ${ESP_MOUNTPOINT}; then
+    echo "Failed to mount ESP"
+    exit 1
+  fi
+}
+
+check_for_upgrade() {
```
wjt: From the name of this function, I'd assume it does not write anything to disk, but in fact it copies `$2` to `$1.new`. Maybe `prepare_for_upgrade` would be a better name.

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+    echo "No root disk found for ${root_part}"
+    exit 1
+  fi
+
+  echo "${root_disk}"
+}
+
+# Returns "gpt" or "dos"
+get_partition_table() {
+  local root_disk=$1
```
wjt: Something like this will raise an error if the parameter is unset or empty:

```bash
local root_disk=${1:?$FUNCNAME: root_disk parameter missing}
```

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+
+  GRUB_AMD64=${GRUB_AMD64[$inst_type]}
+  SHIM=${SHIM[$inst_type]}
+
+  # Check if the files need to be upgraded
+  local files_to_upgrade=( )
+  local f
+  for f in "${EFI_LIST[@]}"; do
+    eval old_file=\$"$f"
+    eval new_file=\$"NEW_$f"
```
wjt: yuck!

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+declare -A GRUB_AMD64=(
+  [A]="${ESP_MOUNTPOINT}/EFI/endless/grubx64.efi"
+  [B]="${ESP_MOUNTPOINT}/EFI/BOOT/grubx64.efi"
+  [C]="${ESP_MOUNTPOINT}/EFI/BOOT/bootx64.efi"
+)
+NEW_GRUB_AMD64=/usr/lib/grub/images/x86_64-efi/grubx64.efi
+
+## SHIM
+declare -A SHIM=(
+  [A]="${ESP_MOUNTPOINT}/EFI/BOOT/bootx64.efi"
```
wjt: So I see that what's happening here is that A, B and C are arbitrary labels, and the data structure you're simulating is something like:

```js
AMD64_LAYOUTS = [
    {grub: '/EFI/endless/grubx64.efi', shims: ['/EFI/BOOT/bootx64.efi', '/EFI/Endless/shim.efi']},
    // ...
]
```

It's a shame to have to disguise this structure as a bunch of strings and string->string maps, complete with runtime-evaluated variable names…

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+  local new_file
+
+  # Mount the ESP
+  mount_ESP "${root_disk}"
+
+  # Get the installation type
+  local inst_type=$(get_installation_type "uefi" "gpt")
+  if [[ -z ${inst_type} ]]; then
+    echo "Installation type unknown"
+    exit 1
```
wjt: I don't think it's appropriate to exit failure from this script on systems with unsupported configurations, because that will mean the systemd unit fails, we'll enter a degraded state, and this causes log messages to be visible on shutdown: https://phabricator.endlessm.com/T17983#419572

eos-grub-upgrade
```diff
@@ -0,0 +1,276 @@
...
+  # Cleanup
+  for file in "${files_to_upgrade[@]}"; do
+    rm -f "${file}.new"
+  done
+
+  echo "All done!"
+}
+
+root_disk=$(get_root_disk)
+echo "Booting disk: ${root_disk}"
```
wjt: I think "Booted disk: " would be clearer

wjt:
I somehow lost my high-level comment: I don't think Bash is an appropriate language to write this program in. (Opinions may differ, and this is a pet peeve of mine, but I think some of the contortions in this script support my opinion!)

It would also be very nice to have some tests.

eos-boot-upgrade.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Updater for GRUB/SHIM
+DefaultDependencies=no
+Conflicts=shutdown.target
```
wjt: OOI what happens if this unit is running when the user shuts down the system? Will this be killed? (I appreciate it is extremely unlikely.)

eos-boot-upgrade
```diff
@@ -0,0 +1,267 @@
...
+    conf_path = '/boot/grub/grub.cfg'
+    new_conf_path = '/usr/lib/grub/conf/grub.cfg'
+
+    try:
+        if filecmp.cmp(conf_path, new_conf_path):
+            print('[*] No configuration file to upgrade')
+            return
+
+        shutil.copyfile(new_conf_path, conf_path);
+        os.sync()
```
wjt: I think it would be safer to copy the new conf to `/boot`, `fsync()` it, then `rename()` it over `conf_path`. This sequence is rumoured to guarantee that in the event of a crash `/boot/grub/grub.cfg` would either contain its old or its new contents. By copying directly over it, its contents might be truncated or filled with `NUL` bytes if the system crashes before `sync()` completes.

eos-boot-upgrade
```diff
@@ -0,0 +1,267 @@
...
+    dcmp = filecmp.dircmp(esp_mount, ostree_boot_efi_path)
+    return _dir_cmp(dcmp)
+
+def do_uefi_upgrade(esp_mount, ostree_boot_efi_path):
+    '''Actually start the UEFI upgrade copying the files'''
+    tmp_dest_directory = os.path.join(esp_mount, 'new')
+
+    try:
+        distutils.dir_util.copy_tree(ostree_boot_efi_path, tmp_dest_directory)
+        upgrade_list = distutils.dir_util.copy_tree(ostree_boot_efi_path, esp_mount)
```
wjt: Similar here, I think it would be safer to `rename()` the new files over the old.

eos-boot-upgrade
```diff
@@ -0,0 +1,267 @@
...
+        disk = get_disk();
+        esp_partition = get_esp(disk);
+
+        with mount(esp_partition.device, esp_mount) as esp_mount:
+            if not check_for_upgrade(esp_mount, ostree_boot_efi_path):
+                print('[*] No UEFI files to upgrade')
+                return
+
+            if not check_esp_space(esp_mount, ostree_boot_efi_path):
+                raise BootUpgradeException('[!] Not enough space in the ESP')
```
wjt: Might be better to just try the upgrade:

* Copy all the files over; if there's not enough space, this will fail
* Now rename them into place; this should need no additional space, I guess?
* On success or error, clean up the temporary directory

eos-boot-upgrade
```diff
@@ -0,0 +1,267 @@
...
+    return False
+
+def check_for_upgrade(esp_mount, ostree_boot_efi_path):
+    '''Recursively compare the files to upgrade'''
+    dcmp = filecmp.dircmp(esp_mount, ostree_boot_efi_path)
+    return _dir_cmp(dcmp)
+
+def do_uefi_upgrade(esp_mount, ostree_boot_efi_path):
+    '''Actually start the UEFI upgrade copying the files'''
+    tmp_dest_directory = os.path.join(esp_mount, 'new')
```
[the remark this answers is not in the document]
carlocaione: isn't the `distutils.dir_util.remove_tree()` in the `finally` clause enough to guarantee that the temp directory is deleted if something goes wrong?

eos-boot-upgrade
carlocaione: `/boot` is on the same partition, so we can directly `rename()` it over.

eos-boot-upgrade
```diff
@@ -0,0 +1,278 @@
...
+        file_list = distutils.dir_util.copy_tree(ostree_boot_efi_path,
+                                                 esp_mount_tmp)
+        os.sync()
+
+        for f_source in file_list:
+            f_rp = os.path.relpath(f_source, esp_mount_tmp)
+            f_dest = os.path.join(esp_mount, f_rp)
+
+            if not os.path.isfile(f_dest) or not filecmp.cmp(f_source, f_dest):
+                shutil.move(f_source, f_dest)
```
wjt: `renameat2(..., RENAME_EXCHANGE)` doesn't work on FAT, so we can't use that to swap the old/new `EFI\Endless` directories atomically. Without some mechanism to juggle a pair of BootOrder entries we can't entirely avoid this window between updating the files. So we need to:

1. make the window as narrow as possible
2. add/replace files in a carefully-determined order to reduce the number of ways that a poorly-timed loss of power could mess us up

Here are two examples of how the order can affect whether the system boots after a loss of power.

Suppose we have to rotate the key our GRUB is signed with. Then suppose while running this script the computer loses power in between overwriting Shim with a new version that doesn't trust the old key, and then overwriting GRUB with a new version signed with a new key. Then the system won't boot in secure boot mode. If we made sure any new GRUB binaries were still signed with the old key as well as the new one, then if we always copied the new GRUB executable before copying the new Shim, the old Shim would still trust the new GRUB.

Another example: before eos3.1, the ESP contained:

* \EFI\BOOT\
  * bootx64.efi (shim)
  * grubx64.efi

From eos3.1:

* \EFI\BOOT\
  * bootx64.efi (shim)
  * fallback.efi
* \EFI\Endless\
  * shim.efi
  * grubx64.efi
  * BOOT.CSV

When upgrading from the former to the latter we need to take care to update the full contents of \EFI\Endless *before* we add fallback.efi, because the logic in Shim is: "if (run as \EFI\BOOT\BOOT*.EFI && \EFI\BOOT\fallback.efi exists) { run fallback } else { run \EFI\BOOT\grubx64.efi }", so if we add fallback.efi before \EFI\Endless is fully-populated, and then lose power, the system will no longer boot.

eos-boot-upgrade
```diff
@@ -0,0 +1,278 @@
...
+
+        self.label = disk_config['partitiontable']['label'] or None
+
+        for part_config in disk_config['partitiontable']['partitions']:
+            part = Partition.parse(part_config, self)
+            self.partitions[part.number] = part
+
+    def get_partition_by_guid(self, guid):
+        '''Get partition by GUID / partition type'''
+        for part_num, conf in self.partitions.items():
```
wjt: part_num is unused – you can do this instead:

```python
for conf in self.partitions.values():
    ...
```

eos-boot-upgrade
```diff
@@ -0,0 +1,278 @@
...
+    total_size = 0
+    for (dirpath, _, filenames) in os.walk(dir_path):
+        for f in filenames:
+            fp = os.path.join(dirpath, f)
+            total_size += os.path.getsize(fp)
+    return total_size
+
+
+def check_esp_space(esp_mount, ostree_boot_efi_path):
+    '''Check that we have enough space in the ESP to copy the new data'''
```
wjt: What's the rationale for doing this rather than just trying to copy the files over? `get_dir_size()` returns an underestimate of the size required because directories occupy space too, so we still have to handle "disk full" in the upgrade function.

eos-boot-upgrade
wjt: In fact, since we never index `self.partitions` by its actual key (the partition number), `self.partitions` could just be a list rather than a dictionary.

eos-boot-upgrade
```diff
@@ -0,0 +1,278 @@
...
+        print('[!] Not an ostree system?')
+        exit(0)
+
+    if not os.path.exists('/sys/firmware/efi'):
+        print('[!] BIOS upgrade not yet implemented')
+        exit(0)
+
+    print('==> Starting UEFI/configuration upgrade...')
+    start_uefi_upgrade()
+    start_conf_upgrade()
```
wjt: To narrow the window of time in which a power failure would leave the system unbootable, it might be better to interleave these two:

1. copy new config file to /boot/grub/grub.cfg.new
2. copy new ESP files to $ESP/EFI.new
3. `os.sync()`
4. rename everything into place, if changed
5. `os.sync()`
6. delete any left-over files

(There will be a lot of disk IO happening because of other post-update jobs so we'd expect `sync()` to be slow.)

eos-boot-upgrade
```diff
@@ -0,0 +1,278 @@
...
+                print(f_rp)
+    finally:
+        if os.path.exists(esp_mount_tmp):
+            distutils.dir_util.remove_tree(esp_mount_tmp)
+
+
+def start_uefi_upgrade():
+    '''Start the UEFI files upgrade'''
+    esp_mount = '/boot/efi'
+    ostree_boot_efi_path = os.path.join(cmdline['ostree'], 'boot/efi')
```
wjt: We should make the ostree deployment path into a command-line parameter. This would allow us to run this script from eos-updater after deploying a new OS update but *before* rebooting into it. (If we'd had this already, we could have fixed the stuck-on-the-old-deployment bug with an OS update.)

eos-boot-upgrade
```diff
@@ -0,0 +1,278 @@
...
+            raise BootUpgradeException('[!] Not enough space in the ESP')
+
+        do_uefi_upgrade(esp_mount, ostree_boot_efi_path)
+
+
+def start_conf_upgrade():
+    '''Start the GRUB configuration upgrade'''
+    conf_path = '/boot/grub/grub.cfg'
+    conf_path_tmp = conf_path + '.tmp'
+    conf_path_new = '/usr/lib/grub/conf/grub.cfg'
```
wjt: This would also need to be relative to the ostree deployment passed on the command line, if any.

PR #189, opened 2018-03-17
Preserve partition attributes during repartitioning

carlocaione:
https://phabricator.endlessm.com/T21538

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -203,7 +203,7 @@ fi
 # remove the last-lba line so that we fill the disk
 # + Randomize PARTUUIDs
 # + Randomize GPT disk UUID
-parts=$(echo "$parts" | sed -e '/^last-lba:/d;s/ uuid=.*//;/^label-id:/d')
+parts=$(echo "$parts" | sed -e '/^last-lba:/d;s/ uuid=[a-fA-F0-9-]\{36\},\{0,1\}//;/^label-id:/d')
 
 if [ $new_size -gt $part_size ]; then
   echo "Try to resize $root_part to fill $new_size sectors"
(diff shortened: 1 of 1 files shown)


dracut/repartition/endless-repartition.sh
```diff
@@ -203,7 +203,7 @@ fi
 # remove the last-lba line so that we fill the disk
 # + Randomize PARTUUIDs
 # + Randomize GPT disk UUID
-parts=$(echo "$parts" | sed -e '/^last-lba:/d;s/ uuid=.*//;/^label-id:/d')
+parts=$(echo "$parts" | sed -e '/^last-lba:/d;s/ uuid=[a-fA-F0-9-]\{36\},\{0,1\}//;/^label-id:/d')
```
dsd: can you remind me what the \{0,1\} bit does here?

Can you include in the commit message the line of sfdisk output showing the UUID followed by partition attributes? so that we have a clear documented example of input data that we are working with here..

dracut/repartition/endless-repartition.sh
```diff
@@ -288,7 +288,9 @@ sed -e "s:$orig_root_part:$root_part:" \
 # In the MBR case the marker partition is removed as part of sfdisk rewriting
 # the partition table above
 if [ "$pt_label" = "gpt" ]; then
-  sfdisk --force --part-attrs $root_disk $partno ''
+  attrs=$(sfdisk --part-attrs $root_disk $partno)
+  attrs=$(echo "$attrs" | sed -e 's/,\{0,1\}GUID:55//g')
```
dsd: again please provide an example of the input data in the commit message

dracut/repartition/endless-repartition.sh
carlocaione: > can you remind me what the {0,1} bit does here?

After the UUID can be present `{1}` or not `{0}` a `,` depending on the number of attributes.

I'll fix the commit messages.

dsd: Thanks that is clearer.

But search for "Remove marker" further down in this file. Is that code not also problematic for you? I guess it also removes the LegacyBIOSBootable flag.

carlocaione: That was addressed in the second commit if I'm identifying the point correctly.

PR #190, opened 2018-03-20
repartition: Allow swap to be enabled by e-i-b

mwleeds:
The image builder now has a hook to create a file /etc/eos-force-swap
with a number of sectors in it, so this commit changes
endless-repartition.sh to read that and create swap accordingly,
instead of following the usual heuristic of checking for 100 GB of free
space.

https://phabricator.endlessm.com/T21667


mwleeds:
There's already a hook in e-i-b that modifies the initramfs, so that should be possible I think

mwleeds:
This strategy is being abandoned in favor of zram

PR #195, opened 2018-04-05
enable persistent journal on durable storage devices

treitter:
This is meant to allow us easier access to detailed logs by ensuring
they already exist by the time an issue has been discovered. This is
particularly important when users often change their behavior just after
enabling logging which can lead to reporting of unhelpful journals.

Specifically, we enable the persistent journal by creating
/var/log/journal if /var is on a non-{MMC,SD} device since those storage
types can be degraded with significant writing (though our configuration
rate-limits the journal's access to storage significantly).

https://phabricator.endlessm.com/T21771

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,6 +12,7 @@ AM_DISTCHECK_CONFIGURE_FLAGS = \
 	$(NULL)
 
 dist_systemdunit_DATA = \
+	eos-config-journal.service \
 	eos-enable-extra-upgrade.service \
 	eos-enable-zram.service \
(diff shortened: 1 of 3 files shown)


treitter:
NOTE: this should only be merged after https://github.com/endlessm/eos-meta/pull/491 so `eos-diagnostics` won't potentially include many boots' worth of journal output.

treitter:
@jprvita huh, not sure how that happened. I rebased on master and force-pushed again. There's no actual difference in the resulting source but this version of the branch no longer has that extra commit.

PR #198, opened 2018-04-27
Cleanup /etc/flatpak/remotes.d take 2

dbnicholson:
The previous fix for this wouldn't actually work in some cases since the flatpak installation can't even be opened if there are duplicate remote entries. This moves similar functionality to eos-ostree-remotes-config, which is setup to run before any other ostree or flatpak services. The work done there is done entirely with python's ConfigParser since ostree will refuse to open the repo when there are duplicate remote entries.

I tested this on my system and I think it should work well, but I didn't actually try it out on old images...

https://phabricator.endlessm.com/T22258

### DIFF
--- a/eos-ostree-remotes-config
+++ b/eos-ostree-remotes-config
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/usr/bin/python3
 
 # eos-ostree-remotes-config - configure /ostree/repo for safe remote adding
 #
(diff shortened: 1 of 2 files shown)


eos-ostree-remotes-config
```diff
@@ -28,16 +28,119 @@
...
+if not read_config_file(repo_config, REPO_CONFIG_PATH):
+    # This is bad...
+    print('Error: could not read', REPO_CONFIG_PATH, file=sys.stderr)
+    sys.exit(1)
+
+# Set the core.add-remotes-config-dir option and write back now in case
+# there are errors later
+if repo_config.has_option('core', REMOTE_CONFIG_OPT):
+    value = repo_config.get('core', REMOTE_CONFIG_OPT)
+    if value != 'false':
```
ramcq: I'd be inclined to rework the logic here so we just change this setting to false every time.

eos-ostree-remotes-config
```diff
@@ -28,16 +28,119 @@
...
+
+# Read all the ostree and flatpak remotes.d files into the
+# configuration. This has the effect of merging them into the repo
+# config.
+remotes_d_confs = []
+for remotes_dir in (OSTREE_REMOTES_PATH, FLATPAK_REMOTES_PATH):
+    print('Reading config files in', remotes_dir)
+    for path in glob.iglob(remotes_dir + '/*.conf'):
+        remotes_d_confs.append(path)
+        read_config_file(repo_config, path)
```
ramcq: Simply merging in the remotes.d entries isn't sufficient to leave the user with a working remote setup if flatpak remote-add failed mid-flow due to the remotes.d error:
```root@endless:~# cat /etc/flatpak/remotes.d/flathub.conf 
[remote "flathub"]
url=https://dl.flathub.org/repo/```
(and no GPG key!)

Its existence will prevent eos-update-flatpak-repos trying to re-add it. We should either wipe them out aggressively, safe in the knowledge (?!) that eos/eos-apps have always been happily defined in config, or we need to make eos-update-flatpak-repos sense check the remotes it would add, and remove --force then re-add them if they are missing eg URL or GPG key. (see 48b1f06c8b2ae4cf09076e252b80af73318f3a5c for prior art)

eos-ostree-remotes-config
dbnicholson: Yeah, you're right. I kept this from my original script where I was trying to be respectful of an existing setting, but I think we always want `false` or things will be broken on Endless.

eos-ostree-remotes-config
dbnicholson: Good point. I wanted to keep any legit user defined entries there, but I agree that the common case is broken remotes handled by `eos-update-flatpak-repos`.

So, I think the thing to do is keep a list of remotes that will be added by `eos-update-flatpak-repos`. If we find any of those in one of the remotes.d dirs, just get rid of it entirely with the knowledge that `eos-update-flatpak-repos` will bring it back correctly.

ramcq:
Sounds good. You could just look in the same dir that the other script uses
to store it's . flatpak repo files so you'll know what will be added back.

dbnicholson:
CR ready again. I pushed a fixup patch and a couple other fixes for `eos-update-flatpak-repos`. I'll squash the fixup if this looks OK.

dbnicholson:
I squashed my fixup now for any new reviewers. If no one's taken a look in a few hours, I'll merge it.

dbnicholson:
Still no takers, so I'm going to merge this.

PR #199, opened 2018-05-07
eos-ostree-remotes-config.service: Run on every boot

dbnicholson:
Only fixing the remote configuration after an upgrade doesn't help you
if the broken remote configuration is preventing you from upgrade. Run
on every boot so that users who get into a broken configuration can
possibly fix it with a reboot.

The service still needs to have DefaultDependencies=no to support being
run before update-only services that run before sysinit.target.

https://phabricator.endlessm.com/T22258

### DIFF
--- a/eos-ostree-remotes-config.service
+++ b/eos-ostree-remotes-config.service
@@ -1,5 +1,8 @@
 [Unit]
 Description=Set /ostree/repo config option core.add-remotes-config-dir false
+
+# Even though this service doesn't need to run early, other update-only
+# services that run before sysinit.target want to run after it.
 DefaultDependencies=no
 Conflicts=shutdown.target
 Wants=local-fs.target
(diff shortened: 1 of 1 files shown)


dbnicholson:
Certainly it doesn't help anyone that's already in this situation right now. It's more to be proactive against future potential issues here. I've gotten sufficiently freaked out by the growing number of people that need manual interventions to upgrade that I thought the minor optimization of only running on upgrades was worth less than allowing people to potentially upgrade if they just rebooted.

pwithnall:
I’m not entirely convinced of the need for this. The thing I’m really worried about is if `core.add-remotes-config-dir` somehow gets unset in future because, as far as I can tell, that’s the only way we could get back into this mess. Anything which writes to `/ostree/repo/config` more than is strictly necessary makes me a bit worried.

I would be a bit happier if `eos-ostree-remotes-config` detected when the configuration was invalid (e.g. by running `ostree remote list` or by checking the value of `core.add-remotes-config-dir`) before it tried to fix anything. Then I’d be happy with running it on every boot.

PR #201, opened 2018-05-10
eos-update-flatpak-repos: Add Before= and Conflicts= for eos-updater

pwithnall:
Ensure that eos-update-flatpak-repos can’t be running at the same time
as eos-updater or eos-updater-flatpak-installer, since they all operate
on the same flatpak repos, and eos-update-flatpak-repos operates at a
fairly low level.

Use Conflicts=eos-updater.service because eos-updater is D-Bus
activated, and isn’t run on boot, so a Before=/After= ordering wouldn’t
make sense.

Use Before=eos-updater-flatpak-installer, since that *is* run at boot,
and would probably benefit from having the right repositories in place
before it runs.

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T22525

### DIFF
--- a/eos-update-flatpak-repos.service
+++ b/eos-update-flatpak-repos.service
@@ -24,6 +24,11 @@ After=eos-fix-flatpak-branches.service
 Requires=eos-extra-settled.target
 After=eos-extra-settled.target
 
+# In lieu of proper locking, we must not run at the same time as anything
+# else which touches flatpak configuration.
+Conflicts=eos-updater.service
+Before=eos-updater-flatpak-installer.service
+
 [Service]
 Type=oneshot
 RemainAfterExit=true


eos-update-flatpak-repos.service
```diff
@@ -24,6 +24,11 @@ After=eos-fix-flatpak-branches.service
 Requires=eos-extra-settled.target
 After=eos-extra-settled.target
 
+# In lieu of proper locking, we must not run at the same time as anything
+# else which touches flatpak configuration.
+Conflicts=eos-updater.service
```
dbnicholson: Does this work? Since this service has `RemainAfterExit=true`, then it seems like `eos-updater.service` won't be able to run if this one has started.

pwithnall: From my testing, it does seem to work — systemd will `SIGTERM` the service which is already running if you explicitly start the other one. D-Bus activation does seem to count as ‘explicitly starting’ `eos-updater.service`. However, you’ve identified the bit of this patch which I’m the most wary of. Further testing (on a system which isn’t dev-converted) would definitely not go amiss.

pwithnall: I’m going to fire up a non-converted eos3.3 image and test this there.

pwithnall: Seems to work as expected on a non-converted eos3.3 system ([log here](https://phabricator.endlessm.com/T22525#584521)). Please do check over the log and make sure I’m not missing anything or making a thinko.

pwithnall: Note that starting `eos-updater.service` by calling `eos-updater-ctl poll` is entirely equivalent to what `eos-autoupdater` does (from systemd’s point of view): `eos-autoupdater` calls D-Bus methods on the updater to start it. The alternative (which would potentially interact differently with `Conflicts=`) would be for `eos-updater.service` to have its own `eos-updater.timer` unit which triggered it. We’d then need to investigate how systemd timers interacted with `Conflicts=`.

That is not the case though: the timer is for `eos-autoupdater.service`.

dbnicholson:
I'm going to close this since #202 was merged.

PR #203, opened 2018-05-21
eos-repartition-mbr: invoke python3, not python (2)

wjt:
This package never had a dependency on python2; as a result, this script
broke when we removed the apparently-unused Python 2.7 from the OS.

In Python 2, sys.stdin.read() reads raw bytes. In Python 3, it reads
Unicode characters. Since we actually do want to read 8 raw bytes, we
read from the underlying raw stream at sys.stdin.buffer.

https://docs.python.org/3/library/sys.html#sys.stdin
https://phabricator.endlessm.com/T22665

### DIFF
--- a/eos-repartition-mbr
+++ b/eos-repartition-mbr
@@ -72,8 +72,8 @@ udevadm settle
 # from sfdisk's output, but we can get it from the primary header:
 # https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_table_header_.28LBA_1.29
 backup_gpt_offset=$(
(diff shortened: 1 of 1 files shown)


wjt:
This actually would have been caught by the test suite, but unfortunately the test suite is a no-op unless you run it with `EBH_ROOT_TESTS=1` since it needs to be able to run `losetup` (ie, be root) and have loads of extra tools installed which are not needed by the package itself (see my rationale in ef198b5f).

(I think our Jenkins actually does run builds as root, but this is not desired.)

PR #208, opened 2018-05-31
endless-repartition: stop creating swap partitions

rshuler:
Overall, we achieve better results with zram than with disk
swap partitions, so we are no longer going to create a swap
partition on first boot.

For existing installations, if we already created the swap partition,
we currently don't do anything to remove it, so for now we need to
continue to explicitly disable the swap partition when we enable zram.

In the future we may want to automatically remove such a partition
to reclaim disk space.

https://phabricator.endlessm.com/T21758

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -5,8 +5,6 @@
 # The purpose of this script is to identify if we are running on target
 # Endless hardware and if so, enlarges the root partition to use
 # all available space.
-# When disk space is plentiful, we also carve out some space at the end of
-# the disk, and use it to create a swap partition.
 #
 # It is important to identify that this system is an Endless-flashed device
 # that desires such treatment.
@@ -69,23 +67,20 @@ get_last_char() {
 # - A different partition numbering scheme as not all configurations have
 #   a ESP and a BIOS boot partition.
 partno=$(get_last_char ${root_part})
-
 swap_partno=$((partno + 1))
+
 case ${root_part} in
   /dev/sd??)
     root_disk=${root_part%?}
-    swap_part=${root_disk}${swap_partno}
     ;;
   /dev/*p[0-9])
     root_disk=${root_part%p?}
-    swap_part=${root_disk}p${swap_partno}
     ;;
 esac
 
 case ${orig_root_part} in
   /dev/mapper/endless-image?)
     root_disk=${orig_root_part%?}
-    swap_part=${root_disk}${swap_partno}
     using_device_mapper=1
     ;;
   /dev/loop?p?)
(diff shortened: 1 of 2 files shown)


dracut/repartition/endless-repartition.sh
```diff
@@ -147,18 +139,6 @@ if [ "$lastparttype" = "BFBFAFE7-A34F-448A-9A5B-6213EB736C22" \
   parts=$(echo "$parts" | sed '$d')
 fi
 
-# check the last partition on the disk
-lastpartno=$(echo "$parts" | sed -n -e '$ s/[^:]*\([0-9]\) :.*$/\1/p')
-
-if [ $lastpartno -eq $swap_partno ]; then
-  # already have an extra partition, perhaps we were halfway through creating
-  # a swap partition but didn't finish. Remove it to try again below.
```
wjt: I don't think we should delete this block that removes any existing swap partition at the end of the disk. As the top of the script mentions:

> This script is written carefully so that it can be interrupted after each
> operation, on next boot it should effectively continue where it left off.

Another advantage of leaving this in place is that we can reclaim space used on an already-repartitioned system by writing a systemd unit that checks for the existence of a swap partition and re-adds the GUID:55 attribute on the root partition. On next boot, this unit will kick in from the initramfs and delete it. (We'd of course want to test this thoroughly…)

wjt: In case it's not clear, the (unlikely) case I'm worried about is:

* User boots a new system with the old version of this script
* For some reason this script is interrupted after creating the swap partition but before unsetting the GUID:55 attribute
* A system update gets installed to bring in this new version of the script
* Now the new version of the script must cope with the presence of the swap partition

dracut/repartition/endless-repartition.sh
```diff
@@ -151,9 +144,17 @@ fi
...
+  # Remove it, now that we are using zram instead of disk swap.
+  # Keep in mind, though, that this code path currently only runs
+  # if we had not yet completed the repartitioning, for example
+  # in the rare case that we were halfway through creating a swap
+  # partition on a previous version of the OS, but did not finish
+  # before upgrading the OS.
   parts=$(echo "$parts" | sed '$d')
+  # Schedule a re-run of the filesystem resizing service
+  # to reclaim the extra disk space.
+  rm -f /var/eos-booted
```
wjt: This script runs from the initramfs: is the filesystem on this drive, and hence `/var`, really mounted at this point? This comment from later in the script suggests not:

```
 # Randomize root filesystem UUID
 # uninit_bg functionality prevents us from doing this later when the
 # filesystem is mounted.
```

I think we'll need to do this from the booted system, in whatever future job readds the GUID:55 marker to trigger this script running on next boot.

jprvita: Agreed

PR #211, opened 2018-06-05
Reclaim disk space from old swap partitions

jprvita:
We are dropping the use of swap partitions in favor of zram, so lets reclaim the space we have previously allocated to swap partitions for the root filesystem.

https://phabricator.endlessm.com/T22832

### DIFF
--- a/eos-enable-zram
+++ b/eos-enable-zram
@@ -3,8 +3,8 @@
 # Licensed under the GPLv2
 
 if [ $# -lt 1 -o "$1" -lt 0 -o "$1" -gt 1 ]; then
-	echo "Error: 0 or 1 must be specified" >&2
-	exit 1
+    echo "Error: 0 or 1 must be specified" >&2
+    exit 1
 fi
 
 if [ "$1" -eq 0 ]; then
--- a/eos-enable-zram
+++ b/eos-enable-zram
@@ -7,18 +7,18 @@ if [ $# -lt 1 -o "$1" -lt 0 -o "$1" -gt 1 ]; then
 	exit 1
 fi
 
-ram_size_kb=$(awk '/MemTotal/{print $2}' /proc/meminfo)
-
-if [ "$1" -ne 0 ]; then
-    if [ ! -e /sys/block/zram0 ]; then
-        modprobe zram
-    fi
-    echo $(($ram_size_kb * 2))K > /sys/block/zram0/disksize
-    # For now, disable the swap partition if in use
-    swapoff -a
-    mkswap /dev/zram0
-    swapon /dev/zram0
-else
+if [ "$1" -eq 0 ]; then
     swapoff /dev/zram0
     echo 1 > /sys/block/zram0/reset
+    exit 0
 fi
+
+if [ ! -e /sys/block/zram0 ]; then
+    modprobe zram
+fi
+
+ram_size_kb=$(awk '/MemTotal/{print $2}' /proc/meminfo)
+echo $(($ram_size_kb * 2))K > /sys/block/zram0/disksize
+mkswap /dev/zram0
+swapoff -a
+swapon /dev/zram0
--- a/eos-reclaim-swap
+++ b/eos-reclaim-swap
@@ -1,0 +1,72 @@
+#!/bin/bash -e
+# Copyright (C) 2018 Endless Mobile, Inc.
+# Licensed under the GPLv2
+#
+# TODO: Tests
+
+# List any active swap areas
+active_swaps=$(swapon --show=NAME --raw --noheadings)
+if [ -z "${active_swaps}" ]; then
+    echo "No active swap areas, nothing to be done here"
+    exit 0
+else
+    echo "Active swap areas: ${active_swaps}"
+fi
+
+# Find root partition, root partition number, root disk, and table type
+orig_root_part=$(systemctl show -p What sysroot.mount)
+orig_root_part=${orig_root_part#What=}
+root_part=$(readlink -f "${orig_root_part}")
+if [ -z ${root_part} ]; then
+    echo "Root device not found"
+    exit 1
+else
+    echo "Found root device ${root_part}"
+fi
+
+case ${root_part} in
+    /dev/sd??)
+        root_disk=${root_part%?}
+        ;;
+    /dev/*p[0-9])
+        root_disk=${root_part%p?}
+        ;;
+esac
+
+case ${orig_root_part} in
+    /dev/mapper/endless-image?)
+        root_disk=${orig_root_part%?}
+        ;;
+esac
+
+if [ -z "${root_disk}" ]; then
+    echo "No root disk found for ${root_part}"
+    exit 1
+else
+    echo "Found root disk ${root_disk}"
+fi
+
+partno=${root_part: -1}
+pt_label=$(blkid -o value -s PTTYPE $root_disk)
+if [ -z "$pt_label" ]; then
+    echo "\"blkid -o value -s PTTYPE '$root_disk'\" failed"
+    exit 1
+fi
+
+# Enable the conditions for endless-repartition.service and
+# eos-firstboot.service so they run on the next boot
+rm -f /var/eos-booted
+
+if [ "$pt_label" = "gpt" ]; then
+    attrs=$(sfdisk --part-attrs ${root_disk} ${partno})
+    attrs="GUID:55 ${attrs}"
+    echo "Setting partition attributes \"${attrs}\" on ${root_disk}${partno}"
+    sfdisk --force --part-attrs ${root_disk} ${partno} "${attrs}"
+else
+    # The only HW we expect to find with MBR layouts are Endless' own x86 HW
+    # (EC-200, Mission One, NL-3) and BIOS boot systems in which the image is
+    # converted from GPT to MBR by eos-installer. On none of these layouts
+    # partition entry 4 is used.
+    echo "Creating partition entry 4 of type DD on ${root_disk}"
+    printf "\xdd" | dd of=${root_disk} bs=1 count=1 seek=498 conv=notrunc
+fi
--- a/eos-reclaim-swap
+++ b/eos-reclaim-swap
@@ -1,8 +1,6 @@
 #!/bin/bash -e
 # Copyright (C) 2018 Endless Mobile, Inc.
 # Licensed under the GPLv2
-#
-# TODO: Tests
 
 # List any active swap areas
 active_swaps=($(swapon --show=NAME --raw --noheadings))
@@ -14,8 +12,14 @@ else
 fi
 
 # Find root partition, root partition number, root disk, and table type
-orig_root_part=$(systemctl show -p What sysroot.mount)
-orig_root_part=${orig_root_part#What=}
+if [ $# -ge 1 ]; then
+    # For testing
+    orig_root_part="$1"
+else
+    orig_root_part=$(systemctl show -p What sysroot.mount)
+    orig_root_part=${orig_root_part#What=}
+fi
+
 root_part=$(readlink -f "${orig_root_part}")
 if [ -z ${root_part} ]; then
     echo "Root device not found"
@@ -37,6 +41,9 @@ case ${orig_root_part} in
     /dev/mapper/endless-image?)
         root_disk=${orig_root_part%?}
         using_device_mapper=1
+        ;;
+    /dev/loop?p?)
+        using_loop=1
         ;;
 esac
 
@@ -132,6 +139,12 @@ fi
 # Device-mapper needs an extra prod to update the block devices
 if [ -n "$using_device_mapper" ]; then
     kpartx -u $root_disk
+    udevadm settle
+fi
+
+# Loop devices need a different extra prod
+if [ -n "$using_loop" ]; then
+    partprobe $root_disk
     udevadm settle
 fi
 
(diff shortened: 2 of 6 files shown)


eos-enable-zram
```diff
@@ -13,13 +13,79 @@ if [ "$1" -eq 0 ]; then
...
-swapoff -a
 mkswap /dev/zram0
 swapon /dev/zram0
+
+if [ -z "${active_swaps}" ]; then
+    exit 0
+fi
+
+# Find root partition, root partition number, root disk, and table type
+orig_root_part=$(systemctl show -p What sysroot.mount)
```
wjt: This all looks very familiar – should we extract this code into a function/separate program rather than copy-pasting it?

eos-enable-zram
```diff
@@ -13,13 +13,79 @@ if [ "$1" -eq 0 ]; then
...
+    /dev/mapper/endless-image?)
+        root_disk=${orig_root_part%?}
+        ;;
+esac
+
+if [ -z "${root_disk}" ]; then
+    exit 0
+fi
+
+get_last_char() {
```
wjt: This is Bash so you don't need this monstrosity:

```
$ x=foobar
$ echo ${x: -1}
r
```

eos-enable-zram
```diff
@@ -13,13 +13,79 @@ if [ "$1" -eq 0 ]; then
...
+esac
+
+case ${orig_root_part} in
+    /dev/mapper/endless-image?)
+        root_disk=${orig_root_part%?}
+        ;;
+esac
+
+if [ -z "${root_disk}" ]; then
+    exit 0
```
wjt: Shouldn't this be a non-zero exit?

eos-enable-zram
```diff
@@ -13,13 +13,79 @@ if [ "$1" -eq 0 ]; then
...
+    while [ "${#ret}" != "1" ]; do
+            ret=${ret#?}
+    done
+    echo $ret
+}
+partno=$(get_last_char ${root_part})
+pt_label=$(blkid -o value -s PTTYPE $root_disk)
+if [ -z "$pt_label" ]; then
+    echo "blkid -o value -s PTTYPE '$root_disk' failed"
+    exit 0
```
wjt: Shouldn't this be a non-zero exit?

eos-enable-zram
```diff
@@ -13,13 +13,79 @@ if [ "$1" -eq 0 ]; then
...
+# eos-firstboot.service so they run on the next boot
+if [ "$pt_label" = "gpt" ]; then
+    attrs=$(sfdisk --part-attrs ${root_disk} ${partno})
+    attrs="GUID:55 ${attrs}"
+    echo "Setting partition attributes \"${attrs}\" on ${root_disk}${partno}"
+    sfdisk --force --part-attrs ${root_disk} ${partno} "${attrs}"
+else
+    # On MBR our marker for expanding the root partition is setting partition
+    # entry 4 with the type 0xdd, which is unused by EOS. Our MBR layouts were
+    # composed of two partitions, id 1 for /boot and id 2 for the root fs, with
```
wjt: We have MBR layouts today, too, converted from GPT after running `eos-installer` on a BIOS system using `eos-repartition-mbr` (in this repo). These have 1 or 2 non-swap partitions, depending on whether you installed before or after 3.4.0. See https://phabricator.endlessm.com/w/software/x86-64-boot-paths/

I thought we also still use MBR layout on some ARM systems, but I can't find evidence of that in the image builder.

So I think this comment should be updated to make clear that MBR layouts are not just a thing of the past; and we should probably ensure there is not a 4th partition defined in the MBR before we take any of this action.

eos-enable-zram
wjt: (but maybe this is now a bit long to be a shell script anyway…)

eos-enable-zram
```diff
@@ -54,6 +54,17 @@ if [ -z "${root_disk}" ]; then
     exit 0
 fi
 
+for swap_area in "${active_swaps[@]}" ; do
+    if [ "${swap_area%?}" = "${root_disk}" ] ; then
+        echo "Found swap partition ${swap_area} on root disk"
+	found=1
```
wjt: mismatched indentation

eos-enable-zram
```diff
@@ -54,6 +54,17 @@ if [ -z "${root_disk}" ]; then
     exit 0
 fi
 
+for swap_area in "${active_swaps[@]}" ; do
+    if [ "${swap_area%?}" = "${root_disk}" ] ; then
```
wjt: Does this work in the `${root_disk}p${partnum}` case?

Could we write a test for this script?

eos-enable-zram
jprvita: Yes, this all comes from `endless-repartition.sh` which runs from the initrd, so we would need to deploy it twice anyway. While we could factor it on a separate file, I'm not sure the benefit is greater from the effort to do it. Maybe we can do it as a separate task, in the all-too-familiar technical-debt fashion? :-)

eos-enable-zram
jprvita: -ECOPYANDPASTE from a `/bin/sh` script.

eos-enable-zram
jprvita: The main purpose of this service is to enable-zram, which at this point happened successfully, so I chose to treat reclaiming the swap space as an extra, and not exit with an error if we fail to do it. Maybe this is a strong enough argument to have this a separate service (which will need to be fully done before eos-enable-zram since I'm using the active swap areas as an input)?

eos-enable-zram
jprvita: Ditto.

eos-enable-zram
jprvita: Good eye, and a good reminder for me to turn my tab highlighter back on!

eos-enable-zram
jprvita: I does not, good catch. And I'll think of a testing strategy for this.

eos-enable-zram
wjt: I also don't think there is any scenario where we would have created a 4th partition, but it would be very sad to erase a real partition by writing this marker!

eos-enable-zram
wjt: Yeah, I don't think it's worth it in that case. Maybe in a different world these would be C programs that could share an internal static library.

eos-enable-zram
wjt: I don't think having a separate service would really change things. This service is only WantedBy=multi-user.target as opposed to RequiredBy so it failing would not break the boot process; and whether the `exit 1` is in this service or another one, there'd still just be one failing service, right?

My concern is that this makes it harder to verify that this script is working correctly, whether in a (hypothetical) test suite or in manual testing on installed systems.

eos-enable-zram
wjt: FWIW the strategy I used for `eos-repartition.sh` is to make it accept arguments on the command line so it operates on files/loopback devices controlled by the test case; with the real values being used if the command line is empty. But in this case the script also looks at global system state so it may be a bit harder.

eos-enable-zram
```diff
@@ -3,22 +3,98 @@
...
+    /dev/mapper/endless-image?)
+        root_disk=${orig_root_part%?}
+        ;;
+esac
+
+if [ -z "${root_disk}" ]; then
+    exit 0
+fi
+
+for swap_area in "${active_swaps[@]}" ; do
```
wjt: On a dual-boot system with swap, are the active swaps listed as `/dev/dm-X` or `/dev/mapper/endless-image-deviceX`? If the latter you'll need to duplicate the special handling of that case here.

eos-enable-zram
jprvita: I don't think failing the service would affect the boot process at all, it is just that in the current situation it would indicate the "enable zram" service had failed when we actually had zram up and running (which could lead to confusion during other analysis).

But I hear you, so IMO the best approach is to separate this code in a separate service (which I was trying to avoid, but this grew more than I initially expected that it actually is a whole separate thing we are doing).

eos-enable-zram
jprvita: I'll need to check that, I don't have a dual-boot system available atm.

eos-reclaim-swap
```diff
@@ -0,0 +1,151 @@
...
+
+echo "Old root ${root_part} start=${root_part_start} size=${root_part_size}"
+echo "Old swap ${swap_part} start=${swap_part_start} size=${swap_part_size}"
+echo "New root ${root_part} start=${root_part_start} size=${new_root_part_size}"
+
+parts=$(echo "${parts}" | grep -v -e "type=${swap_type}")
+parts=$(echo "${parts}" | sed -e "/uuid=${root_part_uuid}/ s/size=[ \t]*${root_part_size}/size=${new_root_part_size}/")
+echo "$parts"
+echo "$parts" | sfdisk --force --no-reread $root_disk
+ret=$?
```
wjt: this line won't be reached if sfdisk doesn't return 0 because this script has `-e` set:

```
$ cat /tmp/x.sh
#!/bin/bash
set -ex
echo hi | false
echo 'exit code:' $?
$ /tmp/x.sh
+ echo hi
+ false
```

eos-reclaim-swap
```diff
@@ -0,0 +1,151 @@
...
+udevadm settle
+
+# Update SPL checksum right away, minimizing the time during which it is
+# invalid
+if [ -x /usr/sbin/amlogic-fix-spl-checksum ]; then
+    /usr/sbin/amlogic-fix-spl-checksum $root_disk
+    udevadm settle
+fi
+
+[ "$ret" != "0" ] && exit 0
```
wjt: In this case I do think we want a non-0 exit: it's an error path

eos-reclaim-swap.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Remove swap partions from root disk and reclaim their space
+Before=multi-user.target eos-enable-zram.service
+After=swap.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=/usr/sbin/eos-reclaim-swap
```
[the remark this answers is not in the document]
jprvita: Makes sense.

tests/util.py
```diff
@@ -35,7 +35,7 @@ def udevadm_settle():
 def get_lsblk_field(device, field):
     return subprocess.check_output([
         'lsblk', '--nodeps', '--noheading', '--output', field, device
-    ]).decode('utf-8').strip()
+    ], stderr=subprocess.PIPE).decode('utf-8').strip()
```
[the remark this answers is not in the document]
jprvita: I've missed `subprocess.DEVNULL` when reading the documentation, thanks. And good catch on the commit msg typo.

eos-reclaim-swap
jprvita: True, I forgot this is running on a different environment in my c&p endevour.

tests/util.py
```diff
@@ -92,3 +98,22 @@ def assert_fstype(self, partition, type_):
...
         self.assertEqual(fstype(partition), type_, msg=msg)
+
+    def assert_not_fstype(self, partition, type_):
+        '''Asserts that if partition exists it 'partition' does not contain a
+        'type_' filesystem.'''
+        msg = 'expected {} to not exist or not have type {!r}'.format(partition, type_)
+        try:
+            fs = fstype(partition)
+        except subprocess.CalledProcessError as e:
+            if (e.args[0] == 32):
```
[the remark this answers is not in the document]
wjt: Maybe add a named constant or a comment?

eos-reclaim-swap
jprvita: As you said previously, this condition will never be true as we'll never reach this point if `$ret != 0`.

eos-reclaim-swap
```diff
@@ -0,0 +1,151 @@
...
+
+if [ -z ${swap_part_start} ] ; then
+    echo "No swap partitions found on the root disk"
+    exit 0
+fi
+
+# Calculate new root partition size
+root_part_start=$(echo "${parts}" | grep -e "${root_part}" | sed -n -e "s/.*start=[ ]\+\([0-9]\+\).*$/\1/p")
+root_part_size=$(echo "${parts}" | grep -e "${root_part}" | sed -n -e "s/.*size=[ ]\+\([0-9]\+\).*$/\1/p")
+root_part_uuid=$(echo "${parts}" | grep -e "${root_part}" | sed -n -e 's/.*uuid=\(.*\),.*$/\1/p')
```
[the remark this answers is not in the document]
jprvita: It is not :-(

eos-reclaim-swap
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash -e
+# Copyright (C) 2018 Endless Mobile, Inc.
+# Licensed under the GPLv2
+
+# List any active swap areas
+active_swaps=($(swapon --show=NAME --raw --noheadings))
```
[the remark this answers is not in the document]
jprvita: Yes, that is probably what I would have done if I haven't started with the idea of using eos-repartition on the next boot, and later kind of basing the whole implementation here on that one.

eos-reclaim-swap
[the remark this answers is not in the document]
jprvita: Me too. I'll give it a try, as the time away from this task was a good reminder how obfuscated bash is by nature.

tests/util.py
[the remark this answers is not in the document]
jprvita: yes

jprvita:
Closing this for now, as this is undergoing some major rewrite.

PR #212, opened 2018-06-07
T22763: clear up / avoid old refs for flatpaks which have been migrated

ramcq:
Previous versions of eos-update-flatpak-repos left behind local ostree refs to the old origin/branch of the apps it had migrated. This meant that once flatpak had upgraded the app and pulled the new ref from the network, the old one was never cleaned up, leaving behind the commit and disk space associated with the old version of the app.

To both fix this problem, and prevent it happening again, we build a list of matching ostree refs which should shadow the installed flatpaks we are migrating, and copy the ref to the new name if the app branch or origin is changed during the migration. The remaining refs in our list at the end of the process have either just been migrated, or were migrated in the past, and are therefore deleted.

https://phabricator.endlessm.com/T22763

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -23,6 +23,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
+import fnmatch
 import glob
 import logging
 import os
@@ -33,8 +34,8 @@ from systemd import journal
 
 import gi
 gi.require_version('Flatpak', '1.0')
-from gi.repository import Flatpak
-from gi.repository import GLib
+gi.require_version('OSTree', '1.0')
+from gi.repository import Flatpak, Gio, GLib, OSTree
 
 
 def _flatpak_inst_get_remote(inst, name):
@@ -447,6 +448,16 @@ def _filter_refs(refs, name=None, prefix=None, kind=None, arch=None,
     return ret
 
 
+def _filter_ostree_refs(refs, kind, name='*', prefix=None, arch='*',
+                        branch='*', origin='*'):
+    if prefix:
+        name = '{}*'.format(prefix)
+
+    pattern = '{}:{}/{}/{}/{}'.format(origin, kind.value_nick, name, arch, branch)
+
+    return fnmatch.filter(refs, pattern)
+
+
 def _replace_in_file(path, old, new):
     _, data = GLib.file_get_contents(path)
     assert data
@@ -533,8 +544,14 @@ def _refresh_ref(inst, kind, name, arch, branch):
 def _migrate_installed_flatpaks():
     insts = Flatpak.get_system_installations()
     inst = insts[0]
-
     refs = inst.list_installed_refs()
+
+    # we need to also operate directly on the ostree refs that shadow each
+    # flatpak so we don't leave behind local refs to the old origin and branch.
+    repo = OSTree.Repo.new(inst.get_path().get_child('repo'))
+    repo.open()
+    _, ostree_refs = repo.list_refs() # dictionary of ref -> commit
+
     for migrate in FLATPAKS_TO_MIGRATE:
         kind = migrate['kind']
         name = migrate.get('name')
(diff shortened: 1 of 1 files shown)


eos-update-flatpak-repos
```diff
@@ -447,6 +448,16 @@ def _filter_refs(refs, name=None, prefix=None, kind=None, arch=None,
...
 
 
+def _filter_ostree_refs(refs, kind, name='*', prefix=None, arch='*',
+                        branch='*', origin='*'):
+    if prefix:
+        name = '{}*'.format(prefix)
+
+    pattern = '{}:{}/{}/{}/{}'.format(origin, kind.value_nick, name, arch, branch)
+
+    return fnmatch.filter(refs, pattern)
```
pwithnall: The worrier in me is worrying about the possibility of this ever matching and including `/ostree/*/*` (deployment refs). Similarly for the other `_filter_refs()` method.

eos-update-flatpak-repos
```diff
@@ -574,26 +610,63 @@ def _migrate_installed_flatpaks():
...
+                    subpaths = new_subpaths
+                    ref = _refresh_ref(inst, kind, name, arch, branch)
+
+                new_ostree_ref = '{}:{}'.format(ref.get_origin(), ref.format_ref())
+
+                # in the case we have only modified the subpaths, the ref does
+                # not change, so we should remove it from the list of old refs
+                # which will be deleted.
+                if new_ostree_ref == old_ostree_ref:
+                    logging.debug('Ostree ref {} unchanged'.format(old_ostree_ref))
```
pwithnall: Nitpick: s/Ostree/OSTree/

eos-update-flatpak-repos
```diff
@@ -574,26 +610,63 @@ def _migrate_installed_flatpaks():
...
             except Exception:
                 logging.exception('Failure applying migration to {}'
                                   .format(name))
                 continue
 
+        # any refs we copied, along with any matches for apps we migrated
+        # in the past, should be deleted.
+        for old_ostree_ref in matching_ostree_refs:
+            logging.debug('Deleting old/stale ref {}'.format(old_ostree_ref))
+            repo.set_ref_immediate(old_ostree_ref.split(':')[0],
```
pwithnall: Do you want to wrap this in a `try..except` and try to delete the other old refs, even if one of them fails?

eos-update-flatpak-repos
```diff
@@ -574,26 +610,63 @@ def _migrate_installed_flatpaks():
...
+
             except Exception:
                 logging.exception('Failure applying migration to {}'
                                   .format(name))
                 continue
 
+        # any refs we copied, along with any matches for apps we migrated
+        # in the past, should be deleted.
+        for old_ostree_ref in matching_ostree_refs:
+            logging.debug('Deleting old/stale ref {}'.format(old_ostree_ref))
```
pwithnall: Maybe log this as a message, rather than a debug message, and also include the checksum of the ref (before it’s deleted), or the new name of the ref, so that we have some scope for manually undoing any erroneous deletions if they happen in future (and we have access to the log output from this script).

eos-update-flatpak-repos
```diff
@@ -574,26 +610,63 @@ def _migrate_installed_flatpaks():
...
+            repo.set_ref_immediate(old_ostree_ref.split(':')[0],
+                                   old_ostree_ref.split(':')[1],
+                                   None)
+            del ostree_refs[old_ostree_ref]
```
pwithnall: Is the `del` necessary?

eos-update-flatpak-repos
ramcq: This... seems intensely unlikely...? In as much as you can rely on anything in Python, provided kind is a FlafpakRefKind, then kind.value_nick can only ever be 'app' or 'runtime', and the /ostree/* refs do not have origin: prefixes. The other refs are FlatpakRefs so examine installed apps and their deploy files.

pwithnall: Maybe something like `assert origin != ''` and `assert kind.value_nick in ['app', 'runtime']`. I may be being over-paranoid, but I’d prefer that over this script silently deleting an important ref in future if something changes in flatpak.

pwithnall:
The fixups look good to me. Add some `assert`s, squash and push at your leisure.

ramcq:
:boom:

PR #220, opened 2018-07-05
T22548: remove legacy fixups/transitions

ramcq:
Systems running eos3a (>= 3.4) will have applied all of these transitions under eos3 (<= 3.3) so they are no longer required.

https://phabricator.endlessm.com/T22548


eos-update-flatpak-repos
```diff
@@ -709,12 +615,5 @@ if __name__ == '__main__':
     # ensure default remotes exist, such as eos-sdk and flathub
     _add_flatpak_repos()
 
-    # remove unused remotes, such as eos-external-apps and gnome-apps
-    _remove_remotes()
-
-    # clear away legacy eos-external-apps repo and runtimes (T14442)
-    _remove_old_external_apps_repo()
-    _remove_runtimes()
```
pwithnall: Nitpick: I’d just empty the lists which drive these functions, since we might want to remove other repos in future, and doing archaeology to resurrect the code for that would be a pain.

tmpfiles.d/eos-boot-helper.conf
```diff
@@ -0,0 +1,3 @@
+r /var/eos-extra-resize
+r /var/lib/.eos-migrate-image-defaults
+r /var/lib/misc/eos-dir-mode-700
```
pwithnall: This file could do with some comments for posterity.

eos-update-flatpak-repos
```diff
@@ -709,12 +615,5 @@ if __name__ == '__main__':
     # ensure default remotes exist, such as eos-sdk and flathub
     _add_flatpak_repos()
 
-    # remove unused remotes, such as eos-external-apps and gnome-apps
-    _remove_remotes()
```
pwithnall: Just checking: have you checked that all the code which is being deleted from `eos-update-flatpak-repos` has not been modified since the 3.4.0 release? I recall that we’ve been modifying it a fair amount recently.

ramcq:
Fixed everything. The deletions from eos-update-flatpak-repos do not include any added transitions/fixes in 3.4.

eos-update-flatpak-repos
```diff
@@ -104,22 +99,8 @@ def _remove_remotes():
...
-EXT_APPS_REPO = '/var/lib/flatpak-external-apps'
-
-
-def _remove_old_external_apps_repo():
-    if os.path.isdir(EXT_APPS_REPO):
-        logging.info("Removing repo {}".format(EXT_APPS_REPO))
-        shutil.rmtree(EXT_APPS_REPO, ignore_errors=True)
-
-
+# dict of remote -> list of runtimes
```
pwithnall: Nitpick: Technically the old value was `remote → set of runtimes`

PR #225, opened 2018-07-26
nvidia-graphics-setup: Respect modprobe.blacklist=nouveau

jprvita:
If modprobe.blacklist=nouveau is present on the kernel command line
nvidia-graphics-setup should not try to load the nouveau module.

https://phabricator.endlessm.com/T23272

### DIFF
--- a/nvidia/nvidia-graphics-setup
+++ b/nvidia/nvidia-graphics-setup
@@ -88,9 +88,30 @@ nvidia_blacklist_check_by_pci_id() {
   return 0
 }
 
+# Return 1 if module name passed as argument is blacklisted through the kernel
+# command line with modprobe.blacklist. Note that modprobe.blacklist accepts a
+# comma-separated list of modules, i.e. modprobe.blacklist=nouveau,nvidia.
+check_cmdline_blacklist() {
+  module=${1}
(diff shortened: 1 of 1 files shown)


nvidia/nvidia-graphics-setup
```diff
@@ -88,9 +88,29 @@ nvidia_blacklist_check_by_pci_id() {
   return 0
 }
 
+# Return 1 if module name passed as argument is blacklisted through the kernel
+# command line with modprobe.blacklist=module
```
dsd: Please note here that you also detect if its blacklisted via the modprobe.blacklist=a,b,c notation. Otherwise at initial glance this seems like a painful way just to do the simple string match that is mentioned.

nvidia/nvidia-graphics-setup
```diff
@@ -88,9 +88,29 @@ nvidia_blacklist_check_by_pci_id() {
...
+check_cmdline_blacklist() {
+  module=${1}
+  set -- $(cat /proc/cmdline)
+  for param in "$@"; do
```
dsd: for param in "$(</proc/cmdline)"

PR #231, opened 2018-09-04
udev/dracut: enable BFQ by default on disk devices

ramcq:
Set the scheduler to BFQ on block devices with a device type of disk, and add
the bfq module to the initrd so this udev rule works from early boot. This will
only impact SCSI devices on Endless systems, where we have switched the default
of the SCSI disk devices to the new MQ scheduler framework, which is a
pre-requisite of BFQ. Other block devices such as MMC which don't currently use
the MQ framework can't access BFQ, in which case this action is a no-op and the
current in-kernel default of CFQ remains.

https://phabricator.endlessm.com/T23725

### DIFF
--- a/dracut/endless.conf
+++ b/dracut/endless.conf
@@ -9,3 +9,7 @@ early_microcode="yes"
 # between nouveau and nvidia.
 # (nouveau cannot be unloaded after it has bound to a device)
 omit_drivers="nouveau"
+
+# Ship the bfq IO scheduler module in the initrd, as it's relied upon by
+# our udev rules that detect block devices
+add_drivers+="bfq"
--- a/50-disk-bfq.rules
+++ b/50-disk-bfq.rules
@@ -1,0 +1,1 @@
+ACTION=="add|change", SUBSYSTEM=="block", ENV{DEVTYPE}=="disk", ATTR{queue/scheduler}="bfq"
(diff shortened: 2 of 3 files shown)


dracut/endless.conf
```diff
@@ -9,3 +9,7 @@ early_microcode="yes"
 # between nouveau and nvidia.
 # (nouveau cannot be unloaded after it has bound to a device)
 omit_drivers="nouveau"
+
+# Ship the bfq IO scheduler module in the initrd, as it's relied upon by
+# our udev rules that detect block devices
+add_drivers+="bfq"
```
wjt: Does this also cause the module to be loaded? Or is that caused by the udev rule? In my testing, adding `scsi_mod.use_mq_default=1` did not cause `bfq` to be loaded and (IIRC) nor did echoing `bfq` into `/sys/block/sda/queue/scheduler` (but I could be wrong).

ramcq: I don't believe this causes the module to be loaded, just to be included in the initrd. echoing bfq into the queue/scheduler file did that for me. I believe this is achieved by a $foo-iosched alias which is provided by the bfq module. Some other modules are literally named $foo-iosched.

wjt: Huh – apparently I did not recall correctly. I've tested locally that the module is loaded I write to `queue/scheduler`.

PR #232, opened 2018-09-10
Support virtio disks as well as SCSI disks

pwithnall:
When running tests in OpenQA (which runs in QEMU), the disk uses the
virtio driver, whose devices are exposed as /dev/vd[a-z][0-9], rather
than /dev/sd[a-z][0-9].

Ensure that the disk repartitioning and resizing code handles these
device names. Critically, this ensures that the Endless partition is
resized to fill the virtual disk, rather than remaining about 6GB big
(which fills up rather fast and hence prevents any flatpaks from being
installed in OpenQA tests).

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T23739

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -66,11 +66,12 @@ get_last_char() {
 # - SCSI/MMC/NVMe block devices with different naming conventions
 # - A different partition numbering scheme as not all configurations have
 #   a ESP and a BIOS boot partition.
+# - Virtual block devices for OpenQA
 partno=$(get_last_char ${root_part})
 swap_partno=$((partno + 1))
 
 case ${root_part} in
-  /dev/sd??)
+  /dev/[sv]d??)
     root_disk=${root_part%?}
     ;;
   /dev/*p[0-9])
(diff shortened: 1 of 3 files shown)


dracut/repartition/endless-repartition.sh
```diff
@@ -66,11 +66,12 @@ get_last_char() {
 # - SCSI/MMC/NVMe block devices with different naming conventions
 # - A different partition numbering scheme as not all configurations have
 #   a ESP and a BIOS boot partition.
+# - Virtual block devices for OpenQA
 partno=$(get_last_char ${root_part})
 swap_partno=$((partno + 1))
 
 case ${root_part} in
-  /dev/sd??)
+  /dev/[sv]d??)
```
wjt: I've always thought this logic was a bit fragile. I wonder if we should instead be using `lsblk --inverse $root_part` to find the parent device?

eos-reclaim-swap
```diff
@@ -29,7 +29,8 @@ class SwapRemover():
         disk = None
 
         # /dev/sd[a-z][0-9]
-        if p.startswith('/dev/sd'):
+        # /dev/vd[a-z][0-9]
+        if p.startswith('/dev/sd') or p.startswith('/dev/vd'):
```
wjt: This is a bit belt-and-braces because since the repartition script has never worked on `/dev/vd*`, we can never have created swap there. But maybe simpler to just handle this case than to explain it.

PR #233, opened 2018-09-14
eos-update-flatpak-repos: check for and handle already-migrated apps

ramcq:
Changes of origin don't change the deployed path on disk, and our
origin-changing code is idempotent, so we only need to check if something is
installed in the case the branch is being changed. If the ref we are migrating
to already exists, uninstall the old one.

https://phabricator.endlessm.com/T21141

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -635,6 +635,25 @@ def _migrate_installed_flatpaks():
             try:
                 new_branch = migrate.get('new-branch')
                 if new_branch:
+                    try:
+                        new_ref = inst.get_installed_ref(kind, name, arch,
+                                                         new_branch)
+                    except GLib.GError as e:
+                        if not e.matches(Flatpak.error_quark(),
+                                         Flatpak.Error.NOT_INSTALLED):
+                            raise
+                    else:
+                        logging.info('{} already installed, removing {}'
+                                     .format(new_ref.format_ref(),
+                                             ref.format_ref()))
+                        inst.uninstall_full(Flatpak.UninstallFlags.NO_PRUNE,
+                                            kind, name, arch, branch)
+
+                        # do not continue to do any origin migrations if we already
+                        # have the app installed. the ostree ref is already part
+                        # of matching_ostree_refs so will be removed at the end.
+                        continue
+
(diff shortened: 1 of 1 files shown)


eos-update-flatpak-repos
```diff
@@ -635,9 +635,29 @@ def _migrate_installed_flatpaks():
...
+                   new_ref = None
+                   try:
+                       new_ref = inst.get_installed_ref(kind, name, arch,
+                                                        new_branch)
+                   except GLib.GError as e:
+                       if e.matches(Flatpak.error_quark(),
+                                    Flatpak.Error.NOT_INSTALLED):
+                           pass
+                       else:
+                           raise
```
wjt:
```
if not e.matches(...):
    raise
```
?

wjt: If you were feeling fancy you could write:

```python
try:
    new_ref = inst.get_installed_ref(...)
except GLib.GError as e:
    if not e.matches(...):
        raise
    # handle not-installed case
else:
    # handle already-installed case
```

but I'm not sure how popular `else` clauses are on `try` blocks.

eos-update-flatpak-repos
```diff
@@ -635,9 +635,29 @@ def _migrate_installed_flatpaks():
...
+
+                   if new_ref:
+                       logging.info('{} already installed, removing {}'
+                                    .format(new_ref.format_ref(),
+                                            ref.format_ref()))
+                       inst.uninstall_full(Flatpak.UninstallFlags.NO_PRUNE,
+                                           kind, name, arch, branch)
```
wjt: Why do we continue on into the code that updates the deploy file in this case? I don't think it's quite right because `new_ref`'s origin may in theory be neither `old-origin` nor `new-origin`.

eos-update-flatpak-repos
```diff
@@ -635,6 +635,26 @@ def _migrate_installed_flatpaks():
             try:
                 new_branch = migrate.get('new-branch')
                 if new_branch:
+                    new_ref = None
```
wjt: This definition is not needed, the scope of the definition inside `try` is enough and it's not used outside the `else` branch

ramcq: yayaya

PR #235, opened 2018-09-17
eos-update-flatpak-repos: use Flatpak API to change branch names

ramcq:
Copy the OSTree refs ahead of any Flatpak migrations so that we can use a NO_PULL deploy to install the Flatpak with the new branch name, as it will be created in a new folder. At a slight performance
cost (as the app will now be redeployed), the hair-raising update branch function which moved folders around and monkey patched the exported files behind Flatpak's back can be removed, significantly improving robustness and maintainability.

We now only directly access the deploy file in the case we are modifying the origin, but this code path is avoided if we've just done redeploy from the new ref, because it will already have the new origin. As this is the last and only action performed outside of the Flatpak API, we don't need to drop the cache or
refresh the ref.

https://phabricator.endlessm.com/T22902

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -642,6 +642,18 @@ def _migrate_installed_flatpaks():
             assert (old_ostree_ref != new_ostree_ref)
 
             try:
+                # copy the old ref to the new one, pointing at the same commit
+                if old_ostree_ref in ostree_refs and \
+                   not new_ostree_ref in ostree_refs:
+                    logging.info('Copying OSTree ref {} [{}] to {}'
+                                 .format(old_ostree_ref,
+                                         ostree_refs[old_ostree_ref],
+                                         new_ostree_ref))
+                    repo.set_ref_immediate(new_origin,
+                                           new_ostree_ref.split(':')[1],
+                                           ostree_refs[old_ostree_ref])
+                    ostree_refs[new_ostree_ref] = ostree_refs[old_ostree_ref]
+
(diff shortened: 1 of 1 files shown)


eos-update-flatpak-repos
```diff
@@ -624,74 +548,71 @@ def _migrate_installed_flatpaks():
...
-
-                    _update_branch(ref, new_branch, refs)
-                    branch = new_branch
-                    ref = _refresh_ref(inst, kind, name, arch, branch)
 
-                deploy_dir = ref.get_deploy_dir()
-                deploy = os.path.join(deploy_dir, 'deploy')
+                        logging.info('Deploying with new ref: {}'.format(new_ostree_ref))
+                        ref = inst.install_full(Flatpak.InstallFlags.NO_PULL, new_origin,
+                                                kind, name, arch, new_branch)
```
ramcq: So... Flatpak checks xa.ref when deploying, and emits a WARNING if this has the wrong branch ID. This warning is promoted to a fatal error if collection IDs are set, which is sadly a merge blocker for this branch.

ramcq: Fixed using code from ostree-upload-server. Thanks @dbnicholson !

pwithnall:
(Not reviewing, as per [backpedalling on the plan](https://phabricator.endlessm.com/T22902#627858). Let me know if you still want a review.)

wjt:
Superseded by #236 et al.

PR #237, opened 2018-09-21
Run flake8 during build

wjt:
I was reviewing https://github.com/endlessm/eos-boot-helper/pull/236 and found bugs just by running `flake8`, so let's automate that.


wjt:
NB the PR will fail; I plan to rebase after https://github.com/endlessm/eos-boot-helper/pull/236 is in and fix up any remaining style complaints.

PR #238, opened 2018-09-21
eos-update-flatpak-repos: configure mass migration of eos-apps to flathub

ramcq:

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -1132,21 +1132,23 @@ def rename_exports(repo, mtree, old_id, new_id, path='export'):
                                      GLib.KeyFileError.KEY_NOT_FOUND):
                         raise
 
-                # update the places inside the desktop file where the ID is embedded
-                for key in [GLib.KEY_FILE_DESKTOP_KEY_ICON,
-                            'X-Endless-SplashBackground']:
-                    try:
-                        value = desktop_file.get_string(
-                            GLib.KEY_FILE_DESKTOP_GROUP, key
-                        )
-                    except GLib.GError as e:
-                        if not e.matches(GLib.KeyFile.error_quark(),
-                                         GLib.KeyFileError.KEY_NOT_FOUND):
-                            raise
-                        continue
-
+                # update the icon name which matches the ID
+                try:
+                    value = desktop_file.get_string(
+                        GLib.KEY_FILE_DESKTOP_GROUP,
+                        GLib.KEY_FILE_DESKTOP_KEY_ICON
+                    )
+                except GLib.GError as e:
+                    if not e.matches(GLib.KeyFile.error_quark(),
+                                     GLib.KeyFileError.KEY_NOT_FOUND):
+                        raise
+                else:
                     value = value.replace(old_id, new_id)
-                    desktop_file.set_string(GLib.KEY_FILE_DESKTOP_GROUP, key, value)
+                    desktop_file.set_string(
+                        GLib.KEY_FILE_DESKTOP_GROUP,
+                        GLib.KEY_FILE_DESKTOP_KEY_ICON,
+                        value
+                    )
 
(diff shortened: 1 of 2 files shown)


eos-update-flatpak-repos
```diff
@@ -467,6 +467,556 @@ FLATPAKS_TO_MIGRATE = [
         'new-branch': 'stable',
         'old-origin': 'eos-apps',
         'new-origin': 'flathub'
+    },
+    # migrate 61 legacy eos-apps to flathub (T23856)
+    {
+        'name': 'com.arduino.App',
+        'new-name', 'cc.arduino.arduinoide',
```
wjt: The reason I requested review on #237 before reviewing this PR is that every migration dictionary added here has the same syntax error, namely a comma rather than a colon after `'new-name'`.

ramcq: Lulz. Yeah. I've not tried this yet. Didn't want to unleash it on any system until we had the X-Flatpak-RenamedFrom ducks in a row.

ramcq: Thats better... the new syntax checker likes it too.

eos-update-flatpak-repos
```diff
@@ -467,6 +467,556 @@ FLATPAKS_TO_MIGRATE = [
...
+        'name': 'org.armagetronad.Armagetronad',
+        'new-name': 'org.armagetronad.ArmagetronAdvanced',
+        'kind': Flatpak.RefKind.APP,
+        'old-branch': 'eos3',
+        'new-branch': 'stable',
+        'old-origin': 'eos-apps',
+        'new-origin': 'flathub'
+    },
+    {
+        'name': 'org.codeblocks.App',
```
wjt: As commented elsewhere, this app's manifest defines a Debug and a Locale extension, but neither appears to be present in the `eos-apps` repository, so no great loss that we don't mangle those extensions in the manifest.

eos-update-flatpak-repos
```diff
@@ -1132,21 +1132,24 @@ def rename_exports(repo, mtree, old_id, new_id, path='export'):
...
+                try:
+                    value = desktop_file.get_string(
+                        GLib.KEY_FILE_DESKTOP_GROUP,
+                        GLib.KEY_FILE_DESKTOP_KEY_ICON
+                    )
+                except GLib.GError as e:
+                    if not e.matches(GLib.KeyFile.error_quark(),
+                                     GLib.KeyFileError.KEY_NOT_FOUND):
+                        raise
+                    continue
```
wjt: This is wrong, it will skip the x-flatpak-renamedfrom block. Move the 'replace and set' stuff into an `else:` block.

wjt: .oO( how did the test pass? )

wjt: Added a (failing) test case that catches this.

wjt: (and a fix to be squashed)

eos-update-flatpak-repos
ramcq: I can only assume they were released to eos-apps at a time that the release-apps script was blissfully unaware of app extensions. Luckily for us... :)

eos-update-flatpak-repos
ramcq: Lucky one of us is awake! :)

ramcq:
> Looks plausible. I mechanically checked that all (old-origin, name//branch) and (new-origin, new-name//new-branch) refs exist with flatpak remote-info --app .... Interestingly it shows "download size" and "installed size" both as 0 for most eos-apps versions, along with a warning about "Commit has no flatpak metadata".

Thank you! I think there is some kind of embedded xa.stuff that gets put into commits noawadays which it didn't used to, so apps which were exported with an old enough version of flatpak are missing various fields from remote-info.

ramcq:
Squashed. Good to merge?

PR #239, opened 2018-09-23
eos-update-flatpak-repos: update X-Flatpak-RenamedFrom

wjt:
Our Flatpak is patched to propagate a previous-ids property from old
to new deploy file when deploying an app update, and to add all previous
IDs to the X-Flatpak-RenamedFrom key in the process.

However, in this script we only set previous-ids after deploying the
patched version of the app (under its new name). In any case, as far as
Flatpak is concerned, this is a wholly new app being installed, so there
would be no previous deployment to carry previous-ids forward from.

Since we are patching the .desktop files anyway while doing the
migration, we can just implement the same logic to update
X-Flatpak-RenamedFrom here, and add a test for this and other .desktop
file mangling.

I have not tested this on a real system, but in my defence I did write a test.

https://phabricator.endlessm.com/T23845

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -590,6 +590,22 @@ def rename_exports(repo, mtree, old_id, new_id, path='export'):
                     value = value.replace(old_id, new_id)
                     desktop_file.set_string(GLib.KEY_FILE_DESKTOP_GROUP, key, value)
 
+                # Add old name to X-Flatpak-RenamedFrom
+                try:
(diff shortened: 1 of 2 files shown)


wjt:
I noticed while writing this that we are assuming that none of the apps that need migrating have extensions. Is that right?

ramcq:
> I noticed while writing this that we are assuming that none of the apps that need migrating have extensions. Is that right?

There is a runtime error enforcing that, yeah. None of the deb bundled apps have extensions, the ones to check are the .json flatpak builder unicorns.

PR #242, opened 2018-10-02
T23856: fixups for 2->1 and no-rename migrations

ramcq:

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -1497,6 +1497,7 @@ def _migrate_installed_flatpaks():
                                                     Flatpak.InstallFlags.NO_TRIGGERS,
                                                     new_origin, kind, new_name,
                                                     arch, new_branch)
+                        refs.append(new_ref)
 
                     logging.info('Uninstalling old ref: {}'.format(old_ostree_ref))
                     inst.uninstall_full(Flatpak.UninstallFlags.NO_PRUNE |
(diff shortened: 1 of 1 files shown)


eos-update-flatpak-repos
```diff
@@ -1497,6 +1497,7 @@ def _migrate_installed_flatpaks():
                                                     Flatpak.InstallFlags.NO_TRIGGERS,
                                                     new_origin, kind, new_name,
                                                     arch, new_branch)
+                        refs.append(new_ref)
```
wjt: I was curious which app it is, and determined that it's Aisleriot. Weird to replace 2 with 1 but okay.

ramcq: Yeah - well... perhaps in hindsight it was weird to take the same app and duplicate it in its entirety under two names just to pass two different command line arguments?

Especially when it places an argv string in the flatpak command= metadata field, which actually expects a binary name, meaning that you can't launch them from the App Center either - https://phabricator.endlessm.com/T23856. :)

dbnicholson: This goes back to our earliest app bundling days. It was determined that we should split solitaire and freecell from aisleriot into 2 apps because someone thought it was weird to have to know how to get into solitaire or freecell mode even though they're distinctly two different card games. https://phabricator.endlessm.com/T3227 but that doesn't have the rationale, unfortunately.

ramcq: Sure, it's a bit weird. Maybe we could try and push some additional desktop files upstream into Aisleriot, but...

PR #245, opened 2018-11-13
eos-reclaim-swap: avoid cylinder alignment for DOS partition tables

dsd:
For DOS partition tables, the msdos_partition_align() function called
in the disk.maximizePartition() codepath was modifying the partition
start, aligning it based on BIOS cylinder calculations. Drop that
constraint so that the sector will be unchanged.

This was causing the new partition to be written with a different start
sector (144585 instead of 131072), but here we are only trying to modify
the end. Add a sanity check to prevent that from happening again for
any reason.

https://phabricator.endlessm.com/T24454

### DIFF
--- a/eos-reclaim-swap
+++ b/eos-reclaim-swap
@@ -55,6 +55,8 @@ class SwapRemover():
         '''
         dev = parted.getDevice(disk_path)
         disk = parted.newDisk(dev)
+        if disk.type == "msdos":
+            disk.unsetFlag(parted.DISK_CYLINDER_ALIGNMENT)
         parts = disk.partitions
 
         for p, part in enumerate(parts):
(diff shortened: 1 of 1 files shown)


eos-reclaim-swap
```diff
@@ -55,6 +55,8 @@ class SwapRemover():
         '''
         dev = parted.getDevice(disk_path)
         disk = parted.newDisk(dev)
+        if disk.type == "msdos":
```
jprvita: Looks good to me, but maybe we should also double check if the flag is available?
```suggestion
        if disk.type == "msdos" and disk.isFlagAvailable(parted.DISK_CYLINDER_ALIGNMENT):
```

PR #247, opened 2018-11-14
Add eos-factory-reset systemd units

dnohales:
This is to complete https://github.com/endlessm/eos-meta/pull/536 . I wasn't sure if you wanted the `eos-factory-reset` scripts here or in `eos-meta` so I leaved it in `eos-meta`. But it'd be probably a good idea to move those to this repo.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -15,6 +15,7 @@ dist_systemdunit_DATA = \
 	eos-config-journal.service \
 	eos-enable-zram.service \
 	eos-extra-resize.service \
+	eos-factory-reset.service \
 	eos-firewall-localonly.service \
 	eos-firstboot.service \
 	eos-image-boot-dm-setup.service \
@@ -74,6 +75,7 @@ dist_sbin_SCRIPTS = \
 	eos-config-journal \
 	eos-enable-zram \
 	eos-extra-resize \
+	eos-factory-reset \
 	eos-firewall-localonly \
 	eos-firstboot \
 	eos-image-boot-dm-setup \
@@ -81,6 +83,12 @@ dist_sbin_SCRIPTS = \
 	eos-reclaim-swap \
 	eos-repartition-mbr \
 	eos-update-flatpak-repos \
+	$(NULL)
+
+eosfactoryresetdir = $(prefix)/lib/eos-factory-reset
+dist_eosfactoryreset_SCRIPTS = \
+	eos-factory-reset-helper \
+	eos-factory-reset-users-helper \
 	$(NULL)
 
 tmpfilesdir = $(prefix)/lib/tmpfiles.d
--- a/eos-factory-reset
+++ b/eos-factory-reset
@@ -1,0 +1,89 @@
+#!/bin/bash -e
+
+cmd_name=$(basename "$0")
+
+usage () {
+    cat <<EOF
+Usage: $cmd_name
+Reboot the machine and perform a factory reset.
+
+  -u               reset all users only
+  -a               make a full factory reset
+  -f, --force      don't ask for user confirmation
+  -h, --help       display this help and exit
+
+EOF
+}
+
+args=$(getopt -o u,a,f,h -l "force,help" -n "$cmd_name" -- "$@") || exit 1
+
+eval set -- "$args"
+
+reset_stages=""
+ask_confirmation=1
+
+while true; do
+    case "$1" in
+        -u)
+            reset_stages="USERS $reset_stages"
+            shift
+            ;;
+        -a)
+            reset_stages="USERS"
+            shift
+            ;;
+        -f|--force)
+            ask_confirmation=0
+            shift
+            ;;
+        -h|--help)
+            usage
+            exit 0
+            ;;
+        --)
+            shift
+            ;;
+        *)
+            break
+            ;;
+    esac
+done
+
+# Check that script was run with superuser privileges.
+if [[ $EUID != 0 ]]; then
+    echo "$cmd_name requires superuser privileges."
+    exit 1
+fi
+
+if [ -z $reset_stages ]; then
+    echo "$cmd_name: No stages were selected"
+    exit 1
+fi
+
+
+if [ $ask_confirmation -eq 1 ]; then
+    echo "This tool will perform a factory reset, which will delete all"
+    echo "users and their documents, images, etc. forever."
+    echo ""
+    echo "During this process, you'll see a black screen for a few seconds"
+    echo "and then the system will be restarted."
+    echo ""
+    echo "Selected factory reset stages: $reset_stages"
+    echo ""
+    echo "WARNING: These changes are IRREVERSIBLE."
+    echo ""
+
+    read -rp "Are you sure you want to proceed? [y/n] "
+    response="${REPLY,,}" # to lower
+    if [[ ! "$response" =~ ^(yes|y)$ ]]; then
+        exit 1
+    fi
+fi
+
+flag_file=/var/.eos-factory-reset
+echo -n > $flag_file
+for reset_stage in $reset_stages; do
+    echo "EOS_FACTORY_RESET_$reset_stage=1" >> $flag_file
+done
+
+systemctl reboot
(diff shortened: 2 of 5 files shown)


eos-factory-reset.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Endless factory reset service
+ConditionPathExists=/var/eos-factory-reset
+Conflicts=getty@tty1.service gdm.service plymouth-start.service
+After=systemd-user-sessions.service getty@tty1.service plymouth-quit.service gdm.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/eos-factory-reset/eos-factory-reset-helper
```
wjt: I feel very strongly that the `.service` file should live in the same repository as the script it runs.

Makefile.am
```diff
@@ -53,10 +55,14 @@ install-data-hook:
 	$(MKDIR_P) '$(DESTDIR)$(systemdunitdir)'
 	$(INSTALL_DATA) $(srcdir)/var-endless_x2dextra.mount \
 	  '$(DESTDIR)$(systemdunitdir)/var-endless\x2dextra.mount'
+	$(MKDIR_P) '$(DESTDIR)$(systemdunitdir)/eos-factory-reset.target.wants'
+	ln -s '../eos-factory-reset.service' \
+	  '$(DESTDIR)$(systemdunitdir)/eos-factory-reset.target.wants'
```
wjt: Do other packages which define new targets do this? I believe that (the equivalent of) `systemd preset-all` is run at boot (on unconverted systems) which enables any unit not explicitly disabled in `50-eos.preset` (in this repo). eos-factory-helper, for example, defines a new .target and .service but doesn't do anything special beyond installing the files.

dnohales:
Moved the scripts here and fixed the nitpiks mentioned in the other PR.

eos-factory-reset-users-helper
```diff
@@ -0,0 +1,26 @@
...
+
+            if [ $? -eq 0 ]; then
+                break
+            else
+                if [ $delete_tries -gt 20 ]; then
+                    echo "Giving up on deleting $user ($user_id)"
+                    break
+                fi
+                sleep 2
+                delete_tries=$((delete_tries + 1))
```
wjt: I don't think Rob's question at https://github.com/endlessm/eos-meta/pull/536#discussion_r233483330 is answered. I agree that it should not be necessary to have a retry loop.

eos-factory-reset.service
```diff
@@ -0,0 +1,15 @@
...
+ExecStartPost=systemctl reboot
```
wjt: I believe this step will not be run if the `ExecStart=` line exits unsuccessfully, so on error the system will just stay at the blank/splash screen forever until the user pulls the plug.

eos-factory-reset.service
```diff
@@ -0,0 +1,15 @@
...
+StandardOutput=journal
+StandardError=journal
```
wjt: These are the defaults; why are they explicitly specified?

eos-factory-reset-users-helper
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+IFS=$'\n'
+for user_row in $(cat /etc/passwd); do
```
wjt: https://stackoverflow.com/a/37947526/173314

eos-factory-reset
```diff
@@ -0,0 +1,93 @@
...
+
+cmd_name=$(basename $0)
+
+usage () {
+    cat <<EOF
+Usage: $cmd_name
+Reboot the machine and perform a factory reset.
+
+  -u               reset all users only
+  -a               make a full factory reset
```
wjt: I guess this is for future-proofing but at present `-u` and `-a` are equivalent, and you have to specify one or the other. Seems a bit weird.

dnohales:
@wjt fixed all you've said but I still cannot figure out how to delete the users without the loop approach.

eos-factory-reset-users-helper
wjt: Try running `systemd-cgls` before the loop and see what the lingering processes are?

wjt: I wonder if it might be enough to add `Conflicts=user@.service` – I don't know if you can use that syntax to match all instantiations of a template, but maybe you can.

jprvita: @wjt's idea is sensible, although we should probably look at either `user.slice` or the `user-<uid>.slice` for each existing user in the system. In any case, with our current logind configuration (inherited from Debian) we may wait forever if the user started a process in `screen`, for example. This can be solved changing `KillUserProcesses` in logind.conf.

But after giving some thought to all of this, I think ideally we should perform the factory reset in a more controlled environment. This way we can be sure no user processes are running nor will spawned (if the user logs in over ssh, for example), like during the next boot or during the shutdown process. On the factory testing process this is not problem because one of the requirements is that no users exist in the system.

dsd: I agree that care should be taken to produce a highly deterministic environment under which this script is ran, where "nothing" is running except from this script. From a distance, I would have expected `systemctl isolate` to achieve this well enough:

>           This is similar to changing the runlevel in a traditional init
>           system. The isolate command will immediately stop processes that
>           are not enabled in the new unit, possibly including the graphical
>           environment or terminal you are currently using.

i.e. the concern about preventing users logging in over ssh should be a simple/natural part of that, since you would not include sshd in the factory-reset target being isolated here.

But if that's not the case (i.e. sshd and/or user session stuff is still running) then I also suggest going deeper into this so that we can understand what is happening and if `systemctl isolate` is or is not a suitable approach.

dnohales:
So, I changed the approach by enabling `eos-factory-reset.service` for `multi-user.target` by default, it should not always run since the `ConditionPathExists` and the conditions imposed by `eos-factory-reset-helper` script. Now `eos-factory-reset` only creates `/var/eos-factory-reset` and reboot the system.

I'm still figuring out what are the correct values for `Before`, `After`, `Conflicts` and `Requires`, to make it predictable, or even if `multi-user` is the proper target.

Please let me know what you think. I don't know if having always enabled could mess something or you feel that it's unsafe (maybe the flag file could not be removed in some situations).

Thanks!

ptomato:
I took a look at this, but I didn't understand from reading the code or comments whether the comment about using `systemctl isolate` was addressed.

dsd:
There is no commit message here.

If the new approach is able to guarantee the fixed & minimal runtime environment then this should be explained/documented/referenced.

At an initial and distant glance, I am not convinced that it brings in those guarantees. It is WantedBy multi-user.target and Before user.slice. Does that mean that everything else in multi-user.target has After=user.slice, therefore guaranteeing that this runs earlier & in isolation? If so, is this a guarantee that we can rely on for the forseeable future?

dnohales:
@dsd @jprvita, does it make sense now?

eos-factory-reset
```diff
@@ -0,0 +1,89 @@
...
+    echo ""
+
+    read -rp "Are you sure you want to proceed? [y/n] "
+    response="${REPLY,,}" # to lower
+    if [[ ! "$response" =~ ^(yes|y)$ ]]; then
+        exit 1
+    fi
+fi
+
+flag_file=/var/eos-factory-reset
```
cosimoc: Perhaps we can make this a hidden file? (e.g. `/var/.eos-factory-reset`, similar to how we have `/var/.updated`)

eos-factory-reset-users-helper
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+IFS=':'
+while read -r user _ uid _ _ _ _; do
+    if [[ $uid -ge 1000 && $uid -le 1999 ]]; then
```
cosimoc: It would be nice to read these from a place such as `/etc/adduser.conf`

eos-factory-reset.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Endless factory reset service
+ConditionPathExists=/var/.eos-factory-reset
+DefaultDependencies=no
```
jprvita: When using `DefaultDependencies=no` you need to explicitly add `Conflicts=shutdown.target`.

dnohales:
Sorry, forgot to say, I've updated the code.

PR #248, opened 2018-12-02
T24741: support extra data and migrate Skype

ramcq:
Add support to eos-update-flatpak-repos to migrate apps with extra data, so that we can migrate the Skype app to the Flathub version.

https://phabricator.endlessm.com/T24741

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -1002,6 +1002,16 @@ FLATPAKS_TO_MIGRATE = [
         'new-branch': 'stable',
         'old-origin': 'eos-apps',
         'new-origin': 'flathub'
+    },
+    # migrate Skype to flathub (T24741)
+    {
+        'name': 'com.microsoft.Skype',
+        'new-name': 'com.skype.Client',
+        'kind': Flatpak.RefKind.APP,
+        'old-branch': 'eos3',
+        'new-branch': 'stable',
+        'old-origin': 'eos-apps',
+        'new-origin': 'flathub'
     }
 ]
 
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -1216,10 +1216,14 @@ def rewrite_app_id(repo, mtree, old_id, new_id):
     metadata.set_string(FLATPAK_METADATA_GROUP_SESSION_BUS_POLICY,
                         old_id, FLATPAK_METADATA_BUS_POLICY_OWN)
 
-    # prohibit Extra Data apps at present, because they cannot be re-deployed
-    # without connectivity to download the original data file
-    if metadata.has_group(FLATPAK_METADATA_GROUP_EXTRA_DATA):
-        raise Exception('Cannot support extra data app {}'.format(old_id))
+    # extra data apps are handled by removing the [Extra Data] from the metadata
+    # of the migrated app to allow the deploy to be done offline. instead,
+    # after the new app is installed, this script will copy the /app/extra
+    # folder from the pre-migration deploy folder. the migration function
+    # examines the commit metadata of the app being migrated to determine
+    # whether the re-deploy has to carry out this extra step.
+    if metadata.remove_group(FLATPAK_METADATA_GROUP_EXTRA_DATA):
+        logging.info('Removing [Extra Data] from {} metadata'.format(old_id))
 
     metadata_str = mtree_add_keyfile(repo, mtree, metadata, info)
 
@@ -1317,6 +1321,10 @@ def copy_commit(repo, src_rev, dest_ref,
         metadata, vendor_prefixes = rewrite_app_id(repo, mtree, old_app_id, new_app_id)
         commit_metadata.insert_value('xa.metadata',
                                      GLib.Variant('s', metadata))
+
+        # unconditionally remove extra data, the caller determines whether the
+        # the extra data dir needs migrating
+        commit_metadata.remove('xa.extra-data-sources')
     else:
         vendor_prefixes = frozenset()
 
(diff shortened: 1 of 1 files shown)


eos-update-flatpak-repos
```diff
@@ -1476,6 +1503,9 @@ def _migrate_installed_flatpaks():
                                                     arch, new_branch)
                         refs.append(new_ref)
 
+                        if migrate_extra:
+                            migrate_extra_data(ref, new_ref)
```
wjt: What will happen if the system loses power mid-way through this step? I think what will happen is that, on next boot, we'll enter the "Found already installed" branch above, the old copy of Skype will be uninstalled, and the user will be left without a working Skype app until they upgrade it.

I think we need to run this step on both the `new_ref` and `not new_ref` branches, and make `migrate_extra_data` do something like:

```

    if os.path.exists(new_extra):
        # either the user manually installed Skype from Flathub,
        # or we got this far but died before uninstalling the old ref last time
        logging.info('%s already has extra data dir', new_ref)
    else:
        logging.debug('Copying extra data from {} to {}'.format(old_ref, new_ref))
        subprocess.check_call(['cp', '-al', old_extra, new_extra + '.tmp'])
        os.rename(new_extra + '.tmp', new_extra)
```

eos-update-flatpak-repos
```diff
@@ -1332,7 +1343,22 @@ def copy_commit(repo, src_rev, dest_ref,
...
+
+    new_deploy = new_ref.get_deploy_dir()
+    new_extra = os.path.join(new_deploy, 'files', 'extra')
+
+    if os.path.exists(new_extra):
+        raise Exception('not expecting {} to already have extra data dir'
+                        .format(new_ref))
+
+    logging.debug('Copying extra data from {} to {}'.format(old_ref, new_ref))
+    subprocess.check_call(['cp', '-al', old_extra, new_extra])
```
wjt: Of course, since this just creates a small handful of directories and hardlinks, it's very fast so the window of error is very small.

eos-update-flatpak-repos
```diff
@@ -1332,7 +1353,22 @@ def copy_commit(repo, src_rev, dest_ref,
...
+def migrate_extra_data(old_ref, new_ref):
+    old_deploy = old_ref.get_deploy_dir()
+    old_extra = os.path.join(old_deploy, 'files', 'extra')
+
+    new_deploy = new_ref.get_deploy_dir()
+    new_extra = os.path.join(new_deploy, 'files', 'extra')
+
+    if os.path.exists(new_extra):
+        raise Exception('not expecting {} to already have extra data dir'
+                        .format(new_ref))
```
wjt:
```suggestion
                        new_ref.format_ref())
```

eos-update-flatpak-repos
```diff
@@ -1332,7 +1353,22 @@ def copy_commit(repo, src_rev, dest_ref,
...
+
+def migrate_extra_data(old_ref, new_ref):
+    old_deploy = old_ref.get_deploy_dir()
+    old_extra = os.path.join(old_deploy, 'files', 'extra')
+
+    new_deploy = new_ref.get_deploy_dir()
+    new_extra = os.path.join(new_deploy, 'files', 'extra')
+
+    if os.path.exists(new_extra):
+        raise Exception('not expecting {} to already have extra data dir'
```
wjt:
```suggestion
        raise Exception('not expecting %s to already have extra data dir',
```

eos-update-flatpak-repos
```diff
@@ -1332,7 +1353,22 @@ def copy_commit(repo, src_rev, dest_ref,
...
+                        .format(new_ref))
+
+    logging.debug('Copying extra data from {} to {}'.format(old_ref, new_ref))
```
wjt:
```suggestion
    logging.debug('Copying extra data from %s to %s',
                  old_ref.format_ref(), new_ref.format_ref())
```

ramcq:
Added a fixup so you can see the change in approach - unconditionally remove extra-data during the migration, and use a simple commit metadata check before the migration to determine whether the extra dir should be moved (in either clean case, or recovering from an interrupted migration).

eos-update-flatpak-repos
```diff
@@ -1335,6 +1353,30 @@ def copy_commit(repo, src_rev, dest_ref,
...
+    new_extra = os.path.join(new_deploy, 'files', 'extra')
+
+    if os.path.exists(new_extra):
+        # either the user manually installed Skype from Flathub, or we got
+        # this far but died before uninstalling the old ref last time
+        logging.info('%s already has extra data dir', new_ref.format_ref())
+    else:
+        logging.debug('Copying extra data from %s to %s',
+                      old_ref.format_ref(), new_ref.format_ref())
+        subprocess.check_call(['cp', '-al', old_extra, new_extra + '.tmp'])
```
wjt: This behaves differently if the target directory already exists, so will not resume correctly if interrupted:

```
$ find /tmp/tmp.HldWAbt7ur/
/tmp/tmp.HldWAbt7ur/
/tmp/tmp.HldWAbt7ur/source
/tmp/tmp.HldWAbt7ur/source/b
/tmp/tmp.HldWAbt7ur/source/a
$ cp -al /tmp/tmp.HldWAbt7ur/source /tmp/tmp.HldWAbt7ur/target
$ cp -al /tmp/tmp.HldWAbt7ur/source /tmp/tmp.HldWAbt7ur/target
$ find /tmp/tmp.HldWAbt7ur/target
/tmp/tmp.HldWAbt7ur/target
/tmp/tmp.HldWAbt7ur/target/source
/tmp/tmp.HldWAbt7ur/target/source/b
/tmp/tmp.HldWAbt7ur/target/source/a
/tmp/tmp.HldWAbt7ur/target/b
/tmp/tmp.HldWAbt7ur/target/a
```

Adding `-T` seems to do the right thing.

```
       -T, --no-target-directory
              treat DEST as a normal file
```

```suggestion
        subprocess.check_call(['cp', '-alT', old_extra, new_extra + '.tmp'])
```

ramcq:
Squashed, merged.

PR #250, opened 2019-01-15
Factory reset

jprvita:
Add an initial version of a factory reset framework, with one simple stage that removes all user accounts and home directories from the system.

### DIFF
--- a/factory-reset/eos-factory-reset
+++ b/factory-reset/eos-factory-reset
@@ -1,0 +1,112 @@
+#!/usr/bin/python3
+import argparse
+import logging
+import subprocess
+import sys
+
+from systemd import journal
+
+# Options list
+# Each entry has the form (short name, long name, description)
+OPTS = [
+    ("D", "disable", "fully remove a scheduled factory reset"),
+]
+
+# Stages dictionary
+# Each entry has the form {name: description}
+# When creating a new stage make sure to check both STAGES and OPTS so names
+# don't clash, as they both create optional command line arguments.
+STAGES = {
+    "users": "remove all user accounts (and home directories)",
+}
+
+
+def yes_or_no(question):
+    answer = ""
+    while answer not in ["y", "n"]:
+        answer = input(f"{question} ").lower()
+    return answer == "y"
+
+
+def toggle_stage(logger, stage, action):
+    unit = f"eos-factory-reset-{stage}.service"
+    cmd = ["systemctl", "--quiet", action, unit]
+    sp = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    status, msg = sp.returncode, sp.stdout.decode().rstrip("\n")
+    if status != 0:
+        logger.error(f"Failed to {action} stage '{stage}': {msg}")
+    return status, msg
+
+
+if __name__ == "__main__":
+    desc = "reboot the machine and perform a factory reset."
+
+    logger = logging.getLogger(__name__)
+    logger.propagate = False
+    logger.setLevel(logging.INFO)
+    logger.addHandler(logging.StreamHandler())
+    logger.addHandler(journal.JournalHandler())
+
+    parser = argparse.ArgumentParser(description=desc,
+                                     argument_default=argparse.SUPPRESS)
+    for n, name, desc in OPTS:
+        parser.add_argument(f"-{n}", f"--{name}", help=desc, action="store_true")
+
+    group = parser.add_argument_group("stages to be performed on the next boot",
+                                      "if no stage is selected all stages will"
+                                      " be scheduled")
+    for name, desc in STAGES.items():
+        group.add_argument(f"--{name}", help=desc, dest="stages", const=name,
+                           action="append_const")
+
+    args = vars(parser.parse_args())
+    disable = args.pop("disable", False)
+    if disable:
+        stages = STAGES.keys()
+    else:
+        stages = args.pop("stages", STAGES.keys())
+
+    if not disable:
+        action = "enable"
+        warn_msg = [
+            f"Selected stages: {' '.join(stages)}",
+            f"WARNING: You are about to schedule IRREVERSIBLE changes to be",
+            f"performed on the next boot. If you do not want to schedule these",
+            f"changes for the next boot say 'n' now.",
+        ]
+        success_msg = [
+            f"A factory reset procedure is scheduled for the next boot.",
+            f"Please reboot the machine to proceed with the reset process.",
+            f"If this was a mistake, run '{parser.prog} --disable'.",
+        ]
+    else:
+        action = "disable"
+        warn_msg = [
+            "You are about to remove the scheduled factory reset process.",
+            "Unless scheduled again, a factory reset will NOT be performed.",
+        ]
+        success_msg = [
+            "A factory reset procedure is not scheduled for the next boot.",
+        ]
+
+    print(*warn_msg, sep="\n", end="\n\n")
+    if not yes_or_no("Do you want to proceed [y/n]?"):
+        logging.shutdown()
+        sys.exit(0)
+    print()
+
+    formatted_verb = f"{action[0].upper()}{action[1:-1]}ing"
+    logger.info(f"{formatted_verb} stages: {' '.join(stages)}")
+    for stage in stages:
+        status, msg = toggle_stage(logger, stage, action)
+        if status != 0:
+            if not disable:
+                logger.warning(f"Disabling selected stages: {' '.join(stages)}")
+                for s in stages:
+                    toggle_stage(logger, s, "disable")
+            logging.shutdown()
+            sys.exit(1)
+
+    logging.shutdown()
+    print()
+    print(*success_msg, sep="\n")
--- a/factory-reset/eos-factory-reset-users
+++ b/factory-reset/eos-factory-reset-users
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+ADM_GROUP="sudo"
+
+if [ -f /etc/adduser.conf ]; then
+    . /etc/adduser.conf
+fi
+
+first_uid=${FIRST_UID:-1000}
+last_uid=${LAST_UID:-59999}
+
+admin_users=()
+nonadmin_users=()
+
+IFS=':'
+while read -r user _ uid _ _ _ _; do
+    if [[ $uid -ge $first_uid && $uid -le $last_uid ]]; then
+        /usr/bin/id -Gn $user | grep --word-regexp --quiet $ADM_GROUP
+        if [[ $? -eq 0 ]] ; then
+            echo "Found admin user '$user' ($uid)"
+            admin_users+=($user)
+        else
+            echo "Found non-admin user '$user' ($uid)"
+            nonadmin_users+=($user)
+        fi
+    fi
+done </etc/passwd
+
+for user in "${nonadmin_users[@]}"; do
+    /usr/sbin/deluser --remove-home "$user"
+    if [[ $? != 0 ]]; then
+        echo "Failed to remove user '$user'"
+        exit 1
+    fi
+done
+
+for user in "${admin_users[@]}"; do
+    /usr/sbin/deluser --remove-home "$user"
+    if [[ $? != 0 ]]; then
+        echo "Failed to remove user '$user'"
+    fi
+done
--- a/factory-reset/eos-factory-reset-users.service
+++ b/factory-reset/eos-factory-reset-users.service
@@ -6,18 +6,21 @@
 # statement, which is executed despite of the service's main process exit
 # status.
 #
-# This unit is ordered between sysinit.target and basic.target (see the bootup
-# manpage for more info), to guarantee early boot services (specifically mount
-# units) are available but no users are logged in.
+# This unit needs to be ordered before any unprivileged users are able to
+# login, but after the D-Bus daemon has already started. dbus.service is
+# ordered after basic.target, so we can't order this unit before basic.target,
+# otherwise userdel will block waiting for a reply from the D-Bus daemon.
+#
+# Ordering this unit before systemd-user-sessions.service guarantees no
+# unprivileged users will login before it finishes, as a tmpfile snippet
+# creates /run/nologin during early boot and systemd-user-sessions.service
+# removes it when the boot is finished. See pam_nologin(8) for more info.
 
 [Unit]
 Description=Remove All Users And Home Directories
-DefaultDependencies=no
-Conflicts=shutdown.target
-Before=basic.target shutdown.target
-Requires=sysinit.target
-After=sysinit.target
 RefuseManualStart=yes
+Wants=dbus.service
+Before=systemd-user-sessions.service
 
 [Service]
 Type=oneshot
@@ -25,4 +28,4 @@ ExecStart=/usr/libexec/eos-factory-reset/eos-factory-reset-users
 ExecStopPost=/usr/bin/systemctl disable %n
 
 [Install]
-WantedBy=basic.target
+WantedBy=multi-user.target
--- a/factory-reset/eos-factory-reset-users.service
+++ b/factory-reset/eos-factory-reset-users.service
@@ -1,0 +1,28 @@
+# eos-factory-reset-users.service - Remove All Users And Home Directories
+#
+# This unit will be disabled by default, and should be enabled to have it
+# remove all user accounts and home directories on the next boot. It will
+# disable itself automatically after running through its ExecStopPost
+# statement, which is executed despite of the service's main process exit
+# status.
+#
+# This unit is ordered between sysinit.target and basic.target (see the bootup
+# manpage for more info), to guarantee early boot services (specifically mount
+# units) are available but no users are logged in.
+
+[Unit]
+Description=Remove All Users And Home Directories
+DefaultDependencies=no
+Conflicts=shutdown.target
+Before=basic.target shutdown.target
+Requires=sysinit.target
+After=sysinit.target
+RefuseManualStart=yes
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/eos-factory-reset/eos-factory-reset-users
+ExecStopPost=/usr/bin/systemctl disable %n
+
+[Install]
+WantedBy=basic.target
--- a/50-eos.preset
+++ b/50-eos.preset
@@ -12,6 +12,7 @@ disable systemd-networkd-wait-online.service
 disable apt-daily.timer
 disable apt-daily-upgrade.timer
 disable bluetooth-init.service
+disable eos-factory-reset-users.service
 disable eos-firewall-localonly.service
 disable eos-update-server.service
 disable eos-update-server.socket
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,4 +1,4 @@
-SUBDIRS = dracut fallback-fb-setup flatpak-repos nvidia psi-monitor tests
+SUBDIRS = dracut factory-reset fallback-fb-setup flatpak-repos nvidia psi-monitor tests
 EXTRA_DIST = debian .flake8
 
 # Install systemd units, generators, preset files, and udev rules
--- a/configure.ac
+++ b/configure.ac
@@ -61,6 +61,7 @@ AC_CONFIG_FILES([
 	dracut/image-boot/Makefile
 	dracut/repartition/Makefile
 	dracut/customization/Makefile
+	factory-reset/Makefile
 	fallback-fb-setup/Makefile
 	flatpak-repos/Makefile
 	nvidia/Makefile
--- a/factory-reset/Makefile.am
+++ b/factory-reset/Makefile.am
@@ -1,3 +1,7 @@
+dist_sbin_SCRIPTS = \
+	eos-factory-reset \
+	$(NULL)
+
 dist_systemdunit_DATA = \
 	eos-factory-reset-users.service \
 	$(NULL)
--- a/factory-reset/Makefile.am
+++ b/factory-reset/Makefile.am
@@ -1,0 +1,8 @@
+dist_systemdunit_DATA = \
+	eos-factory-reset-users.service \
+	$(NULL)
+
+eosfactoryresetdir = $(libexecdir)/eos-factory-reset
+dist_eosfactoryreset_SCRIPTS = \
+	eos-factory-reset-users \
+	$(NULL)


factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,59 @@
...
+
+    if not quiet:
+        msg = 'Scheduling a factory reset for the next boot with stages: '
+        msg += ' '.join(stages)
+        msg += '\nDo you want to proceed [y/n]?'
+        if not yes_or_no(msg):
+            sys.exit(0)
+
+    for stage in stages:
+        if (stage):
```
dsd: no need for ( )

dsd:
Thanks for working through this!

My main concern on https://github.com/endlessm/eos-boot-helper/pull/247 was how we guarantee that this runs signficantly early in boot, the rest of boot is blocked on this, and that the implementation is something we can rely on for the forseeable future, rather than relying on a somewhat random implementation detail of systemd. I see the boot positioning well explained in commit messages (please also mention it in comments in the unit file too) and I guess you are comfortable in treating the "systemd-user-sessions.service" thing as some kind of relatively stable "API".

I think the stages idea could do with a bit of rework and improvement. That important logic mentioned above will have be duplicated into the service file for all stages that get added - sounds like it would be a little too easy to get out of sync later. The python program to enable/disable them is a bit strange also in the way that it works with command line arguments vs stages - maybe you can use parsed arguments for args like "quiet" but then use the remaining normal args for listing the stages (default to all if none specified?).

@dbnicholson you may also be interested in reviewing this

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,81 @@
...
+import argparse
+import sys
+import subprocess
+
+# Stages list
+# Each stage has the form (short option, name, description)
+# In addition from the stages listed bellow, the following stage names and
+# short option names are already taken by other command line arguments:
+# ('q', 'quiet', 'Suppress all output and don\'t ask for confirmation')
+# ('D', 'disable', 'Disable the selected stages instead of enabling them')
```
dnohales: I'd remove the option definitions from the comment, I guarantee you that once anyone adds a new option or change an existing one it's gonna forgot to update the comment :)

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,81 @@
...
+        help='Suppress all output and don\'t ask for confirmation')
+    parser.add_argument('-D', '--disable', action='store_true',
+        help='Disable the selected stages instead of enabling them')
+
+    group = parser.add_argument_group('factory reset process stages',
+        'select which stages of the factory reset process should be performed '
+        'on the next boot. if no stage is selected all stages will be enabled')
+    for n, name, desc in STAGES:
+        group.add_argument('-'+n, '--'+name, help=desc, action='store_true')
+    args = parser.parse_args()
```
dnohales: You could do `args = vars(parser.parse_args())` here and don't use `vars` below.

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,81 @@
...
+
+    quiet = vars(args).pop('quiet', False)
+    disable = vars(args).pop('disable', False)
+    stages = list(vars(args).keys())
+
+    if len(stages) == 0:
+        for n, name, desc in STAGES:
```
dnohales: You could use an underscore for the unused variables.

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,81 @@
...
+            stages.append(name)
+        if not quiet:
+            print('No stage selected, selecting all stages')
```
dnohales: I think we should change the `print` by a logger and change the level of the logger depending of the `quiet` value (`quiet is True -> level = WARNING`, `quiet is False -> level = DEBUG` )

factory-reset/eos-factory-reset
dnohales: In which case this is falsy?

jprvita: @dnohales Good catch, this was leftover my initial draft where I was iterating over the dictionary returned by argparse, and without suppressing unused argumets.

factory-reset/eos-factory-reset
jprvita: Good point. I'm going to change this to two lists, one with stages and another with the non-stage arguments, than build the parser arguments from both lists. This way we have a single source for this info and a good visibility, to ease maintenance.

factory-reset/eos-factory-reset
jprvita: Yes, that definitely looks better.

factory-reset/eos-factory-reset
jprvita: I did not know about that, thanks for the suggestion!

factory-reset/eos-factory-reset
jprvita: I also don't like how the (quiet / not quiet) logic is intertwined with the program's logic. I modeled the quiet after your original force parameter, but now I'm wondering, do we actually need a quiet or a force parameter for this program?

I'm removing `quiet` for now, and I can add it back if we decide it is necessary. I'm also adding some better logging, including saving the messages to the journal. In the future we may also want to add a metric event reporting that we are about to factory-reset the machine (and later arrange for it to be flushed to the server right away, as part of the future plans are to also reset the machine-id).

dnohales: Yes I was also to ask about the `-y` or `-f` parameter to skip confirmation, but I saw you explicitly used `--quiet` for that in the flag description, I'd prefer the `-y` or `-f` though, `--quiet` sounds more like a verbosity flag instead.

Second paragraph: sounds like a good idea.

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,97 @@
...
+    else:
+        print('You are about to remove the scheduled factory reset process.')
+        print('Unless scheduled again, a factory reset will NOT be performed.\n')
+        action = 'disable'
+    if not yes_or_no('Do you want to proceed [y/n]?'):
+        logging.shutdown()
+        sys.exit(0)
+    print()
+
+    log.info('Requesting to %s stages: %s' % (action, ' '.join(stages)))
```
dnohales: Python's logging functions accepts as `*args` the parameters to be merged to the message string using the string formatting operator, so what you normally do is:

```
log.info('Requesting to %s stages: %s', action, ' '.join(stages))
```

Same for the rest of logging lines.

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,97 @@
...
+import subprocess
+import sys
+
+from systemd import journal
+
+# Options list
+# Each stage has the form (short name, long name, description)
+OPTS = [
+    ('D', 'disable', 'fully remove a scheduled factory reset'),
+]
```
dnohales: Ok, I think I'd add the `-y` or `-f` flags to skip user confirmation, unless the UI for factory reset enables the stages by hand instead of calling this script.

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,97 @@
...
+    args = vars(parser.parse_args())
+    disable = args.pop('disable', False)
+    if disable:
+        stages = list()
+    else:
+        stages = list(args.keys())
+
+    if len(stages) == 0:
+        for _, name, _ in STAGES:
+            stages.append(name)
```
dnohales: What about:

```
stages = [name for _, name, _ in STAGES]
```

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,97 @@
...
+    print()
+
+    log.info('Requesting to %s stages: %s' % (action, ' '.join(stages)))
+    for stage in stages:
+        unit = 'eos-factory-reset-' + stage + '.service'
+        cmd = ['systemctl', '--quiet', action]
+
+        log.info('Trying to %s %s' % (action, unit))
+        cmd.append(unit)
+        systemctl = subprocess.run(cmd, stderr=subprocess.PIPE)
```
dnohales: If you create a function `def toggle_stage(stage, action)` it would simplify a bit of this and remove duplicated code in the disable-on-failure loop.

factory-reset/eos-factory-reset-users.service
```diff
@@ -0,0 +1,31 @@
...
+# This unit will be disabled by default, and should be enabled to have it
+# remove all user accounts and home directories on the next boot. It will
+# disable itself automatically after running through its ExecStopPost
+# statement, which is executed despite of the service's main process exit
+# status.
+#
+# This unit needs to be ordered before any unprivileged users are able to
+# login, but after the D-Bus daemon has already started. dbus.service is
+# ordered after basic.target, so we can't order this unit before basic.target,
+# otherwise userdel will block waiting for a reply from the D-Bus daemon until
```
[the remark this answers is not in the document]
jprvita: ...until I learn to copy and paste properly :-)

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,97 @@
...
+    answer = ''
+    while answer not in ['y', 'n']:
+        answer = input(msg + ' ').lower()
+    return answer == 'y'
+
+
+if __name__ == '__main__':
+    desc = 'reboot the machine and perform a factory reset.'
+
+    log = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
jprvita: Makes sense.

factory-reset/eos-factory-reset
jprvita: Thanks for the heads-up.

factory-reset/eos-factory-reset
jprvita: I'm re-structured the code a bit to make it super simple to add `-f`, and added it in a separate commit on branch `T24613-force`. But I don't like the idea of having a way for the user to schedule a factory reset without confirmation, so I don't think we should merge that branch.

I believe the UI should not call this script directly, but some API from a library (that would be used by this program as well).

factory-reset/eos-factory-reset
jprvita: Thanks for the suggestion! You are obviously more pythonic than I am :-)

factory-reset/eos-factory-reset
jprvita: Agreed.

dsd:
Looking better now.

I'd appreciate review from @wjt and @dbnicholson who have proven to make good calls over this kind of thing. As this is a system that we expect to extend later, it would be nice to get wider review/approval of the design now.

My remaining feedback items:
1. When done, please print a message indicating that the user should reboot for the factory reset to take effect.
2. The way that two sets of arguments are combined to be passed to argparser, only to subtract them again to be dealt with separately by the following code, is a little uncomfortable. Alternatives to consider:
 - `factory-reset --stages="users network"` - treat the stages arg as a space-separated list
 - `factory-reset --quiet users network` - treat the stages list as the remaining args that were not parsed. I think you can do this with `argparse.REMAINDER`.
 - Use argparser to construct the list of stages directly. You can make the `--users` and `--network` args both append to the same list by using the `dest` parameter. See the [append_const example](https://docs.python.org/3/library/argparse.html#action) as something along these lines.
3. Briefly mention somewhere in the commit message or comments why we don't just `systemctl isolate`

And the wider design question for other reviewers:
Right now there is only one stage (users) but in future there will be more. Does the idea of separating this into different stages with different systemd units make sense? The users stage has tricky ordering requirements, but we think that other stages (like one to reset the machine id) might be less strict or have different requirements. There might be a handful of stages in future like resetting network connections, resetting keyboard layout preference, reset machine id, uninstall added apps, etc.

jprvita:
> 1. When done, please print a message indicating that the user should reboot for the factory reset to take effect.

Good reminder.

> 2. The way that two sets of arguments are combined to be passed to argparser, only to subtract them again to be dealt with separately by the following code, is a little uncomfortable. Alternatives to consider:
> 

I think we are really bikeshedding on this at this point, as these arguments will most likely never be used by users. My original thought was that maybe developers or technicians would use them to recover from problems, but even for that I don't have a real scenario of when it would be used. In any case, please find my comments bellow.

> * `factory-reset --stages="users network"` - treat the stages arg as a space-separated list

To have that exact description I'd have to manually parse the string "users network" and validate the sub-strings. I could do `--stage=users --stage=network`, but still it does not allow to add a description of each of the stages to the usage text (`-h`), which I believe is important.

> * `factory-reset --quiet users network` - treat the stages list as the remaining args that were not parsed. I think you can do this with `argparse.REMAINDER`.

This allows users to pass whichever strings they want to the program, which means we would have to manually do the arguments validation that we now get for free. And essentially we would be doing the same thing you seem to be trying to avoid, build the arguments list, remove the non-stage arguments, and user the remainder of the list as the stages list.

> * Use argparser to construct the list of stages directly. You can make the `--users` and `--network` args both append to the same list by using the `dest` parameter. See the [append_const example](https://docs.python.org/3/library/argparse.html#action) as something along these lines.
> 

Yes, that works. I don't see much difference with what I have currently proposed, but I guess it looks a bit better and may be easier to read. I've changed it to use this approach now.

> 3. Briefly mention somewhere in the commit message or comments why we don't just `systemctl isolate`
> 

Honestly, I think it is a bit strange to document approaches we did not take in the commit message or the unit file (which is the only place that define the behavior we implemented in place of isolating a target). We already have this clearing documented on the ticket: https://phabricator.endlessm.com/T24613#668877.

jprvita:
In addition to addressing the comments above, I actually caught a bug in the path that disables all stages if enabling one of them fails, which I also addressed as part of this update (I was calling `toggle_stage` with the failed stage instead of the loop iterator).

factory-reset/eos-factory-reset-users
```diff
@@ -0,0 +1,16 @@
...
+first_uid=${FIRST_UID:-1000}
+last_uid=${LAST_UID:-59999}
+
+IFS=':'
+while read -r user _ uid _ _ _ _; do
+    if [[ $uid -ge $first_uid && $uid -le $last_uid ]]; then
+        echo "Deleting user $user ($uid)"
+        userdel -r "$user"
+    fi
+done </etc/passwd
```
dbnicholson: The `/etc/passwd` that comes from the ostree only has root in it. Is there any reason why you're trying to limit the IDs?

dbnicholson: Alternatively, the actual "factory" ID files are in `/usr/etc/passwd`. You could compare them, remove any entries not in the factory and then just copy them over.

Furthermore, this doesn't handle `/etc/group`, `/etc/shadow` or `/etc/gshadow`. Out of the box root has no password, but someone could have set one. I think what I'd actually do:
```
# Remove non-root users and their files
while IFS=: read -r -a entry; do
    user=${entry[0]}
    if [ -z "$user" ] || [ "$user" = root ]; then
        continue
    fi
    # --remove-all-files might be too slow since it walks the entire system.
    # --remove-home would be faster
    deluser --remove-all-files "$user"
done

# Restore factory ID and password files
cp -a /usr/etc/{passwd,group,shadow,gshadow} /etc

# Wipe any backup ID and password files
rm -f /etc/{passwd,group,shadow,gshadow}-
```
I think if you want to read in both `/etc/passwd` and `/usr/etc/passwd` and compare them to make sure you're not deleting any factory users, then I'd probably get out of shell.

factory-reset/eos-factory-reset-users
```diff
@@ -0,0 +1,16 @@
...
+fi
+
+first_uid=${FIRST_UID:-1000}
+last_uid=${LAST_UID:-59999}
+
+IFS=':'
+while read -r user _ uid _ _ _ _; do
+    if [[ $uid -ge $first_uid && $uid -le $last_uid ]]; then
+        echo "Deleting user $user ($uid)"
+        userdel -r "$user"
```
dbnicholson: I think it would make more sense to use the `deluser` wrapper that debian/ubuntu use since it's also what AccountsService (and any packages) are setup to use. That actually corresponds correctly with sourcing `/etc/adduser.conf` (rather than `/etc/login.defs` that the lower level shadow utilities `useradd` and `userdel` use).

factory-reset/eos-factory-reset-users
jprvita: > The `/etc/passwd` that comes from the ostree only has root in it. Is there any reason why you're trying to limit the IDs?
>
> Alternatively, the actual "factory" ID files are in `/usr/etc/passwd`. You could compare them, remove any entries not in the factory and then just copy them over.
>
> Furthermore, this doesn't handle `/etc/group`, `/etc/shadow` or `/etc/gshadow`. Out of the box root has no password, but someone could have set one. I think what I'd actually do:
> 
> ```
> # Remove non-root users and their files
> while IFS=: read -r -a entry; do
>     user=${entry[0]}
>     if [ -z "$user" ] || [ "$user" = root ]; then
>         continue
>     fi
>     # --remove-all-files might be too slow since it walks the entire system.
>     # --remove-home would be faster
>     deluser --remove-all-files "$user"
> done
> 
> # Restore factory ID and password files
> cp -a /usr/etc/{passwd,group,shadow,gshadow} /etc
> 
> # Wipe any backup ID and password files
> rm -f /etc/{passwd,group,shadow,gshadow}-
> ```
> I think if you want to read in both `/etc/passwd` and `/usr/etc/passwd` and compare them to make sure you're not deleting any factory users, then I'd probably get out of shell.

This is an initial version of the tool, which only addresses the requirements of T24613, which are to remove the user-created accounts so on the next boot the user is taken back to the FBE. This limited scope has been acknowledged by @cosimoc on https://phabricator.endlessm.com/T24613#668534. Its expansion to remove system users and reset other parts of the system like network config, locale, machine-id etc, is covered by T23396 and T10627. But I'll take these points for the more fully-fledged factory reset tool.

`userdel -r` and `deluser --remove-home` both take care of removing the user and groups owned by the user from all of passwd, group, shadow, gshadow.

factory-reset/eos-factory-reset-users
jprvita: It makes sense to match whatever is used by the UI to create / remove user accounts, so I'm changing it `deluser --remove-home`. But both `userdel -r` and `deluser --remove-home` seem to be doing the exact same thing, with `deluser` looking for the homedir and mail spool by itself and then calling userdel (without `-r`).

factory-reset/eos-factory-reset-users
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+ADM_GROUP="sudo"
+
+if [ -f /etc/adduser.conf ]; then
+    . /etc/adduser.conf
+fi
+
+first_uid=${FIRST_UID:-1000}
```
dsd: It would be interesting to know if this value has ever changed over the history of Endless. If it was increased then it would mean some old user accounts would potentially not be deleted by this script.

Same consideration for any future changes to this number, although we could adjust this script at the same time (if we remember this detail).

factory-reset/eos-factory-reset-users
```diff
@@ -0,0 +1,42 @@
...
+last_uid=${LAST_UID:-59999}
+
+admin_users=()
+nonadmin_users=()
+
+IFS=':'
+while read -r user _ uid _ _ _ _; do
+    if [[ $uid -ge $first_uid && $uid -le $last_uid ]]; then
+        /usr/bin/groups $user | grep --quiet $ADM_GROUP
```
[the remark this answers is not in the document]
jprvita: Yes, good point, sorry for letting this slip.

factory-reset/eos-factory-reset-users
jprvita: Unless ostree can provide any sort of history for `/etc/adduser.conf` I don't know of any quick way to do that. For future improvements of the factory reset procedure we should follow @dbnicholson's suggestion and restore `/usr/etc/passwd`, but this would require other considerations like making sure none of the daemons that have system users are running, and perhaps cleaning-up `/var`.

dsd: You can check the entire package history on OBS.

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,113 @@
...
+
+
+def toggle_stage(logger, stage, action):
+    unit = 'eos-factory-reset-' + stage + '.service'
+    cmd = ['systemctl', '--quiet', action]
+    cmd.append(unit)
+    sp = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    status, msg = sp.returncode, sp.stdout.decode().rstrip('\n')
+    if status != 0:
+        logger.error('Failed to %s stage \'%s\': %s', action, stage, msg)
```
wjt:
```suggestion
        logger.error("Failed to %s stage '%s': %s", action, stage, msg)
```

factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,113 @@
...
+        warn_msg = [
+            'Selected stages: ' + ' '.join(stages),
+            'WARNING: You are about to schedule IRREVERSIBLE changes to be',
+            'performed on the next boot. If you do not want to schedule these',
+            'changes for the next boot say \'n\' now.',
+        ]
+        success_msg = [
+            'A factory reset procedure is scheduled for the next boot.',
+            'Please reboot the machine to proceed with the reset process.',
+            'If this was a mistake, run \'' + parser.prog + ' --disable\'.',
```
wjt:
```suggestion
            "If this was a mistake, run '{} --disable'.".format(parser.prog),
```

factory-reset/eos-factory-reset
jprvita: I applied this manually and changed all string quotation marks from single to double, to keep it consistent.

factory-reset/eos-factory-reset
jprvita: I've decided to go with [f-strings](https://realpython.com/python-f-strings/), which made the code much more readable.

factory-reset/eos-factory-reset-users
jprvita: We don't have history before EOS 2.6 or between 3.0 and 3.2 on OBS, so I ended up checking the versions on the packages.txt files for every release. None of these is what I would call quick though, thus my previous question.

Anyway, `adduser.conf` was the same between EOS 2.0 and EOS 3.4, when `LAST_UID` and `LAST_GID` where increased from 29999 to 59999, but `FIRST_UID` was kept the same, so we should be fine. I could not find anything about EOS < 2.0.

dsd:
I'm ready to merge this however there are conflicts.

wjt:
@jprvita 

> I've decided to go with f-strings, which made the code much more readable.

Since we want this on eos3.5, we cannot use f-strings. They were added in Python 3.6, but eos3.5 has Python 3.5. This is a merge blocker.

jprvita:
> I'm ready to merge this however there are conflicts.

I've rebased on master and solved the conflicts. It was only build-system stuff which I was expecting due to the fallback-fb-setup work.

> @jprvita
> 
> > I've decided to go with f-strings, which made the code much more readable.
> 
> Since we want this on eos3.5, we cannot use f-strings. They were added in Python 3.6, but eos3.5 has Python 3.5. This is a merge blocker.

Yeah, I forgot to check on eos3.5. I've created https://github.com/endlessm/eos-boot-helper/pull/252 for eos3.5, which we need anyway because of the conflicts mentioned above, and changed the f-strings to str.format() there only. If you guys agree with that I would rather keep the f-strings on master, to keep using them for future expansions of `eos-factory-reset`. Otherwise it is just a matter of `git cherry-pick T24613-eos3.5` on top of this branch.

PR #251, opened 2019-01-23
 Clear the fallback framebuffer during boot

jprvita:
This adds a new boot service, fallback-fb-setup, which clears the
fallback framebuffer displayed by DRM drivers when there is no client
framebuffer installed. In practice, this framebuffer is displayed when
the user session ends and GDM or Plymouth did not installed their
framebuffer yet. Since some drivers (namely i915 atm) keep the contents
from the firmware framebuffer as their fallback framebuffer, to provide
smoother graphics during boot, these contents end-up being displayed on
logout / shutdown unless we clear them.

This is a side-effect of the effort to have a flicker-free boot.

https://phabricator.endlessm.com/T24682

### DIFF
--- a/fallback-fb-setup/fallback-fb-setup.c
+++ b/fallback-fb-setup/fallback-fb-setup.c
@@ -10,14 +10,34 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <linux/fb.h>
+#include <linux/kd.h>
 
 #define FBDEV "/dev/fb0"
+#define TTYDEV "/dev/tty0"
 
 int main(int argc, char **argv) {
-    int ret = 0, fb_fd = -1;
+    int ret = 0, fb_fd = -1, tty_fd = -1, tty_mode;
     uint8_t *fbp = MAP_FAILED;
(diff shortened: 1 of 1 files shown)


fallback-fb-setup/fallback-fb-setup.c
```diff
@@ -0,0 +1,56 @@
...
+    screensize = vinfo.xres * vinfo.yres * bytes_per_pixel;
+
+    fbp = mmap(NULL, screensize, PROT_WRITE, MAP_SHARED, fb_fd, (off_t) 0);
+    if (fbp == MAP_FAILED) {
+        perror("Failed to mmap " FBDEV);
+        ret = errno;
+        goto exit;
+    }
+
+    for (p = fbp; p < fbp + screensize; p++)
```
dsd: How about memset()?

jprvita: Yes, good point.

jprvita: I've now updated the PR to use `memset`.

PR #262, opened 2019-03-13
eos-preset: Disable wpa_supplicant-wired@.service

jprvita:
We don't want to enable any unit from wpa_supplicant as we don't
configure interfaces individually outside of Network Manager. Network
Manager starts wpa_supplicant via D-Bus when needed.

This also times-out during boot, adding a 90s delay bacause it requires
sys-subsystem-net-devices-multi-user.device, which does not exist.

https://phabricator.endlessm.com/T25608

### DIFF
--- a/50-eos.preset
+++ b/50-eos.preset
@@ -29,8 +29,9 @@ disable speech-dispatcherd.service
 disable ssh*.service
 disable ssh.socket
 disable systemd-nspawn@.service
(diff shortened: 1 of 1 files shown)


wjt:
Thanks for doing the archaeology to find out [why this wasn't a problem before](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=871488)!

PR #265, opened 2019-04-11
eos-preset: disable crio

wjt:
This is pulled in as a dependency of podman. crio(8) describes it as:

> OCI-based implementation of Kubernetes Container Runtime Interface Daemon

It fails to start at boot:

    crio[465]: time="2019-03-28 23:35:29.396993833Z" level=fatal msg="no default routes found in "/proc/net/route" or "/proc/net/ipv6_route""
    systemd[1]: crio.service: Main process exited, code=exited, status=1/FAILURE
    systemd[1]: crio.service: Failed with result 'exit-code'.

It succeeds if you start it later, so this is likely a dependency issue.
However, since we don't explicitly want it, let's just disable it.

https://phabricator.endlessm.com/T25542

### DIFF
--- a/50-eos.preset
+++ b/50-eos.preset
@@ -12,6 +12,7 @@ disable systemd-networkd-wait-online.service
 disable apt-daily.timer
 disable apt-daily-upgrade.timer
 disable bluetooth-init.service
+disable crio.service
 disable eos-factory-reset-users.service
 disable eos-firewall-localonly.service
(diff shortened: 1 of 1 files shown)


ramcq:
Thanks for the speedy review but... we might not need cri-o-1.12 in the ostree at all, it could just be a bogus dependency of the podman deb. Trying to figure it out with the upstream folks.

PR #270, opened 2019-05-08
migrate Minecraft to flathub

ramcq:
https://phabricator.endlessm.com/T26588

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -1269,7 +1269,7 @@ COMMIT_BODY_INDEX = 4
 
 def copy_commit(repo, src_rev, dest_ref,
                 old_app_id=None, new_app_id=None,
-                collection_id=None):
+                collection_id=None, migrate_extra=False):
     """Copy commit src_rev to dest_ref
     This makes the new commit at dest_ref have the proper collection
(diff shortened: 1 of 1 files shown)


eos-update-flatpak-repos
```diff
@@ -1269,7 +1269,7 @@ COMMIT_BODY_INDEX = 4
 
 def copy_commit(repo, src_rev, dest_ref,
                 old_app_id=None, new_app_id=None,
-                collection_id=None):
+                collection_id=None, migrate_extra=False):
```
wjt: Please sir, could I have a test? You should be able to just parametrize the ID(s) in `test_rename_extra_data`? I know it doesn't actually test the extra data is copied…

ramcq: Re-using that test would not actually test the code I changed, because it, after set-up, calls eufr.rename_app_id and verifies that it correctly strips the extra data in the process of renaming it. All I am changing in this patch is adding a new scenario in which rename_app_id is called, so this change is entirely outside the existing test coverage - testing that this flag is set correctly in a non-rename case would require scaffolding to simulate an app in a repo with a deploy with an extra data subdir... :cold_sweat:

PR #273, opened 2019-05-22
 eos-update-flatpak-repos: avoid un-necessary metadata changes

ramcq:
In the case we are migrating an extra data app without renaming it, we have to
rewrite the metadata to strip the extra data section, but these other changes
are redundant.

https://phabricator.endlessm.com/T26588


ramcq:
Not sure if this is worthwhile, I made the fixup before I realised you had already merged. #272

wjt:
Looks fine but I think it should have a better commit message if we want this – we'll never rebase so the autosquash notation is not meaningful

ramcq:
Reworded commit message.

PR #275, opened 2019-06-07
eos-update-flatpak-repos: Migrate endlessnetwork apps

mazen-asef:
https://phabricator.endlessm.com/T25928

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -1031,6 +1031,25 @@ FLATPAKS_TO_MIGRATE = [
         'new-branch': 'stable',
         'old-origin': 'eos-apps',
         'new-origin': 'flathub'
+    },
+    # migrate endlessnetwork apps to flathub (T25928)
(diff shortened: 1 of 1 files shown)


eos-update-flatpak-repos
```diff
@@ -1031,6 +1031,39 @@ FLATPAKS_TO_MIGRATE = [
         'new-branch': 'stable',
         'old-origin': 'eos-apps',
         'new-origin': 'flathub'
+    },
+    # migrate endlessnetwork apps to flathub (T25928)
+    {
+        'name': 'co.com.btg.Conductor',
```
wjt: https://github.com/flathub/co.com.btg.Conductor doesn't exist. These all need to specify the new name.

mazen-asef:
Sorry, updated the correct apps to be migrated - it seems only missilemath and midnightmareteddy were deployed to prod, so I'm only bothering with those.

PR #278, opened 2019-07-30
image-boot: move /etc overlayfs setup into real root

dsd:
Our image boot solution involves creating overlayfses at every
point in the filesystem hierarchy where things might be written to.
We set up all of these overlayfses from eos-live-boot-overlayfs-setup
which runs during early boot from the real root, however due to
complications faced with /etc/machine-id during the initial implementation,
the /etc overlayfs is created much earlier, from the initramfs.

The /etc overlayfs being created separately inconveniences upcoming
work to support persistent live storage, so we want to move it
to the later point in eos-live-boot-overlayfs-setup instead.

The /etc/machine-id complication is now averted by our ostree
including an empty /etc/machine-id file. On read-only media boots,
systemd's behaviour to create a random /run/machine-id and bind-mount
it over the top of /etc/machine-id is now active.

Creating the /etc overlayfs a bit later though, as is done here, will
lose the visibility of that bind mount. So eos-live-boot-overlayfs-setup
now also takes care to make it included and visible in the overlayfs.

https://phabricator.endlessm.com/T27158


eos-live-boot-overlayfs-setup
```diff
@@ -47,18 +44,41 @@ setup_ostree_flatpak_overlay() {
...
     mount -t overlay -o \
         rw,upperdir=/run/eos-live/$dir,lowerdir=/$dir,workdir=/run/eos-live/$dir-workdir \
         eos-live-$dir /$dir
+}
+
+setup_overlay etc
+
+# If we booted and systemd was unable to find or create a machine-id,
+# it will create one at /run/machine-id and bind mount it at /etc/machine-id.
+# But now that we just setup an overlay at /etc, we lost that bind mount, so
```
wjt: Huh, I did not expect that mounting an overlay over `/etc` with lowerdir `/etc` would lose the bind-mount within `/etc`.

eos-live-boot-overlayfs-setup
```diff
@@ -47,18 +44,41 @@ setup_ostree_flatpak_overlay() {
...
+#
+# In addition to recovering the machine-id, we also want to use this
+# opportunity to store it in the overlayfs, in case the overlayfs is backed
+# by persistent storage.
+#
+# Both goals can be satisfied in simplistic fashion by copying systemd's
+# boot-generated machine-id into place, in the case where we didn't find one
+# in the overlay.
+if [[ ! -s /etc/machine-id && -s /run/machine-id ]]; then
+  cp -v /run/machine-id /etc/machine-id
```
wjt: In this implementation there is a window of time where `/etc/machine-id` becomes empty once again. It's probably not a big problem but we could imagine doing this the other way round:

* Before setting up the overlay, if `/etc/machine-id` is a symlink to `/run/machine-id`, then copy `/run/machine-id` to `/run/eos-live/etc/machine-id`
* Now set up the overlay

Worthwhile? I'm not sure.

dsd: > Before setting up the overlay, if /etc/machine-id is a symlink to /run/machine-id

It always will be as we're booting live media here.

There are other ways of detecting and avoiding the gap though, and I just pushed the simplest approach I can think of. Please take a look

wjt: Good point.

I'd never seen `mount --move` before – nice solution.

PR #279, opened 2019-08-02
record-boot-success: Create com.endlessm.RecordBootSuccess

jprvita:
Creates a D-Bus activatable service that when started will tell the
bootloader that the boot was successfull. Also install desktop files so
this service is started when GDM's login window is reached, when
gnome-initial-setup is reached, or automatic login is used.

https://phabricator.endlessm.com/T26791


configure.ac
```diff
@@ -54,6 +54,31 @@ AC_ARG_WITH([modprobedir],
                             [Path to the system modprobe.d directory])],
             [modprobedir="$withval"], [modprobedir="/lib/modprobe.d"])
 AC_SUBST(modprobedir)
+AC_ARG_WITH([dbusdir],
+            [AS_HELP_STRING([--with-dbusdir],
+                            [Path to the system dbus-1 directory])],
+            [dbusdir="$withval"], [dbusdir="/usr/share/dbus-1"])
+AC_SUBST(dbusdir)
```
wjt: Where is this used?

configure.ac
```diff
@@ -54,6 +54,31 @@ AC_ARG_WITH([modprobedir],
...
+AC_ARG_WITH([dbussystemconfigdir],
+            [AS_HELP_STRING([--with-dbussystemconfigdir],
+                            [Path to the system dbus-1/system.d directory])],
+            [dbussystemconfigdir="$withval"], [dbussystemconfigdir="/usr/share/dbus-1/system.d"])
+AC_SUBST(dbussystemconfigdir)
+AC_ARG_WITH([dbussystemservicesdir],
+            [AS_HELP_STRING([--with-dbussystemservicesdir],
+                            [Path to the system dbus-1/system-services directory])],
+            [dbussystemservicesdir="$withval"], [dbussystemservicesdir="/usr/share/dbus-1/system-services"])
```
wjt: `pkg-config --variable=system_bus_services_dir dbus-1` or use `@datadir@/dbus-1/system-services`?

configure.ac
```diff
@@ -54,6 +54,31 @@ AC_ARG_WITH([modprobedir],
...
+AC_SUBST(dbussystemservicesdir)
+AC_ARG_WITH([xdgautostartdir],
+            [AS_HELP_STRING([--with-xdgautostartdir],
+                            [Path to the system dbus-1 directory])],
+            [xdgautostartdir="$withval"], [xdgautostartdir="/etc/xdg/autostart"])
```
wjt: Do you need to parametrize this? Could you just use `$sysconfdir/xdg/autostart` in the relevant place?

configure.ac
```diff
@@ -54,6 +54,31 @@ AC_ARG_WITH([modprobedir],
...
+            [dbussystemconfigdir="$withval"], [dbussystemconfigdir="/usr/share/dbus-1/system.d"])
+AC_SUBST(dbussystemconfigdir)
+AC_ARG_WITH([dbussystemservicesdir],
+            [AS_HELP_STRING([--with-dbussystemservicesdir],
+                            [Path to the system dbus-1/system-services directory])],
+            [dbussystemservicesdir="$withval"], [dbussystemservicesdir="/usr/share/dbus-1/system-services"])
+AC_SUBST(dbussystemservicesdir)
+AC_ARG_WITH([xdgautostartdir],
+            [AS_HELP_STRING([--with-xdgautostartdir],
+                            [Path to the system dbus-1 directory])],
```
wjt:
```suggestion
                            [Path to the xdg autostart directory])],
```

configure.ac
```diff
@@ -54,6 +54,31 @@ AC_ARG_WITH([modprobedir],
...
+            [xdgautostartdir="$withval"], [xdgautostartdir="/etc/xdg/autostart"])
+AC_SUBST(xdgautostartdir)
+AC_ARG_WITH([gdmgreeterautostartdir],
+            [AS_HELP_STRING([--with-gdmgreeterautostartdir],
+                            [Path to the system dbus-1 directory])],
```
wjt:
```suggestion
                            [Path to the gdm greeter's autostart directory])],
```

record-boot-success/com.endlessm.RecordBootSuccess.service
```diff
@@ -0,0 +1,5 @@
+[D-BUS Service]
+Name=com.endlessm.RecordBootSuccess
+User=root
+SystemdService=record-boot-success.service
+Exec=/usr/lib/eos-boot-helper/record-boot-success/record-boot-success
```
wjt: Are you sure that this is what `$libexecdir` (which is where this gets installed) expands to? This file should be templated, I think.

record-boot-success/record-boot-success
```diff
@@ -0,0 +1,41 @@
...
+
+BUSNAME = "com.endlessm.RecordBootSuccess"
+
+
+def record_boot_success():
+    envfile = "/boot/grub/grubenv"
+    program = "/usr/bin/grub-editenv"
+
+    try:
+        if os.path.getsize(program) == 0:
```
wjt: What's the purpose of this? In what scenario would this file exist but be empty? Surely running it a few lines below would fail in that case?

wjt: The shell script this is replacing is:

```
 [ -s /boot/grub/grubenv ] || rm -f /boot/grub/grubenv 
```

Note that it's checking for a zero-size `grubenv`, not a zero-size executable.

record-boot-success/record-boot-success.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Record Boot Success
+
+[Service]
+Type=dbus
+BusName=com.endlessm.RecordBootSuccess
+ExecStart=/usr/lib/eos-boot-helper/record-boot-success/record-boot-success
```
wjt: `$libexecdir` here again.

configure.ac
jprvita: It was used in Makefile.am, but I think it makes sense to just drop it.

configure.ac
jprvita: Makes sense, but we'll have to also add a build-dependency on libdbus1-dev.

configure.ac
jprvita: I don't really need to parametrize any of these, I was just following what was being done for the other variables there.

configure.ac
jprvita: I'll just drop this parameter.

configure.ac
jprvita: I'll just drop this parameter.

record-boot-success/com.endlessm.RecordBootSuccess.service
jprvita: Yes, I'm sure about the variable expansion, as I used it before for the factory reset scripts and fallback-fb-setup.

And when I wrote the first one of them I thought about having the variable in the service file and substituting at build time, but failed miserably to implement it. IIRC the variable was expanding to a different value in the service file that what it was in Makefile.am, but this was quite some time ago so I may be getting confused.

In any case, this package does not need to be super portable.

record-boot-success/record-boot-success
jprvita: Yep, that was a typo. And I don't have the answer for why `/boot/grub/grubenv` would be empty, but this comes from the original Debian init script that replaced on https://github.com/endlessm/grub/commit/e3a6fc3f37818f6bb169100017588679147ceacb.

Manually making `/boot/grub/grubenv` an empty file makes grub-editenv fail, and if we boot in that state GRUB prints an error message and boots fine, but it does not set the "recordfail" mark, so it is probably a good idea to keep checking for that.

record-boot-success/record-boot-success.service
jprvita: Same as for `com.endlessm.RecordBootSuccess.service`.

jprvita:
> Somewhere, you'll need to arrange for this script not to run on systems without GRUB.

You are right, previously this was part of the grub-common package, so it would never run on systems where grub was not installed. I believe checking `ConditionFileIsExecutable=/usr/bin/grub-editenv` in record-boot-success.service should be enough. AFAIU the D-Bus manager will trigger the systemd unit when it is present and not try to start the service on its own, so it should be enough.

record-boot-success/com.endlessm.RecordBootSuccess.service
wjt: It's not portability I'm worried about. I don't think it is a theoretical possibility that this path could become wrong.

Search https://manpages.debian.org/testing/debhelper/debhelper.7.en.html for "v12". Debian's policy about libexecdir has relatively recently changed – if we ever change `debian/compat` to `12` or greater for any other reason, `$(libexecdir)` in the `Makefile` will change to `/usr/libexec` and this will stop working, but the build will succeed. We'll only notice if we happen to read the journal.

record-boot-success/record-boot-success
```diff
@@ -0,0 +1,41 @@
...
+
+
+def record_boot_success():
+    envfile = "/boot/grub/grubenv"
+    program = "/usr/bin/grub-editenv"
+
+    try:
+        if os.path.getsize(envfile) == 0:
+            os.remove(envfile)
+    except OSError:
```
wjt:
```suggestion
    except FileNotFoundError:
```

record-boot-success/record-boot-success.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Record Boot Success
+ConditionFileIsExecutable=/usr/bin/grub-editenv
```
wjt: Have you tested what happens if this condition fails? Do we get errors in the journal from `record-boot-success.desktop`? I guess `StartServiceByName` will fail and so the `dbus-send` incantation will fail. Does this have adverse consequences?

record-boot-success/record-boot-success
```diff
@@ -0,0 +1,41 @@
...
+        pass
+
+    cmd = [program, envfile, "unset", "recordfail"]
```
wjt: What about new PAYG systems that boot with sd-boot? I believe these will use the regular x86_64 ostree, so `grub-editenv` will be executable. But I don't think `/boot/grub` will exist. I bet this will fail.

PR #283, opened 2019-09-04
endless-repartition: Update loader efi var so it works on real computers

ManMower:
Overwriting the LoaderDevicePart variable doesn't work on all EFI
firmware implementations. Some (including tianocore, where this
behaviour can be verified by inspection in Variable.c) don't allow
volatile variables to be altered in any way at run time.

It works on VirtualBox, where I've been testing this code, but not
on any of the physical hardware I've tried on.

Instead of writing downstream hacks into systemd to make this work,
bind mount our chosen variable contents over the file we can't change.
This leaves a harmless bind mount lying around on the first boot -
this can be seen in 'df' output.

https://phabricator.endlessm.com/T27041


dsd:
As an alternative is it possible to delete and then recreate the variable?

ManMower:
No - I tried that first.   You can't append, delete, create, or overwrite any "volatile" variables after boot services.

PR #284, opened 2019-09-09
 image-boot: Create partition for persistent data storage 

dsd:
In order to add persistent data storage capabilities to our
live+installer setup (so that you don't lose your data on reboot),
add another step in the boot process that creates the partition.

Similar to the repartitioning case for first-boot of installed systems,
this must be done in the initramfs, before any filesystems are mounted,
so that the kernel is able to update it's view of the partition table
without requiring a reboot.

Even though there are some similarities to endless-repartition.sh, the
actual details are quite different so I found it more appropriate to
implement as a separate script.

The actual filesystem creation is deferred to the boot from the "real
root" and will be handled separately.

https://phabricator.endlessm.com/T27158

### DIFF
--- a/dracut/image-boot/eos-live-storage-setup
+++ b/dracut/image-boot/eos-live-storage-setup
@@ -1,0 +1,98 @@
+#!/bin/sh
+# Copyright (C) 2019 Endless Mobile, Inc.
+# Licensed under the GPLv2
+#
+# When performing a live boot, create a 3rd partition filling the remaining
+# space on the device. This partition will later be used for persistent data
+# storage, so that your documents are kept for next time you boot.
+#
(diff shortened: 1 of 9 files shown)


dracut/image-boot/eos-live-storage-setup
```diff
@@ -0,0 +1,98 @@
+#!/bin/sh
+# Copyright (C) 2019 Endless Mobile, Inc.
+# Licensed under the GPLv2
+#
+# When performing a live boot, create a 3rd partition filling the remaining
```
wjt:
```suggestion
# When performing a live boot from an ISO image on a writeable medium, create a 3rd partition filling the remaining
```

PR #289, opened 2019-10-09
hack: Add new service to remove the global override

danigm:
Old hack computer has a global flatpak override file. In the new hack
experience we're going from using a custom product EOS to use the main
EOS branch so, we need to remove this global override and we'll use
local override for each application.

This patch adds a service that will remove the global override, only for
old hack-computer users, it checks for the HackComponents flatpak that
should only be installed in old hack computers, and if that exists it'll
remove the global override.

This service is based on the eos-update-flatpak-repos to be run only on
update.

https://phabricator.endlessm.com/T27984

## Here we've the steps to follow to test this:

 1. Place the systemd service in your system:

```
cp hack-remove-global-override.service /etc/systemd/system
```

2. You should need to have `com.endlessm.HackComponents`
3. Create the global override file if it doesn't exists in `/var/lib/flatpak/overrides/global`
4. Update your system and reboot! Then, after the reboot, the global override should be removed.

You can try to uninstall the `com.endlessm.HackComponents` and try again, in that case the global override won't be removed.

If it's hard to have a new system update you can modify the service to be run without the update replacing this lines:

```
# Only run on updates
Before=multi-user.target systemd-update-done.service
ConditionNeedsUpdate=|/etc
ConditionNeedsUpdate=|/var
```

and leaving only:

```
Before=multi-user.target
```

### DIFF
--- a/hack-remove-global-override.service
+++ b/hack-remove-global-override.service
@@ -1,0 +1,21 @@
+[Unit]
+Description=Remove hack flatpak global override
+DefaultDependencies=no
+Conflicts=shutdown.target
+Wants=local-fs.target
+After=local-fs.target
+
+# Only run on updates
+Before=multi-user.target systemd-update-done.service eos-updater-flatpak-installer.service
+ConditionNeedsUpdate=|/etc
+ConditionNeedsUpdate=|/var
+
+ConditionPathExists=/var/lib/flatpak/overrides/global
+ConditionPathExists=/var/lib/flatpak/app/com.endlessm.HackComponents
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/rm /var/lib/flatpak/overrides/global
+
+[Install]
+WantedBy=multi-user.target
--- a/Makefile.am
+++ b/Makefile.am
@@ -23,6 +23,7 @@ dist_systemdunit_DATA = \
 	eos-live-boot-overlayfs-setup.service \
 	eos-reclaim-swap.service \
 	eos-update-flatpak-repos.service \
+	hack-remove-global-override.service \
 	$(NULL)
 
 dist_systemdgenerator_SCRIPTS = \


hack-remove-global-override.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
+Description=Remove hack flatpak global override
+DefaultDependencies=no
+Conflicts=shutdown.target
+Wants=local-fs.target
+After=local-fs.target
+
+# Only run on updates
+Before=multi-user.target systemd-update-done.service
```
cfoch: :+1:

hack-remove-global-override.service
```diff
@@ -0,0 +1,21 @@
...
+ConditionNeedsUpdate=|/etc
+ConditionNeedsUpdate=|/var
+
+ConditionPathExists=/var/lib/flatpak/overrides/global
```
wjt: Looking at [the hook](https://github.com/endlessm/eos-image-builder/blob/eos3.6/hooks/image/51-hack-flatpak-overrides.chroot), on Hack systems this file will be a symbolic link into the HackComponents app. What happens if, on a current hack system, you run `pkexec flatpak override --talk-name=com.example.DummyName`? I can think of 3 possibilities:

1. The operation fails because the symlink points to a file which is read-only
2. The operation succeeds, and replaces the symlink with a file
3. The operation succeeds, and modifies the file inside the HackComponents app, corrupting the Flatpak repo

I really hope it's 1 or 2!

If 1, what we have here is probably fine. If 2 or (gulp) 3, we might consider whether we want to handle this case differently and just remove the overrides that come from HackComponents, in case users have used global overrides? Or is that considered to be vanishingly unlikely?

systemd does support:

```ini
ConditionPathIsSymbolicLink=/var/lib/flatpak/overrides/global
```

but I'm not sure if that helps.

danigm: I've just tested and the final result is the number 2, the symlink is replaced with a file with the same content and the new override.

I think we've two options here:

1. Remove the file only if it's a symbolic link, as you said, and in that case, if the users has modified the global override, the new clubhouse won't work correctly, because that override will replace the one in apps and we expect that new apps will have the `com.hack_computer.Clippy.Extension` not the old one.

2. The other option is to keep as is right now and it'll be removed so the user will loose their custom changes.

I think that the best option is to remove it and if the user has changed that (it should be a rare user case, it's not usual to override global flatpaks), he will loose the changes and should do again, but the new clubhouse and the old apps will continue working with the local overrides created by the clubhouse during the migration quest.

wjt: I think you're right that it will be quite rare that a Hack user will have applied a system-wide global Flatpak override.

hack-remove-global-override.service
```diff
@@ -0,0 +1,21 @@
...
+ConditionPathExists=/var/lib/flatpak/app/com.endlessm.HackComponents
```
wjt: Are we going to keep `com.endlessm.HackComponents` installed on Hack 1.0 systems forever?

danigm: The `com.endlessm.HackComponents` is needed for old hack toy apps, so I think we'll keep that until we replace all the hack toy apps or when we drop.

wjt: OK – I think we need `Before=eos-updater-flatpak-installer.service` on this unit, assuming that would be the mechanism used to remove that Flatpak.

Otherwise – suppose we drop com.endlessm.HackComponents in 3.8.0. Then it's racy whether systemd checks this job's conditions or runs `eos-updater-flatpak-installer.service` which would (most likely) be the thing that uninstalls that Flatpak.

danigm: I've just added that service to the list of Before in the Unit.

PR #293, opened 2019-11-18
eos-update-flatpak-repos: skip parental controls checks

wjt:
Installing a Flatpak now involves talking to accountsservice, even if
running as root. Without this change, the script job fails with:

    ERROR:root:Failure applying migration to org.squeakland.Etoys
      File "/usr/sbin/eos-update-flatpak-repos", line 1589, in _migrate_installed_flatpaks
        arch, new_branch)
    gi.repository.GLib.GError: g-io-error-quark: Could not connect: No such file or directory (1)

https://phabricator.endlessm.com/T28386

### DIFF
--- a/eos-update-flatpak-repos.service
+++ b/eos-update-flatpak-repos.service
@@ -28,5 +28,10 @@ Type=oneshot
 RemainAfterExit=true
 ExecStart=/usr/sbin/eos-update-flatpak-repos
 
+# Flatpak checks parental controls at deploy time. In order to do this, it
+# needs to talk to accountsservice on the system bus, neither of which are
+# running when this job runs.
+Environment=FLATPAK_SKIP_PARENTAL_CONTROLS_NO_SYSTEM_BUS=1
+
 [Install]
 WantedBy=multi-user.target


eos-update-flatpak-repos.service
```diff
@@ -28,5 +28,10 @@ Type=oneshot
 RemainAfterExit=true
 ExecStart=/usr/sbin/eos-update-flatpak-repos
 
+# Flatpak checks parental controls at deploy time. In order to do this, it
+# needs to talk to accountsservice on the system bus, neither of which are
+# running when this job runs.
+Environment=FLATPAK_SKIP_PARENTAL_CONTROLS_NO_SYSTEM_BUS=1
```
pwithnall: Note: This environment variable was supposed to be a temporary workaround to unblock flatpak development. And yet, predictably, it’s now showing up all over the place to fix things. I don’t have a better expedient suggestion though.

wjt: Yeah, I know…

I think we should revisit this and see if there is a way to implement the policy "uid 0 ⇒ no parental controls checks". But as you say, I don't think there is a better expedient solution.

PR #296, opened 2020-03-12
eos-update-flatpak-repos: Fix remote URL snafu

dbnicholson:
During the 3.7.6 release I made a poorly thought through decision to
override the image builder URLs so that they pulled directly from our
server instead of our CDN that kept failing. I thought that the URLs
would only be used while pulling the data and not end up as the final
URLs. That was not the case.

Add a function to walk through all the OSTree and Flatpak remotes and
swap our internal ostree.endlessm-sf.com domain for the CDN
ostree.endlessm.com domains.

I didn't test this at all yet but wanted to get it out for review.

https://phabricator.endlessm.com/T29535

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -29,6 +29,7 @@ import logging
 import os
 import subprocess
 import sys
+from urllib.parse import urlparse, urlunparse
 
 from systemd import journal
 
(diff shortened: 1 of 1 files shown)


eos-update-flatpak-repos
```diff
@@ -153,6 +154,46 @@ def _remove_runtimes():
                                '--runtime', refspec])
 
 
+# From 3.7.6 through 3.7.8, images were shipped with remote URLs using
+# our internal ostree.endlessm-sf.com domain instead of our
+# ostree.endlessm.com CDN domain.
```
wjt: Perhaps a reference to the ticket would be good?

dbnicholson: Added

dbnicholson:
OK, now I actually tested this and found several bugs. Primarily I was not handling ostree's "boolean return with out values" API well, which is nice in C but not so much in python. And I forgot that you need to call `Flatpak.Installation.modify_remote()` after changing a `FlatpakRemote` to get it to actually do anything.

eos-update-flatpak-repos
```diff
@@ -153,6 +154,55 @@ def _remove_runtimes():
...
+
+
+def _fix_remote_urls():
+    _, repo = OSTree.Sysroot().get_repo()
+    repo.open()
+    for remote in repo.remote_list():
+        # Only adjust non-flatpak remotes (xa.disable = true)
+        _, is_os_repo = repo.get_remote_boolean_option(remote, 'xa.disable',
+                                                       False)
+        if not is_os_repo:
```
wjt: The OS repo is always called `eos`, right, so perhaps we could simplify this?

wjt: What we have here is probably fine though.

dbnicholson: True. I was thinking it could potentially be renamed like a git repo, but ostree doesn't support that. This is now generic, though. I can go either way.

PR #297, opened 2020-03-12
Allow persistent storage for combo usb sticks

ManMower:
Let endless use a large file created by the installer on the exfat partition for persistent storage.

### DIFF
--- a/dracut/image-boot/eos-map-image-file
+++ b/dracut/image-boot/eos-map-image-file
@@ -26,7 +26,7 @@ if [ "$#" != "3" ]; then
   exit 1
 fi
 
-if ! fstype=$(lsblk --noheadings -o FSTYPE "${host_device}"); then
+if ! fstype=$(lsblk --noheadings --nodeps -o FSTYPE "${host_device}"); then
   echo "image-boot: failed to detect filesystem type for ${host_device}" >&2
   exit 1
 fi
--- a/eos-live-boot-overlayfs-setup
+++ b/eos-live-boot-overlayfs-setup
@@ -8,7 +8,7 @@ EXTRA_PATHS=$@
 # user.
 mount -o remount,user_xattr /run
 
-# Attempt to find the partition to act as backing storage for the overlayfses
+# Attempt to find a target to act as backing storage for the overlayfses
 find_storage_partition() {
 	local bootdev=$(grep -oP "endless\.image\.device=UUID=[^ ]+" /proc/cmdline)
 	bootdev=${bootdev:21}
@@ -16,6 +16,17 @@ find_storage_partition() {
 
 	local root_partition=$(blkid --output device --match-token "${bootdev}")
 	[ -b "${root_partition}" ] || return 1
+
+
+	dumpexfat -f /endless/persistent.img ${root_partition} &> /dev/null
+	if [[ $? = 0 ]]; then
+		udevadm settle
+		/usr/libexec/eos-map-image-file ${root_partition} /endless/persistent.img endless-live_storage
+		if [[ $? = 0 ]]; then
+			echo /dev/mapper/endless-live_storage
+			return 0
+		fi
+	fi
 
 	# Check for ISO layout - root partition is partition 1, storage partition
 	# exists at partition 3.
--- a/Makefile.am
+++ b/Makefile.am
@@ -86,6 +86,10 @@ dist_sbin_SCRIPTS = \
 	eos-update-flatpak-repos \
 	$(NULL)
 
+dist_libexec_SCRIPTS = \
+	dracut/image-boot/eos-map-image-file \
+	$(NULL)
+
 tmpfilesdir = $(prefix)/lib/tmpfiles.d
 dist_tmpfiles_DATA = \
 	tmpfiles.d/avahi-service-writers.conf \
--- a/eos-image-boot-dm-setup.service
+++ b/eos-image-boot-dm-setup.service
@@ -4,6 +4,7 @@ DefaultDependencies=no
 After=ostree-remount.service
 Before=local-fs.target
 ConditionKernelCommandLine=endless.image.device
+ConditionKernelCommandLine=!endless.live_boot
 
 [Service]
 Type=oneshot


ManMower:
If someone could give tips on how to test this on iso images written with 'dd' under linux, I'd appreciate it.  The process of unsquashing/resquashing the squashfs, putting it back in the iso file, then recreating an appropriate partition table (since the first partition is now too large to fit before the ESP) eludes me.  Is there an easier way to do this?

The second patch is perhaps a bit sloppy, as it installs a file in 2 places. There are many ways to do this and I'm not sure what's best (put it in libexec, update both users.  put it in sbin, update both users.  leave it in dracut dirs, run it directly from there in the setup script...)  I went with this way because it was easiest, and there's still no duplication of the file in the repo. I'll happily change it to whatever is preferred.

ManMower:
I should probably also mention that I have local changes to the rufus winstaller that I've used to test this code.  I'll make a PR of that after I get back to fixing up the rufus rebase, but a preview is at https://github.com/endlessm/rufus/commit/0430dad58cdf236c452191eb2e335c7c9d5efe44

dracut/image-boot/eos-map-image-file
```diff
@@ -26,7 +26,11 @@ if [ "$#" != "3" ]; then
   exit 1
 fi
 
-if ! fstype=$(lsblk --noheadings -o FSTYPE "${host_device}"); then
+# lsblk can output many lines, we only want the first. However, we use
```
wjt: Why does it output many lines? would `lsblk --nodeps` do the right thing, without needing to murge its output like this?

eos-live-boot-overlayfs-setup
```diff
@@ -17,6 +17,14 @@ find_storage_partition() {
 	local root_partition=$(blkid --output device --match-token "${bootdev}")
 	[ -b "${root_partition}" ] || return 1
 
+	udevadm settle
+
+	eos-map-image-file ${root_partition} /endless/live_storage endless-live_storage
```
wjt: It's a bikeshed but since I guess this is a disk image file, could we call it something like `/endless/persistent.img`?

wjt: I think you are relying on this failing if the file doesn't exist but this will cause spurious error messages in the boot log for ISO images. Perhaps a quick `[[ -f $PATH ]]` could prevent that?

eos-live-boot-overlayfs-setup
```diff
@@ -17,6 +17,14 @@ find_storage_partition() {
...
+	if [ $? = 0 ]; then
```
wjt: Rather than doing this you could use:

```
if [[ -f "$PERSISTENT_IMG_PATH" ]] && eos-map-image-file …; then
```

wjt:
> If someone could give tips on how to test this on iso images written with 'dd' under linux, I'd appreciate it. The process of unsquashing/resquashing the squashfs, putting it back in the iso file, then recreating an appropriate partition table (since the first partition is now too large to fit before the ESP) eludes me. Is there an easier way to do this?

Does the modified & rebuilt initramfs fit in the ESP in the ISO? If so, I suggest:

- Loop-mount the ESP
- Copy the initramfs into the ESP
- Unmount it
- Boot the ISO, holding Shift (on BIOS) or mashing Esc (on EFI) to get to the GRUB menu
- Manually edit the GRUB boot options to use your modified initramfs rather than the one out of the squashfs

> The second patch is perhaps a bit sloppy, as it installs a file in 2 places. There are many ways to do this and I'm not sure what's best (put it in libexec, update both users. put it in sbin, update both users. leave it in dracut dirs, run it directly from there in the setup script...) I went with this way because it was easiest, and there's still no duplication of the file in the repo. I'll happily change it to whatever is preferred.

I think putting it in both libexec and the dracut dir is fine. On an OSTree system, files with identical contents are hardlinks to one another.

dracut/image-boot/eos-map-image-file
ManMower: Ah, that looks like it'll do nicely.

eos-live-boot-overlayfs-setup
ManMower: I'd rather make the udevadm settle conditional on the file existence too anyway, since I've seen
other commits where we go out of our way to avoid it (e4642cce73)

eos-live-boot-overlayfs-setup
ManMower: I hadn't considered the log message - you are correct about my intent.

I don't think I can do the presence test quite that simply, because I don't think we mount that filesystem at all - but I could do dumpexfat and pipe the results to /dev/null to check...

I've forced pushed changes that do this.

ManMower:
I've force pushed changes intended to resolve concerns - and I've added a new patch that prevents image-boot-dm-setup from running on live_boot systems.

I've still not comprehensively tested these changes - they're not all in the dracut image so I can't just build a new image and stuff that in the ESP. I need to change the iso contents.

ManMower:
There is bug.  do not land.

ManMower:
This has now been successfully tested on a dd'd iso image by:
mounting the image, extracting the squashfs contents, loopback mounting the third partition of that, replacing eos-live-boot-overlayfs-setup and adding eos-map-image-file to libexec
mksquashfs, then running xorriso in the same way the image builder does.

This testing revealed a noob bug that has now been fixed.

eos-live-boot-overlayfs-setup
```diff
@@ -17,6 +17,17 @@ find_storage_partition() {
 	local root_partition=$(blkid --output device --match-token "${bootdev}")
 	[ -b "${root_partition}" ] || return 1
 
+
+	dumpexfat -f /endless/persistent.img ${root_partition} > /dev/null 2>&1
+	if [[ $? = 0 ]]; then
+		udevadm settle
```
wjt: Is this needed because `dumpexfat` accesses the partition device directly?

eos-live-boot-overlayfs-setup
```diff
@@ -17,6 +17,17 @@ find_storage_partition() {
 	local root_partition=$(blkid --output device --match-token "${bootdev}")
 	[ -b "${root_partition}" ] || return 1
 
+
+	dumpexfat -f /endless/persistent.img ${root_partition} > /dev/null 2>&1
```
wjt:
```suggestion
	dumpexfat -f /endless/persistent.img ${root_partition} &> /dev/null
```

eos-live-boot-overlayfs-setup
ManMower: No, it's because eos-map-image-file uses lsblk

eos-live-boot-overlayfs-setup
ManMower: I'll make that change, thanks!

PR #298, opened 2020-03-25
eos-live-boot-overlayfs-setup: Fix eos-map-image-file invocation

ManMower:
I made the false assumption that dist_libexec_SCRIPTS installed
into /usr/libexec, but I didn't create packages to adequately
confirm installation location.  The actual location on an eos
system is /usr/lib/eos-boot-helper

Fix the script to look in the right place.

https://phabricator.endlessm.com/T14891

### DIFF
--- a/eos-live-boot-overlayfs-setup
+++ b/eos-live-boot-overlayfs-setup
@@ -21,7 +21,7 @@ find_storage_partition() {
 	dumpexfat -f /endless/persistent.img ${root_partition} &> /dev/null
 	if [[ $? = 0 ]]; then
 		udevadm settle
-		/usr/libexec/eos-map-image-file ${root_partition} /endless/persistent.img endless-live_storage
+		/usr/lib/eos-boot-helper/eos-map-image-file ${root_partition} /endless/persistent.img endless-live_storage
 		if [[ $? = 0 ]]; then
 			echo /dev/mapper/endless-live_storage
 			return 0


ManMower:
Multiple ways to fix this, wasn't sure which to pursue.
There's also making the script into a .in file and fixing it up at 'build' time
or moving eos-map-image-file into /usr/libexec

maybe there are more that I hadn't thought of.

jprvita:
There are other scripts and unit files that hardcode `/usr/lib/eos-boot-helper` and use `$(libexecdir)` in their Makefile, so at least the current PR consistently expands on that problem. However, we should find some time to get all of them fixed and use the value of libexecdir everywhere (so making all invocations into `.in` files). I tried that approach when implementing one of factory-reset, fallback-fb-setup or record-boot-success and found some inconsistencies in how that variable was expanded, so I ended up with this ugly workaround. But IIRC from researching at the time, Debian may be switching libexecdir from `/usr/lib/<pkg>` to `/usr/libexec/<pkg>` soon, and this will obviously break when that happens.

I guess this was a long way to say I'm fine with this for the current release, but let's pay this debt on master after we branch.

wjt:
As described on https://phabricator.endlessm.com/T27643 Debian has already switched, but packages opt in by bumping `debian/compat`.

jprvita:
> As described on https://phabricator.endlessm.com/T27643 Debian has already switched, but packages opt in by bumping `debian/compat`.

I forgot I had already filled that, thanks for having a better memory than mine :smile:

PR #299, opened 2020-05-06
eos-update-flatpak-repos: Migrate Firefox to Flathub

wjt:
As well as changing (a single bit in) the app ID, this migration
involves a previously-unseen complication.

Firefox stores users' profile data in ~/.mozilla. In our Flatpak, we
simply granted the Flatpak access to the user's entire home directory –
presumably because it predates Firefox learning how to use Flatpak
portals for filesystem access.

However, Mozilla's Flatpak takes a different route: it's built with
--persist=.mozilla. This causes ~/.var/app/$APP_ID/.mozilla to be
bind-mounted to ~/.mozilla inside the sandbox. The real ~/.mozilla is
inaccessible. This means that, without some intervention during the
migration, users would lose their Firefox configuration across the
migration.

There are, essentially, two options:

1. Use `flatpak override --system --filesystem=~/.mozilla
   org.mozilla.firefox` to grant the Mozilla-provided Firefox Flatpak
   access to the real ~/.mozilla. This appears to take precedence over
   `--persist=.mozilla`, but it's not clear that Flatpak guarantees
   this. In addition, this keeps us further away from the upstream
   configuration.

2. Migrate the data to its new location with a script which runs when a
   user logs in (under certain conditions).

We choose option 2. However, this introduces a further complication:
after booting into the new build of Endless OS, but before updating any
apps, the installed 'org.mozilla.firefox' Flatpak is actually *our*
Flatpak rebranded by the migration script, with full home-directory
access. So again we have two options:

1. Make ~/.mozilla a symbolic link to the config's new home.

2. Adjust the installed Flatpak's permissions to be the same as
   Mozilla's.

Once again we choose option 2.

The migration goes as follows:

a. At boot, system-wide: construct and install a org.mozilla.firefox
   Flatpak based on the installed org.mozilla.Firefox Flatpak, with its
   permissions adjusted to more closely match Mozilla's.

b. At login, per-user: if ~/.mozilla exists,
   ~/.var/app/org.mozilla.Firefox exists (indicating that the user has
   launched our Firefox Flatpak at least once) but
   ~/.var/app/org.mozilla.Firefox/.mozilla does not exist, move
   ~/.mozilla to ~/.var/app/org.mozilla.Firefox/.mozilla.

There is yet one more hurdle we have to clear. The Firefox configuration
defines multiple profile directories, and there is not just one default:
there is one default per Firefox installation path. The path to the
Firefox executable in our Flatpak is different to that in Mozilla's.
Work around this by patching the configuration to add a section for
Mozilla's installation path.

https://phabricator.endlessm.com/T29746


wjt:
Does not work because `profiles.ini` needs to be modified because the default profile is keyed by a hash of Firefox's install path, which is different between the two flatpaks.

wjt:
I have tested this as follows:

1. Install org.mozilla.Firefox from eos-apps dev channel
2. Launch it, set my homepage to nyan.cat
3. `eos-dev-unlock --hotfix`
4. Install a `.deb` built from this branch
5. Reboot
6. Log in
7. Launch Firefox → the homepage is still correctly displayed in the migrated copy of our Firefox Flatpak, version 75.
8. Launch App Center → Firefox shows as updatable from Flathub
9. Update it → it updates successfully
10. Launch Firefox → Nyan Cat appears in Firefox 77.

This branch contains some detours into the build system for the tests, with the result that more tests are run in CI now. There's one left that isn't run here but I gave up and filed https://phabricator.endlessm.com/T30327.

eos-migrate-firefox-profile.service
```diff
@@ -0,0 +1,14 @@
...
+Description=Migrate Firefox profile to new path
+
+ConditionPathIsDirectory=%h/.mozilla
+ConditionPathIsDirectory=%h/.var/app/org.mozilla.Firefox
+ConditionPathIsDirectory=!%h/.var/app/org.mozilla.Firefox/.mozilla
+
+[Service]
+Type=simple
+ExecStart=/usr/lib/eos-boot-helper/eos-migrate-firefox-profile
+Restart=no
```
dbnicholson, line 11: I think for a script where you want the unit to not be considered started until the script completes, you would want `Type=oneshot`. To persist it, add `RemainAfterExit=yes`. See [systemd.service(5)](https://man7.org/linux/man-pages/man5/systemd.service.5.html). With `Type=Simple`, systemd will consider the unit started as soon as it finishes forking the script. That means `gnome-session.target` could be considered started before this script completes.

eos-update-flatpak-repos
```diff
@@ -276,6 +277,49 @@ def write_deploy_file(deploy_file_path, orig_variant, idx, new_val):
...
 
 
+def mangle_firefox_metadata(metadata):
+    """Endless's Firefox Flatpak stored its data in the real ~/.mozilla in the user's
+    home directory. Mozilla's Firefox Flatpak stores its data in what it thinks is
+    ~/.mozilla, but which is actually ~/.var/app/$APP_ID/.mozilla through the magic of
+    --persist=.mozilla. Monkeypatch our Flatpak to match.
+
+    eos-migrate-firefox-profile is run via /etc/xdg/autostart to actually move the
+    data."""
```
dbnicholson, line 287: Is that true? I thought it was handled by the systemd user session.

wjt: Great point. I was going to use xdg autostart but then decided to use systemd instead. In hindsight I'm not sure this was a good choice because of all the extra build system plumbing needed, but being able to use `systemctl status --user` was nice when debugging.

dbnicholson: I'm super pro-systemd and all the robust control over the execution environment, so for me it's clearly better than an autostart desktop file.

PR #300, opened 2020-05-07
Filter "sector-size" to sfdisk

valentindavid:
Some version (v2.35) of sfdisk outputs header "sector-size" but does
not accept it back. So repartition fails if not filtered.

We encountered that in GNOME OS which is using the repartition script from this repository.


wjt:
ok to test

PR #303, opened 2020-06-15
firefox: Update MIME type associations

wjt:
Strictly speaking we should have done this for all the other
applications that we have migrated. However, the browser is more
noticeable, since there is always at least one other, higher-priority
browser installed.

https://phabricator.endlessm.com/T29746

### DIFF
--- a/eos-migrate-firefox-profile
+++ b/eos-migrate-firefox-profile
@@ -39,6 +39,61 @@ PROFILES_INI = os.path.join(HOME_DOT_MOZILLA, "firefox", "profiles.ini")
 ENDLESS_FLATPAK_INSTALL_GROUP = "InstallFF6C2EBF42BF07E5"
 MOZILLA_FLATPAK_INSTALL_GROUP = "InstallCF146F38BCAB2D21"
 
+MIMEAPPS_LIST = os.path.join(GLib.get_user_config_dir(), "mimeapps.list")
+MIMEAPPS_GROUPS = [
+    "Default Applications",
+    "Added Associations",
+    "Removed Associations",
+]
+OLD_DESKTOP_FILE = "org.mozilla.Firefox.desktop"
+NEW_DESKTOP_FILE = "org.mozilla.firefox.desktop"
+
+
+def update_mimeapps_list(path):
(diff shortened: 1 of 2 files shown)


eos-migrate-firefox-profile
```diff
@@ -39,6 +39,61 @@ PROFILES_INI = os.path.join(HOME_DOT_MOZILLA, "firefox", "profiles.ini")
...
 MOZILLA_FLATPAK_INSTALL_GROUP = "InstallCF146F38BCAB2D21"
 
+MIMEAPPS_LIST = os.path.join(GLib.get_user_config_dir(), "mimeapps.list")
+MIMEAPPS_GROUPS = [
+    "Default Applications",
+    "Added Associations",
+    "Removed Associations",
+]
+OLD_DESKTOP_FILE = "org.mozilla.Firefox.desktop"
+NEW_DESKTOP_FILE = "org.mozilla.firefox.desktop"
```
dbnicholson, line 49: This is fine for a single desktop file to migrate, but might want a dict or a list of tuples to make it easy to add to in the future.

tests/test_migrate_firefox_profile.py
```diff
@@ -12,6 +12,74 @@
...
+        ).lstrip()
+        expected_data = orig_data
+
+        self._test(orig_data, expected_data)
+
+    def test_there(self):
+        orig_data = textwrap.dedent(
+            """
+            [Default Applications]
+            text/html=org.mozilla.Firefox.desktop
```
dbnicholson, line 42: I'm sure GKeyFile will do the right thing, but might be good to make the test data have multiple entries in the list to make sure the correct entry is replaced and the order is maintained.

eos-migrate-firefox-profile
wjt: In principle it would have been good to do this for the other migrated apps whose IDs changed, but at this point almost everything from eos-apps has been retired, which is why I just scoped this to this Firefox-specific script.

tests/test_migrate_firefox_profile.py
wjt: Good point. I added an `[Added Applications]` section which is where, in practice, there are plural entries.

PR #304, opened 2020-07-25
Remove sysrq_enable_oom()

hakavlad:
Note that the value of `/proc/sys/kernel/sysrq` influences only the invocation via a keyboard. Invocation of any operation via `/proc/sysrq-trigger` is always allowed (by a user with admin privileges). [1] 

Changing the value in `/proc/sys/kernel/sysrq` is not necessary. So, we can safely remove `sysrq_enable_oom()`.

1. https://www.kernel.org/doc/html/latest/admin-guide/sysrq.html#how-do-i-enable-the-magic-sysrq-key

### DIFF
--- a/psi-monitor/psi-monitor.c
+++ b/psi-monitor/psi-monitor.c
@@ -15,10 +15,8 @@
 #define RECOVERY_INTERVAL  15
 #define MEM_THRESHOLD      10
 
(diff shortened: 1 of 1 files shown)


jprvita:
ok to test

jprvita (approved):
This makes sense, thanks for the contribution and for posting the documentation reference.

PR #305, opened 2020-08-01
live: Make kolibri data from eoslive available in EOS

jprvita:
Ship a mount unit to always mount /dev/mapper/endless-image-device on live USB systems. This is a device-mapper device which maps the 'eoslive' partition minus the image file that the system boots from.

If there is Kolibri content available in it, make it available for the Kolibri app in EOS.

This also incidentally fixes a bug on live+installer combo USBs when launching the installer from the FBE, backing off, then trying to launch it from the user session.

https://phabricator.endlessm.com/T22735

### DIFF
--- a/eoslive-kolibrihome-setup.service
+++ b/eoslive-kolibrihome-setup.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Setup Kolibri to use data from Endless Live USB Storage
+After=systemd-tmpfiles-setup.service
+ConditionKernelCommandLine=endless.image.device
+ConditionKernelCommandLine=endless.live_boot
+ConditionDirectoryNotEmpty=/run/kolibrihome
+
+[Service]
+Type=oneshot
+ExecStart=/bin/bash -c 'echo -e "[Environment]\nKOLIBRI_HOME=/run/kolibrihome" > /run/flatpak/overrides/org.learningequality.Kolibri'
+
+[Install]
+WantedBy=multi-user.target
--- a/run-kiwixhome.mount
+++ b/run-kiwixhome.mount
@@ -1,0 +1,13 @@
+[Unit]
+Description=Kiwix data from Endless Live USB Storage
+ConditionKernelCommandLine=endless.image.device
+ConditionKernelCommandLine=endless.live_boot
+ConditionDirectoryNotEmpty=/run/media/eoslive/.kiwix
+
+[Mount]
+What=/run/media/eoslive/.kiwix
+Where=/run/kiwixhome
+Options=bind
+
+[Install]
+WantedBy=multi-user.target
--- a/run-kolibrihome.mount
+++ b/run-kolibrihome.mount
@@ -1,0 +1,13 @@
+[Unit]
+Description=Kolibri data from Endless Live USB Storage
+ConditionKernelCommandLine=endless.image.device
+ConditionKernelCommandLine=endless.live_boot
+ConditionDirectoryNotEmpty=/run/media/eoslive/.kolibri
+
+[Mount]
+What=/run/media/eoslive/.kolibri
+Where=/run/kolibrihome
+Options=bind
+
+[Install]
+WantedBy=multi-user.target
--- a/run-media-eoslive.mount
+++ b/run-media-eoslive.mount
@@ -1,0 +1,11 @@
+[Unit]
+Description=Endless Live USB Storage
+ConditionKernelCommandLine=endless.image.device
+ConditionKernelCommandLine=endless.live_boot
+
+[Mount]
+What=/dev/mapper/endless-image-device
+Where=/run/media/eoslive
+
+[Install]
+WantedBy=local-fs.target
--- a/tmpfiles.d/flatpak-extension-kiwix-usb-content.conf
+++ b/tmpfiles.d/flatpak-extension-kiwix-usb-content.conf
@@ -1,0 +1,3 @@
+D /run/flatpak/extension/com.endlessm.encyclopedia.en.Content/x64_64 0755 - - -
+L /run/flatpak/extension/com.endlessm.encyclopedia.en.Content/x64_64/eos3 - - - - /run/kiwixhome/flatpak
+L /var/lib/flatpak/extension - - - - /run/flatpak/extension
--- a/Makefile.am
+++ b/Makefile.am
@@ -24,7 +24,9 @@ dist_systemdunit_DATA = \
 	eos-live-boot-overlayfs-setup.service \
 	eos-reclaim-swap.service \
 	eos-update-flatpak-repos.service \
+	eoslive-kolibrihome-setup.service \
 	hack-remove-global-override.service \
+	run-kolibrihome.mount \
 	run-media-eoslive.mount \
 	$(NULL)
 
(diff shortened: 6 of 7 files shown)


run-kolibrihome.mount
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Kolibri data from Endless Live USB Storage
+ConditionKernelCommandLine=endless.image.device
+ConditionKernelCommandLine=endless.live_boot
+ConditionDirectoryNotEmpty=/run/media/eoslive/.kolibri
+
+[Mount]
+What=/run/media/eoslive/.kolibri
+Where=/run/kolibrihome
+Options=bind
```
dylanmccall, line 10: We need this to be a readable / writable directory, at least for desktop users running the Kolibri app. (This is a bit gross for a multi-user system so we have a much better system usually, but in this case it's a bit of a weird hack). My initial thought is to add `DirectoryMode=777`. Do you think that would be okay here?

jprvita: In my tests, `DirectoryMode` ends up not really having any effect since the `What` of this unit is backed by an exFAT filesystem, which does not have access control information. But yeah, in theory we would need that if the backing fs was any linux filesystem.

dylanmccall: Ah, got it. Thanks :)

eoslive-kolibrihome-setup.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Setup Kolibri to use data from Endless Live USB Storage
+After=systemd-tmpfiles-setup.service
+ConditionKernelCommandLine=endless.image.device
+ConditionKernelCommandLine=endless.live_boot
+ConditionDirectoryNotEmpty=/run/kolibrihome
+
+[Service]
+Type=oneshot
+ExecStart=/bin/bash -c 'echo "KOLIBRI_HOME=/run/kolibrihome" > /run/flatpak/overrides/org.learningequality.Kolibri'
```
wjt, line 10: Are we sure this is the correct format for flatpak overrides files? I would expect it to be:

```ini
[Environment]
KOLIBRI_HOME=/run/kolibrihome
```

tmpfiles.d/flatpak-extension-kiwix-usb-content.conf
```diff
@@ -0,0 +1,3 @@
+D /run/flatpak/extension/com.endlessm.encyclopedia.en.Content/x64_64 0755 - - -
+L /run/flatpak/extension/com.endlessm.encyclopedia.en.Content/x64_64/eos3 - - - - /run/kiwixhome/flatpak
+L /var/lib/flatpak/extension - - - - /run/flatpak/extension
```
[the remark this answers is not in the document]
jprvita: No, it doesn't. On current images it gets created by a different tmpfiles snippet (xserver-xorg-video-nvidia.conf), but that may be dropped at some point in the future. systemd-tmpfiles is smart enough to deal with duplicate instances of the same directive, and this way we don't depend on external files.

tmpfiles.d/flatpak-extension-kiwix-usb-content.conf
```diff
@@ -0,0 +1,3 @@
+D /run/flatpak/extension/com.endlessm.encyclopedia.en.Content/x64_64 0755 - - -
+L /run/flatpak/extension/com.endlessm.encyclopedia.en.Content/x64_64/eos3 - - - - /run/kiwixhome/flatpak
```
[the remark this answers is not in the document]
jprvita: IDK, the way I have thought of this is making it a runtime extension, not overriding the runtime. Maybe it would be better to make it a runtime as you suggest, let's discuss on the next meeting.

run-kiwixhome.mount
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Kiwix data from Endless Live USB Storage
+ConditionKernelCommandLine=endless.image.device
+ConditionKernelCommandLine=endless.live_boot
+ConditionDirectoryNotEmpty=/run/media/eoslive/content/kiwix
```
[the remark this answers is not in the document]
jprvita: sounds good to me, I'll update this for the next revision.

eoslive-kolibrihome-setup.service
wjt: Or invoke `flatpak override`?

tmpfiles.d/flatpak-extension-kiwix-usb-content.conf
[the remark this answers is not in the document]
dylanmccall: Ah, that path is for unmanaged extensions! Can't find real docs, but https://blog.tingping.se/2018/08/26/flatpak-host-extensions.html :)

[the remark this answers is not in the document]
wjt: We also discussed that this means the files on the exFAT partition must be in the layout expected by the SDK. The layout appears to be:

```
/$SUBSCRIPTION_NAME
  /manifest.json
  /1.zim
```

eoslive-kolibrihome-setup.service
jprvita: I think you are right here, I'll fix it for the next revision.

wjt: Unfortunately `flatpak override` does not have a way to write to `/run`.

Perhaps:

```ini
[Service]
Type=oneshot
ExecStart=cat
StandardInput=data
StandardInputText=[Environment]
StandardInputText=KOLIBRI_HOME=/run/kolibrihome
StandardOutput=file:/run/flatpak/overrides/org.learningequality.Kolibri
```

wjt: Or a file on disk and a symlink?

jprvita: `echo -e "[Environment]\nKOLIBRI_HOME=/run/kolibrihome"`?

tmpfiles.d/flatpak-extension-kiwix-usb-content.conf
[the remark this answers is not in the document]
jprvita: Per the discussion on today's meeting, I'm keeping the paths as-is here.

jprvita:
Addressed previous comments, ready for another round of review.

run-media-eoslive.mount
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Endless Live USB Storage
+ConditionKernelCommandLine=endless.image.device
+ConditionKernelCommandLine=endless.live_boot
+
+[Mount]
+What=/dev/mapper/endless-image-device
+Where=/run/media/eoslive
```
wjt, line 8: As I lay awake in the middle of the night, I had a thought: what will the permissions on this directory be? We need it to be writable by all users on the system:

- `eos-installer` saves eos-diagnostics to it on error, and may run either as `gnome-initial-setup` or as a "proper" user
- We are using (a subdirectory of) it as `$KOLIBRI_HOME`, and we don't (in general) know the UID of the user(s) who will run Kolibri

jprvita: I don't recall the mount options from the top of my head, but in my tests it was world-writable (and the fs being exFAT, we only need to worry about the mountpoint permissions, which are defined at mount time).

jprvita:
Merging after clarifying a few points and getting the final OK during the meeting today.

PR #306, opened 2020-08-20
Move eoslive mount units to a generator

jprvita:
Mount units do not work as one might expect with respect to conditions: https://github.com/systemd/systemd/issues/9875. Let's move them over to a generator.


eos-live-boot-generator
```diff
@@ -11,14 +11,68 @@ is_live()
...
+After=systemd-tmpfiles-setup.service
+ConditionKernelCommandLine=endless.image.device
+ConditionDirectoryNotEmpty=/run/kolibrihome
+
+[Service]
+Type=oneshot
+ExecStart=/bin/bash -c 'echo -e "[Environment]\nKOLIBRI_HOME=/run/kolibrihome" > /run/flatpak/overrides/org.learningequality.Kolibri'
+
+[Install]
+WantedBy=multi-user.target
```
wjt, line 63: Is just writing the service to `$dest_dir` sufficient to cause it to be activated? Does `systemctl preset` or `preset-all` (or the moral equivalent) get run after the generators run? Naïvely I would expect to need to put it into `multi-user.target.wants/` but I am always surprised by these things…

wjt (approved):
It seems a bit sad that these files are now embedded in a shell script rather than being files on disk, but there we go!

I will press ahead assuming that I am wrong in my question about whether the `eoslive-kolibrihome-setup.service` needs to be explicitly enabled.

eos-live-boot-generator
jprvita: Yep, you are right -- we can't call `systemctl` from generators since d-bus is not around yet, but we should be manually enabling these by creating syslinks in the right place, i.e., `/run/systemd/generator/multi-user.target.wants`.

I'm fixing this atm.

PR #307, opened 2020-08-20
live: enable generated mount units

jprvita:
We also need to enable generated units from the generator.

In addition, I've found that the bind-mount units were breaking the installer, so I have dropped them.

https://phabricator.endlessm.com/T22735


dylanmccall:
Looks :100: to me as well :)

PR #310, opened 2020-08-27
Only add Encyclopedia content symlink if appropriate

wjt:
Previously, the tmpfiles.d snippet that makes a directory within
/run/media/eoslive appear as the com.endlessm.encyclopedia.en.Content
extension was run unconditionally. This means that on all systems,
/var/lib/flatpak/extension/com.endlessm.encyclopedia.en.Content/x86_64/eos3
was a symlink to /run/media/eoslive/.kiwix/flatpak.

On non-live systems, or on live systems which don't have this unusual
configuration, this causes the encyclopedia to fail to start:

    wjt@camille:~/src/endlessm/eos-boot-helper$ com.endlessm.encyclopedia.en
    bwrap: Can't find source path /var/lib/flatpak/extension/com.endlessm.encyclopedia.en.Content/x86_64/eos3: No such file or directory
    error: ldconfig failed, exit status 256

Move the tmpfiles.d snippet to a different directory, and only invoke it
if it is reasonable to do so: this is a live system, and the target
directory exists.

While looking into this I also noticed an error in the existing unit, and fixed it. It doesn't cause a runtime warning but `systemd-analyze validate` on the generated unit does warn.

(We could consider changing these units to be on disk but not `[Install]`ed so they could more easily be statically checked, and edited outside a shell script?)


wjt:
It is rather hard to test this because in the live environment, you can't install packages, because `eos-dev-unlock` doesn't work, because it gets confused by the `persistent.img` bind mounts.

So I tested this as follows:

- On my non-live system:
    1. Built the `.deb`
    2. Unlocked my system and installed it
    3. Verified that `/usr/share/eos-boot-helper/tmpfiles.d/flatpak-extension-kiwix-usb-content.conf` is present rather than its old path. At this point I stopped: I am satisfied that nothing will spontaneously run it.
- In the live system with this weird layout:
    1. Manually set up an overlayfs on `/usr` to allow me to modify stuff there
    2. Removed the magic symlink
    3. Updated `/usr/lib/systemd/system-generators/eos-live-boot-generator`
    4. Ran `systemctl daemon-reload` – the new unit was generated
    5. Ran `systemctl start eoslive-encyclopedia-setup.service` – the symlink was corrected created and and the encyclopedia worked

eos-live-boot-generator
```diff
@@ -58,3 +58,20 @@ Type=oneshot
 ExecStart=/bin/bash -c 'echo -e "[Context]\nfilesystems=/run/media/eoslive/.kolibri;\n\n[Environment]\nKOLIBRI_HOME=/run/media/eoslive/.kolibri" > /run/flatpak/overrides/org.learningequality.Kolibri'
 SETUP_KOLIBRIHOME
 ln -sf "$unit_path" "$multi_user_target_wants/"
+
+unit_path="$dest_dir/eoslive-encyclopedia-setup.service"
```
dnohales, line 62: Maybe to be consistent you should call this `flatpak-extension-kiwix-usb-content-setup.service`

wjt: Perhaps – I picked it to be consistent with the unit above, and to emphasise the fact that it's specific to live systems.

PR #315, opened 2020-10-09
image-boot-setup: Workaround busybox issue by manually specifying fs

ManMower:
We recently started using the dracut busybox module, and this has
replaced util-linux mount with busybox's implementation of mount.

For some reason busybox isn't figuring out what fs type our .iso
files are when the isofs kernel module isn't already loaded.

Specify it manually when mounting to work around this problem.

https://phabricator.endlessm.com/T30937

### DIFF
--- a/dracut/image-boot/eos-image-boot-setup
+++ b/dracut/image-boot/eos-image-boot-setup
@@ -101,7 +101,13 @@ exfat | ntfs)
 iso9660)
   # Mount the ISO image
   mkdir /cd
-  if ! mount ${host_device} /cd; then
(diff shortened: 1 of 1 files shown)


andrunko (approved):
Change itself lgtm but I have never looked at this code before so take that with a grain of salt :) -  feel free to merge if confident it will work (it can't get much worst than what we have atm anyway...).

andrunko:
Maybe we should pass `-t` param before the other arguments but don't think that makes any diff.

andrunko (changes requested):

andrunko (approved):

ManMower:
Forced push changes the parameter order and correct the typo

PR #316, opened 2020-10-15
eos-image-boot-setup: Fix calls to busybox's losetup

jprvita:
Busybox losetup does not have --show. Its -f will show the next
available loop device, or, if you pass an image file, it will use (but
not show) the next available loop device.

https://phabricator.endlessm.com/T30937

### DIFF
--- a/dracut/image-boot/eos-image-boot-setup
+++ b/dracut/image-boot/eos-image-boot-setup
@@ -113,7 +113,8 @@ iso9660)
   fi
 
   # Create a loopback device backing the squashfs
-  if ! image_device=$(losetup --find --show /cd/${image_path}); then
+  image_device=$(losetup -f)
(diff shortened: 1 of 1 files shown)


dracut/image-boot/eos-image-boot-setup
```diff
@@ -113,7 +113,8 @@ iso9660)
   fi
 
   # Create a loopback device backing the squashfs
-  if ! image_device=$(losetup --find --show /cd/${image_path}); then
+  image_device=$(losetup -f)
+  if ! image_device=$(losetup ${image_device} /cd/${image_path}); then
```
wjt, line 117: There's a race here of course – what if something else uses the loop device between the two invocations of `losetup`? In practice we don't expect anything else to be using loop devices at this point in the boot process so I guess it's OK.

wjt (approved):

PR #321, opened 2020-11-11
T31079 Forwardport runtime migrations

mwleeds:
These are cherry-picks from the eos3.3 branch.

https://phabricator.endlessm.com/T31079

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -385,7 +385,8 @@ FLATPAKS_TO_MIGRATE = [
         'old-origin': 'eos-apps',
         'new-origin': 'flathub'
     },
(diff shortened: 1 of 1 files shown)


mwleeds:
Added one commit that's not a cherry-pick to fix a trivial CI failure

wjt (approved):

wjt:
Philosophically this seems a bit wrong because, if someone has crossed the checkpoint and switched to eos3a, then they no longer have the `gnome` remote. And if they're already stuck unable to upgrade due to this issue, then us fixing it in a future update won't help automatically unstick them.

But… they can use `ostree admin upgrade` to get round that and then it'll be autofixed. So, better than not doing it.

PR #322, opened 2020-11-12
Migrate Chromium to Flathub

andrunko:
General notes on migration:
- `$config_dir/Default/Preferences` have a few mentions to `/usr/lib/chromium-browser/resources/` which doesn't even exist in the current version. The values are not updated during migration.
- Some services credentials (e.g. gmail) are lost during migration (requiring re-login)
- Some log files have entries referring to `~/.config/chromium`. This is ignored and log files are unchanged as they may be useful in case we need to differ logs from old and new version
- The migration always runs if `~/.var/app/org.chromium.Chromium/.migrated` (created at the end of the migration) doesn't exist, even if `~/.var/app/org.chromium.Chromium` does:
  - This is such that we can update `~/.config/mimeapps.list` to point the new chromium in case the old chromium is specified as default there but the user never ran it
  - And also to update any old desktop shortcut on `~/.local/share/applications` that'd refer to the old chromium as the executable (`Exec=` starts with `/usr/bin/chromium-browser`)
  - Note that the actual user profile migration is skipped if `~/.var/app/org.chromium.Chromium` already exists or if        `~/.config/chromium` doesn't exist by the time the migration runs

https://phabricator.endlessm.com/T30997

### DIFF
--- a/eos-migrate-chromium-profile
+++ b/eos-migrate-chromium-profile
@@ -1,0 +1,150 @@
+#!/usr/bin/python3
+#
+# eos-migrate-chromium-profile: move users' Chromium profile to its new home
+#
+# This script is based on eos-migrate-firefox-profile.
+#
+# Copyright © 2020 Endless OS Foundation LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+import fileinput
+import os
+from gi.repository import GLib
+
+USER_HOME_DIR = os.path.expanduser("~/")
+DESKTOP_SHORTCUTS_DIR = os.path.join(USER_HOME_DIR, ".local", "share", "applications")
+OLD_CHROMIUM_CONFIG_DIR = os.path.join(USER_HOME_DIR, ".config", "chromium")
+NEW_CHROMIUM_DATA_DIR = os.path.join(USER_HOME_DIR, ".var", "app",
+                                     "org.chromium.Chromium")
+NEW_CHROMIUM_CONFIG_DIR = os.path.join(NEW_CHROMIUM_DATA_DIR, "config", "chromium")
+
+OLD_WIDEVINE_SYSTEM_DIR = os.path.join(os.path.sep, "usr", "lib",
+                                       "chromium-browser", "WidevineCdm")
+
+MIMEAPPS_LIST = os.path.join(GLib.get_user_config_dir(), "mimeapps.list")
+MIMEAPPS_GROUPS = [
+    "Default Applications",
+    "Added Associations",
+    "Removed Associations",
+]
+OLD_DESKTOP_FILE = "chromium-browser.desktop"
+NEW_DESKTOP_FILE = "org.chromium.Chromium.desktop"
+
+
+def update_mimeapps_list(path):
+    """Update file associations, which in particular includes x-scheme-handler/http and
+    friends to specify the default web browser.
+
+    We cannot use GLib's own API to query what mime types the old Chromium desktop file
+    is a handler for, because we can't construct a GDesktopAppInfo for it, because its
+    desktop file no longer exists.
+    """
+
+    keyfile = GLib.KeyFile()
+    try:
+        keyfile.load_from_file(
+            path, GLib.KeyFileFlags.KEEP_COMMENTS | GLib.KeyFileFlags.KEEP_TRANSLATIONS,
+        )
+    except GLib.GError as gerror:
+        if gerror.matches(GLib.file_error_quark(), GLib.FileError.NOENT):
+            return
+
+        raise
+
+    changed = False
+
+    for group in MIMEAPPS_GROUPS:
+        if not keyfile.has_group(group):
+            continue
+
+        keys, _length = keyfile.get_keys(group)
+        for key in keys:
+            values = keyfile.get_string_list(group, key)
+            try:
+                i = values.index(OLD_DESKTOP_FILE)
+            except ValueError:
+                pass
+            else:
+                values[i] = NEW_DESKTOP_FILE
+                keyfile.set_string_list(group, key, values)
+                changed = True
+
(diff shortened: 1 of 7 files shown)


andrunko:
Bot, retest this please.

andrunko:
CI is failing with some linter errors, I will update it later but would appreciate any input here in the meantime.

andrunko:
Converted to draft as I think we should only merge this and related changes once chromium hits flathub stable (atm we don't configure a flathub-beta remote and thus can't install it).

eos-migrate-chromium-profile
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/python3
+#
+# eos-migrate-chromium-profile: move users' Chromium profile to its new home
+#
+# This script is based on eos-migrate-firefox-profile.
```
wjt, line 5: A bit of a sad amount of copypasta but I guess we can live with it for the upside of these being simple scripts.

eos-migrate-chromium-profile
```diff
@@ -0,0 +1,122 @@
...
+#
+# Copyright © 2020 Endless OS LLC
```
wjt, line 7:
```suggestion
# Copyright © 2020 Endless OS Foundation LLC
```

eos-migrate-chromium-profile
andrunko: Yeah, exactly.

eos-migrate-chromium-profile
andrunko: Done.

andrunko:
Force pushed to fix linter errors (no actual change).

eos-migrate-chromium-profile
```diff
@@ -0,0 +1,151 @@
...
+    if not keyfile.has_group(group):
+        return
+
+    exec_cmd = keyfile.get_string(group, "Exec")
+    if "/usr/bin/chromium-browser" not in exec_cmd:
+        return
+
+    keyfile.set_string(group, "Exec",
+                       exec_cmd.replace("/usr/bin/chromium-browser",
+                                        "/usr/bin/flatpak run org.chromium.Chromium"))
```
wjt, line 104: It's an edge case but I think this should be:

```python3
if not exec_cmd.startswith("/usr/bin/chromium-browser"):
    return

keyfile.set_string(group, "Exec", "/usr/bin/flatpak run org.chromium.Chromium" + exec_cmd[len("/usr/bin/chromium-browser"):])
```

rather than replacing it anywhere in the string.

eos-migrate-chromium-profile
```diff
@@ -0,0 +1,151 @@
...
+
+    update_mimeapps_list(MIMEAPPS_LIST)
+    update_desktop_shortcuts()
+
+    if (
+        os.path.isdir(OLD_CHROMIUM_CONFIG_DIR) and
+        not os.path.isdir(NEW_CHROMIUM_CONFIG_DIR)
+    ):
+        os.makedirs(NEW_CHROMIUM_CONFIG_DIR)
+        os.rename(OLD_CHROMIUM_CONFIG_DIR, NEW_CHROMIUM_CONFIG_DIR)
```
wjt, line 139: TIL that you can rename over a directory that exists. I would normally spell this as:

```python3
        os.makedirs(os.path.dirname(NEW_CHROMIUM_CONFIG_DIR), exist_ok=True)
        os.rename(OLD_CHROMIUM_CONFIG_DIR, NEW_CHROMIUM_CONFIG_DIR)
```

but it does appear that what you have works.

eos-migrate-chromium-profile
```diff
@@ -0,0 +1,151 @@
...
+        update_old_config_references(
+            os.path.join(NEW_CHROMIUM_CONFIG_DIR,
+                         "WidevineCdm", "latest-component-updated-widevine-cdm"))
+
+    if not os.path.exists(NEW_CHROMIUM_DATA_DIR):
+        os.makedirs(NEW_CHROMIUM_DATA_DIR)
```
wjt, line 145:
```
os.makedirs(NEW_CHROMIUM_DATA_DIR, exist_ok=True)
```

eos-migrate-chromium-profile
```diff
@@ -0,0 +1,151 @@
...
+
+    os.mknod(migrated_file)
```
wjt, line 147: I've never seen `mknod` used to create a normal file before!

wjt (changes requested):

eos-migrate-chromium-profile
andrunko: From what I checked this was the only way to create files without opening... Would you prefer to change it?

wjt: No, it's fine! I have always used:

```python3
with open(migrated_file, "w"):
    pass
```

to touch a file. But this (which merely needs to create, not update a timestamp) works fine too.

eos-migrate-chromium-profile
andrunko: Done.

eos-migrate-chromium-profile
andrunko: My idea here is that we wouldn't migrate the user profile if the user had already ran the flatpak (`NEW_CHROMIUM_CONFIG_DIR` already exists). In that case we would only update the mime associations and desktop shortcuts.

Left as is but can update if you prefer.

eos-migrate-chromium-profile
andrunko: Done.

eos-migrate-chromium-profile
wjt: Yes, I understand the intent of the `not os.path.isdir()` check. I was more surprised that creating an empty directory then renaming over it works.

eos-migrate-chromium-profile
```diff
@@ -0,0 +1,150 @@
...
+
+    if (
+        os.path.isdir(OLD_CHROMIUM_CONFIG_DIR) and
+        not os.path.isdir(NEW_CHROMIUM_CONFIG_DIR)
+    ):
+        os.makedirs(NEW_CHROMIUM_CONFIG_DIR)
+        os.rename(OLD_CHROMIUM_CONFIG_DIR, NEW_CHROMIUM_CONFIG_DIR)
+        update_old_config_references(
+            os.path.join(NEW_CHROMIUM_CONFIG_DIR,
+                         "WidevineCdm", "latest-component-updated-widevine-cdm"))
```
wjt, line 142: If I were feeling pedantic I'd say: well, this should run before the rename, because otherwise if the script gets interrupted between the `rename` and this block, then gets re-run on the next boot, the Widevine stuff won't update.

I think it's unlikely.

wjt (approved):

eos-migrate-chromium-profile
andrunko: Ah, I misread your comment and didn't notice the `os.path.dirname`, fixed it now.

eos-migrate-chromium-profile
andrunko: Done.

PR #323, opened 2020-11-16
eos-update-flatpak-repos: Fix a missing ref edge case

mwleeds:
A user on the forum hit this error:

```
INFO:root:Found gnome:runtime/org.gnome.Platform/x86_64/3.28 to migrate to flathub:runtime/org.gnome.Platform/x86_64/3.28
ERROR:root:Failure applying migration to org.gnome.Platform
  File "/sysroot/home/irena/eos-update-flatpak-repos", line 1698, in _migrate_installed_flatpaks
    migrate_extra = is_extra_data(repo, ostree_refs[old_ostree_ref])
KeyError: 'gnome:runtime/org.gnome.Platform/x86_64/3.28'
```

So, make the code a bit more robust so that it doesn't error out when
the old ref doesn't exist and we can still migrate the deploy file.
We had already encoded the assumption that the old ref might not exist
with the check "if old_ostree_ref in ostree_refs and not new_ref:", so
this just makes it so we don't error out in that case.

Reference:
https://community.endlessos.com/t/issues-with-updates-i-am-not-able-to-update-anything/14838

https://phabricator.endlessm.com/T31079

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -1580,6 +1580,9 @@ def migrate_extra_data(old_ref, new_ref):
     new_deploy = new_ref.get_deploy_dir()
     new_extra = os.path.join(new_deploy, 'files', 'extra')
 
+    if not os.path.exists(old_extra):
+        return
+
     if os.path.exists(new_extra):
         # either the user manually installed Skype from Flathub, or we got
         # this far but died before uninstalling the old ref last time
@@ -1695,7 +1698,6 @@ def _migrate_installed_flatpaks():
 
             try:
                 vendor_prefixes = frozenset()
-                migrate_extra = is_extra_data(repo, ostree_refs[old_ostree_ref])
 
                 # copy the old ref to the new one, pointing at the same commit
                 if old_ostree_ref in ostree_refs and not new_ref:
@@ -1703,6 +1705,7 @@ def _migrate_installed_flatpaks():
                                  .format(old_ostree_ref,
                                          ostree_refs[old_ostree_ref],
                                          new_ostree_ref))
+                    migrate_extra = is_extra_data(repo, ostree_refs[old_ostree_ref])
                     repo.prepare_transaction()
                     try:
                         kwargs = {}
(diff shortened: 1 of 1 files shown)


eos-update-flatpak-repos
```diff
@@ -1742,15 +1745,18 @@ def _migrate_installed_flatpaks():
...
-                    triggers_needed = True
+                    try:
+                        logging.info('Uninstalling old ref: {}'.format(old_ostree_ref))
+                        inst.uninstall_full(Flatpak.UninstallFlags.NO_PRUNE |
+                                            Flatpak.UninstallFlags.NO_TRIGGERS,
+                                            kind, name, arch, branch)
+
+                        triggers_needed = True
+                    except GLib.Error as err:
+                        if not err.matches(Flatpak.error_quark(), Flatpak.Error.NOT_INSTALLED):
```
wjt, line 1758:
```
/build/src/eos-boot-helper_1+dev740.c7036ab/eos-update-flatpak-repos:1758:89: E501 line too long (95 > 88 characters)
```

eos-update-flatpak-repos
```diff
@@ -1580,6 +1580,9 @@ def migrate_extra_data(old_ref, new_ref):
     new_deploy = new_ref.get_deploy_dir()
     new_extra = os.path.join(new_deploy, 'files', 'extra')
 
+    if not os.path.exists(old_extra):
```
wjt, line 1583: Under what situation would this code path run without the old extra-data existing?

wjt: (I think it's fine to be robust against it not existing but I can't see how it would happen.)

eos-update-flatpak-repos
```diff
@@ -1742,15 +1745,18 @@ def _migrate_installed_flatpaks():
...
-                                        kind, name, arch, branch)
-
-                    triggers_needed = True
+                    try:
+                        logging.info('Uninstalling old ref: {}'.format(old_ostree_ref))
+                        inst.uninstall_full(Flatpak.UninstallFlags.NO_PRUNE |
+                                            Flatpak.UninstallFlags.NO_TRIGGERS,
+                                            kind, name, arch, branch)
+
+                        triggers_needed = True
```
wjt, line 1756: I think this needs to be set unconditionally since if we have reached this point, we have definitely installed the new ref, so triggers (update exported files etc.) do need to run.

wjt (changes requested):

eos-update-flatpak-repos
mwleeds: In case the ref is just a regular one not an extra-data one

eos-update-flatpak-repos
mwleeds: Good point

eos-update-flatpak-repos
mwleeds: Fixed

eos-update-flatpak-repos
wjt: Ah, I see, the caller now calls this regardless of whether the app is extra-data or not. Gotcha!

wjt (approved):

PR #325, opened 2020-12-09
eos-update-flatpak-repos: Use the correct kind for app extensions

wjt:
Without this change, having matched `runtime/com.hack_computer.Clubhouse.Locale/x86_64/eos3`, the script would try and fail to uninstall `app/com.hack_computer.Clubhouse.Locale/x86_64/eos3`.

Unfortunately the error was masked by c262b983eaf649b91bdb0c9a25d2d3214bee4ea2.

Add a warning in the previously silently-failing case, and fix the logic error that caused the wrong thing to be uninstalled.

https://phabricator.endlessm.com/T31185

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -1704,6 +1704,7 @@ def _migrate_installed_flatpaks():
 
         for ref in matching_refs:
             origin = ref.get_origin()
+            kind = ref.get_kind()
             name = ref.get_name()
             arch = ref.get_arch()
             branch = ref.get_branch()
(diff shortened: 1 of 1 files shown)


wjt:
Pushed an alternative patch based on @ramcq's Slack review pointing out that we already reassign `name` from the outer loop in the inner loop, so why not also reassign `kind`.

I think this is confusing and hard to reason about but it is consistent with the existing code which is also confusing and hard to reason about. :stuck_out_tongue_closed_eyes: Really this giant function should be refactored but there is substantial shared state between the inner and outer loops so it is not a quick job.

erikos:
Great spotting! And thanks for fixing, the explanation makes sense. Good from my side.

PR #328, opened 2021-01-05
eos-update-flatpak-repos: Workaround to fix issue migrating com.hack_computer_Clubhouse.Locale

andrunko:
This adds an workaround to fix an issue when migrating `com.hack_computer_Clubhouse.Locale` from `eos-apps` to `flathub`. The current migration fails when attempting to update the runtime when partially installed (e.g. only has a subset of the available
directories, corresponding to the locales on the system). To work around that we make sure to uninstall the old (from `eos-apps`) runtime before attempting the migration.

That has the side effect that the new `.Locale` won't get installed until the app (Clubhouse) gets updated next (either via automatic updates or manually) or if the user manually installs it.

Given the `.Locale` runtime is not mandatory the app should still work fine.

https://phabricator.endlessm.com/T31185

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -132,6 +132,11 @@ def _remove_remotes():
 
 # dict of remote -> list of runtimes
 RUNTIMES_TO_REMOVE = {
+    "eos-apps": [
+        # remove partially-installed, empty extension (which can't currently
+        # be migrated) before migrating main app (T31185)
+        "com.hack_computer.Clubhouse.Locale",
+    ],
 }
 
 
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -157,7 +157,7 @@ def _remove_runtimes():
         logging.info("Removing runtime {} from {}"
                      .format(refspec, inst.get_path().get_path()))
         subprocess.check_call(['flatpak', 'uninstall', '--system',
-                               '--runtime', refspec])
+                               '--runtime', '--noninteractive', refspec])
 
 
 # From 3.7.6 through 3.7.8, images were shipped with remote URLs using


eos-update-flatpak-repos
```diff
@@ -152,7 +155,7 @@ def _remove_runtimes():
         logging.info("Removing runtime {} from {}"
                      .format(refspec, inst.get_path().get_path()))
         subprocess.check_call(['flatpak', 'uninstall', '--system',
-                               '--runtime', refspec])
+                               '--runtime', '--noninteractive', refspec])
```
andrunko, line 158: Unsure if it should use `-y` instead or even if really needed (but it seems to be from my tests, otherwise the systemd service fails when getting a prompt).

eos-update-flatpak-repos
```diff
@@ -132,6 +132,9 @@ def _remove_remotes():
 
 # dict of remote -> list of runtimes
 RUNTIMES_TO_REMOVE = {
+    "eos-apps": [
+        "com.hack_computer.Clubhouse.Locale",
```
wjt, line 136:
```suggestion
        # Remove partially-installed, empty extension (which can't currently be migrated) before migrating main app (T31185)
        "com.hack_computer.Clubhouse.Locale",
```

except with appropriate line wrapping.

wjt (approved):
Functionally looks fine. One comment suggestion and there are some typos in the commit message. I took the liberty of rewording it:

```
eos-update-flatpak-repos: Don't try to migrate partially-installed runtime

com.hack_computer.Clubhouse.Locale cannot currently be migrated from eos-apps
to flathub together with the main app.  The current implementation fails
because the runtime is partially installed – i.e. only a subset of the
directories are present, corresponding to the locales on the system.

The .Locale runtime is optional, and does not contain any (meaningful)
translations, so the app works fine without it.

Work around the limitation by removing the runtime before migrating the app.
When Clubhouse is next updated from Flathub, either automatically or manually,
the .Locale extension will be re-installed.

https://phabricator.endlessm.com/T31185
```

Please merge after making these changes.

andrunko:
bot, retest this please.

PR #330, opened 2021-01-12
eos-live-boot-overlayfs-setup: Limit blkid output to one device

jprvita:
If eos-image-boot-dm-setup.service has already run when this script
runs, blkid will return two devices matching the token passed via
endless.image.device, the real partition and the dm-device created by
eos-image-boot-dm-setup.

Limiting the output of blkid to one device only fixes a problem with the
logic in eos-live-boot-overlayfs-setup that was preventing
persistent.img from getting mapped when that happened, since it expects
only one device to be output.

https://phabricator.endlessm.com/T31223


ManMower (approved):
Useful changes. This stuff is Very Hard to debug, and these changes will help a lot.

PR #331, opened 2021-02-16
chromium: Don't crash if .local/share/applications doesn't exist

wjt:
os.listdir() raises FileNotFoundError if the given directory doesn't
exist. Previously, eos-migrate-chromium-profile would crash in this
case, such as for every new user profile:

      File "/usr/lib/eos-boot-helper/eos-migrate-chromium-profile", line 150, in <module>
        main()
      File "/usr/lib/eos-boot-helper/eos-migrate-chromium-profile", line 132, in main
        update_desktop_shortcuts()
      File "/usr/lib/eos-boot-helper/eos-migrate-chromium-profile", line 110, in update_desktop_shortcuts
        for filename in os.listdir(DESKTOP_SHORTCUTS_DIR):
    FileNotFoundError: [Errno 2] No such file or directory: '/sysroot/home/erikos/.local/share/applications'

In the case where the user had an old Chromium profile at
~/.config/chromium but no ~/.local/share/applications, this would mean
that their Chromium profile would not be migrated to the
Flatpak-compliant location.

https://phabricator.endlessm.com/T31446

### DIFF
--- a/eos-migrate-chromium-profile
+++ b/eos-migrate-chromium-profile
@@ -107,7 +107,12 @@ def update_desktop_shortcut(path):
 
 
 def update_desktop_shortcuts(path=DESKTOP_SHORTCUTS_DIR):
-    for filename in os.listdir(path):
+    try:
+        shortcuts = os.listdir(path)
+    except FileNotFoundError:
+        return
+
+    for filename in shortcuts:
         if filename.endswith(".desktop"):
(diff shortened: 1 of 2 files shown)


wjt:
Now with improved test coverage!

jprvita:
retest this please

wjt:
Updated to fix the linter warning.

jprvita (approved):

PR #333, opened 2021-02-25
Remove xterm dependency

andrunko:
This dep is not used by anything on eos-boot-helper.

https://phabricator.endlessm.com/T31519

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -25,7 +25,6 @@ Description: Endless boot helper
 Depends: ${misc:Depends},
          dracut,
          xinit,
-         xterm,
          iptables,
          ostree,
          flatpak,
--- a/debian/control
+++ b/debian/control
@@ -24,7 +24,6 @@ Architecture: any
 Description: Endless boot helper
 Depends: ${misc:Depends},
          dracut,
-         xinit,
          iptables,
          ostree,
          flatpak,


andrunko:
@dsd I see you added this dep when originally adding the debian packaging but it doesn't seem used - if any other packages use it they should depend on it directly, or am I missing something?

dsd:
I'm pretty sure this is used directly by the factory test hook included in eos-boot-helper - it opens an X environment with an xterm and then shows the script output inside. But let me know if that's not the case any more or if I'm missing something.

Also previous users of the factory test script would open more xterms to be able to show things on different parts of the screen, or to fill the screen, etc.

andrunko:
From what I checked the xterm dep is actually on eos-factory-tools and not eos-boot-helper - [here](https://github.com/endlessm/eos-factory-tools/blob/master/factory-test/eos-factory-test.service#L7).

I will double check tomorrow.

dsd:
Ah - that sounds about right.

debian/control
```diff
@@ -25,7 +25,6 @@ Description: Endless boot helper
 Depends: ${misc:Depends},
          dracut,
          xinit,
```
wjt, line 27: Isn't the `xinit` dependency actually only for `eos-factory-tools` too?

andrunko: Indeed, updated - checked and `eos-factory-tools` already includes it.

wjt (approved):

PR #335, opened 2021-03-23
Add plumbing to add openh264 flatpak runtime to library search path when available

andrunko:
This change adds a tmpfiles.d config to link the openh264 flatpak runtime at /run/lib/openh264, a ld.so.conf.d config to make this dir available in the library search path and systemd units to re-run ldconfig when the runtime is installed/uninstalled in order to make it to the ldconfig cache.
    
We use a watcher service in order to re-run ldconfig (restart the unit responsible to run it) as path units can only start other units, not restart them.
    
https://phabricator.endlessm.com/T23316

### DIFF
--- a/flatpak-extension-runtime-fdo-openh264-watcher.service
+++ b/flatpak-extension-runtime-fdo-openh264-watcher.service
@@ -1,0 +1,9 @@
+[Unit]
+After=systemd-tmpfiles-setup.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/systemctl restart flatpak-extension-runtime-fdo-openh264-ldconfig.service
+
+[Install]
+WantedBy=multi-user.target
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,6 +25,9 @@ dist_systemdunit_DATA = \
 	eos-live-boot-overlayfs-setup.service \
 	eos-reclaim-swap.service \
 	eos-update-flatpak-repos.service \
+	flatpak-extension-runtime-fdo-openh264-ldconfig.service \
+	flatpak-extension-runtime-fdo-openh264-watcher.path \
+	flatpak-extension-runtime-fdo-openh264-watcher.service \
 	hack-remove-global-override.service \
 	$(NULL)
 
@@ -111,6 +114,7 @@ dist_tmpfiles_DATA = \
 	tmpfiles.d/avahi-service-writers.conf \
 	tmpfiles.d/chromium-system-services.conf \
 	tmpfiles.d/eos-boot-helper.conf \
+	tmpfiles.d/flatpak-extension-runtime-fdo-openh264.conf \
 	tmpfiles.d/flatpak-overrides.conf \
 	$(NULL)
 
@@ -119,3 +123,6 @@ dist_extratmpfiles_DATA = \
 	tmpfiles.d/flatpak-extension-kiwix-usb-content.conf \
 	$(NULL)
 
+ldconfdir = $(sysconfdir)/ld.so.conf.d
+dist_ldconf_DATA = \
+	ld.so.conf.d/flatpak-extension-runtime-fdo-openh264.conf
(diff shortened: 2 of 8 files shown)


flatpak-extension-runtime-fdo-openh264-watcher.path
```diff
@@ -0,0 +1,8 @@
+[Unit]
+After=systemd-tmpfiles-setup.service
+
+[Path]
+PathChanged=/var/lib/flatpak/runtime/org.freedesktop.Platform.openh264/x86_64/2.0/active/files/extra/libopenh264.so.6
```
andrunko, line 5: Just a note that I tried watching files under /run/lib/openh264 instead but in this case it would only trigger the unit when installing the runtime but not when uninstalling it.

Also tried watching the dir itself, but that is racy as the unit may be triggered before all files are written to the dir.

dbnicholson: In the tmpfiles snippet you're substituting `@FLATPAK_ARCH@`, but here `x86_64` is hardcoded.

flatpak-extension-runtime-fdo-openh264-watcher.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+After=systemd-tmpfiles-setup.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/systemctl restart flatpak-extension-runtime-fdo-openh264-ldconfig.service
```
dbnicholson, line 6: Instead of having a service that starts another service, is it possible to just have `flatpak-extension-runtime-fdo-openh264-ldconfig.path` that watches the that flatpak path?

andrunko: No, unless we only want it to be triggered once (e.g. not on every install/uninstall of the runtime) from what I checked - it will start the service but not restart it.

dbnicholson: You would have to drop the `RemainAfterExit=yes` from the service, though.

flatpak-extension-runtime-fdo-openh264-watcher.path
andrunko: Good catch, should be fixed now.

flatpak-extension-runtime-fdo-openh264-watcher.service
andrunko: Truth being told I got this pattern from https://superuser.com/questions/1171751/restart-systemd-service-automatically-whenever-a-directory-changes-any-file-ins and could make make it work reliably otherwise.

andrunko: I tried dropping RemainAfterExit (and even explicitly setting it to no) but could not make it work, but I tried so many different combinations that I may have let something slip.

dbnicholson: I think we can just go forward with this, but here's what I just tested:
test.service:
```
[Service]
Type=oneshot
ExecStart=/bin/true

[Install]
WantedBy=multi-user.target
```
test.path:
```
[Path]
PathChanged=/var/tmp/foo

[Install]
WantedBy=multi-user.target
```
The service is started at boot from multi-user.target, so if the runtime is already installed, `ldconfig` will be run. But if the path changes at any time after that then it will be run again.

Anyways, we can always fine tune this later if needed. What you have here should work fine.

dbnicholson (approved):

flatpak-extension-runtime-fdo-openh264-watcher.service
andrunko: It may be the case I missed something but I'm pretty sure that was the first thing I tried and hit some issues (maybe because the dirs are symlinked and what not). As you said I would go with this and fine tune later if needed.

PR #337, opened 2021-04-30
Split flatpak repo out of system ostree repo

dbnicholson:
This tool splits the flatpak repo resources out of /ostree/repo and sets
up /var/lib/flatpak/repo up as a normal flatpak repo.

https://phabricator.endlessm.com/T31877


dbnicholson:
This is ready for review now. I changed the implementation so that it duplicates the repo and then removes unwanted things so that it doesn't have to deal with pulling partial commits.

The only part that needs to be fixed is to fill in the locking whenever https://phabricator.endlessm.com/T31868 is figured out.

eos-split-flatpak-repo
```diff
@@ -0,0 +1,307 @@
...
+            # Most likely var/lib/flatpak is a symlink to
+            # sysroot/flatpak that was needed to ensure the checkouts
+            # were on the same filesystem as the repo. Undo that.
+            if os.path.islink(flatpak_dir_path):
+                real_flatpak_dir_path = os.path.realpath(flatpak_dir_path)
+                logger.info('Replacing Flatpak dir symlink %s with %s',
+                            flatpak_dir_path, real_flatpak_dir_path)
+                replace_symlink(flatpak_dir_path, real_flatpak_dir_path)
+
+            # Remove Flatpak remotes and refs from the flatpak repo
```
pwithnall, line 282: “…from the OSTree repo”

eos-split-flatpak-repo
```diff
@@ -0,0 +1,307 @@
...
+            flatpak_repo_config = new_flatpak_repo.copy_config()
+            flatpak_repo_config.set_value('core', 'mode', 'bare-user-only')
+            new_flatpak_repo.write_config(flatpak_repo_config)
+
+            # Swap the new flatpak repo into place.
+            logger.info('Replacing Flatpak repo %s with %s',
+                        flatpak_repo_path, new_flatpak_repo_path)
+            replace_symlink(flatpak_repo_path, new_flatpak_repo_path)
+
+            # Most likely var/lib/flatpak is a symlink to
```
pwithnall, line 273: A failure after this point will not be solvable by re-running the script (since the flatpak repo will have been separated by this point). Should the code below try a bit harder to carry on in the face of error (for example, pruning the OSTree repo even if the `flatpak_dir_path` symlink can’t be fixed)?

Overall, what’s the error handling strategy for the script? Mostly it seems to be relying on everything being atomic on the `replace_symlink()` call above, which seems reasonable. Might be good to have that documented.

I can’t see a way to make the code below fit in with the above atomic switch to make it inherently safe.

eos-split-flatpak-repo.service
```diff
@@ -0,0 +1,24 @@
...
+# be exclusively locked.
+Before=eos-autoupdater.service eos-updater.service
+Before=eos-update-flatpak-repos.service
+Before=eos-updater-avahi.service eos-update-server.service
+Before=eos-updater-flatpak-installer.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=/usr/sbin/eos-split-flatpak-repos
```
pwithnall, line 21: If `--prune` is not passed by default, when does pruning happen?

tests/test_split_flatpak_repo.py
```diff
@@ -0,0 +1,359 @@
+#!/usr/bin/python3
+"""
+Tests eos-split-flatpak-repo
+"""
```
pwithnall, line 4: Is a copyright header needed?

pwithnall:
This looks pretty great to me. I have only nitpicks (sorry).

eos-split-flatpak-repo
```diff
@@ -0,0 +1,307 @@
...
+    os.rename(link, link_tmp)
+    try:
+        logger.info('Renaming %s to %s', path, link)
+        os.rename(path, link)
+    except:  # noqa: E722
+        logger.error('Restoring link %s from %s', link, link_tmp)
+        os.rename(link_tmp, link)
+        raise
+    logger.info('Removing old link %s', link_tmp)
+    os.unlink(link_tmp)
```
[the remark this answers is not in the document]
dbnicholson: Oh, good call.

eos-split-flatpak-repo
dbnicholson: Right.

eos-split-flatpak-repo
dbnicholson: Yeah, I really couldn't figure out what to do here since it's orthogonal but related to the actual repo split. Everything will continue to work fine if `/var/lib/flatpak` is a symlink to `/sysroot/flatpak`, but it's not clean. And currently if there's an error at this point rerunning the script won't address it since it only checks if `/var/lib/flatpak/repo` is a symlink.

I guess there should really be 2 distinct checks and actions:

* Check if `/var/lib/flatpak/repo` is a symlink and split the repos if it is
* Check if `/var/lib/flatpak` is a symlink and replace the symlink with the real flatpak dir

This also affects the systemd unit, which I haven't tested and currently only has `ConditionPathIsSymbolicLink=/var/lib/flatpak/repo`. I believe that would have to be updated to:
```
ConditionPathIsSymbolicLink=|/var/lib/flatpak
ConditionPathIsSymbolicLink=|/var/lib/flatpak/repo
```

eos-split-flatpak-repo.service
dbnicholson: Good question, this should probably be discussed somewhere here. The repos would be pruned the next time something else prunes the repos. So long as the refs are removed, the objects will be cleaned up. This would happen when eos-updater deploys a new OS and gnome-software/flatpak update or install a flatpak in the system repo. The presumption is that would happen sooner or later, but obviously there's no guarantee.

The idea is to make this operation fast since the prune would be the slow part. Since the objects are all hard linked during the split, it won't initially use any extra disk space. However, until both repos have been pruned there will be some excess. For instance, if you deployed a new OS and `/ostree/repo` was pruned, it would cleanup both the unused flatpak objects as well as the non-rollback OS objects. However, until `/var/lib/flatpak/repo` was pruned, those non-rollback OS objects would still exist on disk. Maybe this is a premature optimization, but I've only run this on smallish repos.

tests/test_split_flatpak_repo.py
dbnicholson: :+1:

dbnicholson:
The other thing I forgot about here was to sync after splitting the repos to make sure the changes actually reach the disk.

eos-split-flatpak-repo
pwithnall: > I guess there should really be 2 distinct checks and actions:

Sounds reasonable to me.

eos-split-flatpak-repo.service
pwithnall: As long as flatpak and eos-updater both prune after doing something which makes the repository content significantly change, I think this approach is fine. It would be good to have it documented in a comment somewhere (you could probably copy/paste what you wrote above).

eos-split-flatpak-repo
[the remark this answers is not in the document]
dbnicholson: Actually, the first `os.rename()` call does create `link_tmp`. The call to `tempfile.mktemp` is the unsafe version where it only generates a random name for you without actually creating the file. I think `tempfile.mkstemp` could be used to create the file and rename over it, which would require unlinking it if the renaming fails.

pwithnall:
LGTM

pwithnall (approved):

dbnicholson:
Thanks. I'm merging this and will address the FIXME for the repo locking if/when the locking API becomes public.

PR #339, opened 2021-05-21
Migrate Freecell and Solitaire to Aisleriot, and fix a few other migrations

wjt:
These two apps were not previously migrated because they are both superseded by the same app, and we had wanted to make `org.gnome.Aisleriot.desktop` contain both in `X-Flatpak-RenamedFrom`.

I don't think this actually matters in practice, though. It will have the first one immediately, and then after being upgraded to the real Flathub version it will have both.

I also fixed a regression caused by a Flatpak becoming stricter, which broke the migration of at least 4 other apps.

https://phabricator.endlessm.com/T23936

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -1510,7 +1510,7 @@ COMMIT_BODY_INDEX = 4
 
 def copy_commit(repo, src_rev, dest_ref,
                 old_app_id=None, new_app_id=None,
-                collection_id=None, migrate_extra=False,
+                collection_id=None,
                 mangle_metadata=None):
     """Copy commit src_rev to dest_ref
     This makes the new commit at dest_ref have the proper collection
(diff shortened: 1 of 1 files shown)


ramcq:
> The actual migration of extra-data is conditional on the app ID and/or
> branch having changed, which is weird, but is fine in practice because
> that is always the case for extra-data apps we migrate.

I think in the case the app ID or branch has not been renamed, and we are just switching the origin in the deploy file, the extra data does not need to be copied as it's already in the correct place. So perhaps this commit message hunk could be dropped?

ramcq:
Do we have any data on whether the application is more frequently launched in its Solitaire guise compared to its Freecell guise? Or whether we tend to favour one or the other when populating our default icon grids? Am wondering whether it might make sense to put Solitaire first so that's the icon that's kept, if it's more popular.

wjt:
You're right, Solitaire is 50% more popular

Https://metabase.endlessm.com/question/182

wjt:
Dropped that commit message hunk & swapped the order.

The funny thing is… if you have both pinned to the taskbar before the upgrade, then you end up with two copies of the same app on the taskbar after the upgrade, at least until you log out and in again.

pwithnall (approved):

PR #340, opened 2021-05-21
eos-split-flatpak-repo: Handle refs from deleted remotes

dbnicholson:
It's unusual but not unexpected to have refs leftover from a remote that
was deleted. Previously an error was raised, but now the tool does it's
best to figure out if it was a flatpak or OS ref. If it can't be figured
out, the ref will stay in both repos.

https://phabricator.endlessm.com/T31877


dbnicholson:
Wrong task/branch.

PR #341, opened 2021-05-21
eos-split-flatpak-repo: Handle refs from deleted remotes

dbnicholson:
It's unusual but not unexpected to have refs leftover from a remote that
was deleted. Previously an error was raised, but now the tool does it's
best to figure out if it was a flatpak or OS ref. If it can't be figured
out, the ref will stay in both repos.

https://phabricator.endlessm.com/T32002

### DIFF
--- a/eos-split-flatpak-repo
+++ b/eos-split-flatpak-repo
@@ -114,6 +114,53 @@ def gather_remotes(repo):
     return os_remotes, flatpak_remotes
 
 
+def get_refspec_repo_type(refspec, os_remotes, flatpak_remotes):
+    """Determine the repo type this refspec should be a part of
+
+    Returns a RepoType or None if it can't be determined.
+    """
+    _, remote, ref = OSTree.parse_refspec(refspec)
+    if remote:
+        if remote in os_remotes:
+            return RepoType.OS
+        elif remote in flatpak_remotes:
+            return RepoType.Flatpak
+        else:
+            # The remote has been deleted, so fall through to the
+            # ref heuristics.
+            logger.warning('Refspec %s remote %s no longer exists',
+                           refspec, remote)
+
+    ref_parts = ref.split('/')
+    if len(ref_parts) > 1:
+        ref_prefix = ref_parts[0]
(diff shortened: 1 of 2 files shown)


eos-split-flatpak-repo
```diff
@@ -114,47 +114,69 @@ def gather_remotes(repo):
...
+    """Determine the repo type this refspec should be a part of
+
+    Returns a RepoType or None if it can't be determined.
+    """
+    _, remote, ref = OSTree.parse_refspec(refspec)
+    if remote:
+        if remote in os_remotes:
+            return RepoType.OS
+        elif remote in flatpak_remotes:
+            return RepoType.Flatpak
```
pwithnall, line 127: Is it possible for the remote to be in both `os_remotes` and `flatpak_remotes`?

eos-split-flatpak-repo
```diff
@@ -114,47 +114,69 @@ def gather_remotes(repo):
...
+            # ran flatpak build-update-repo.
+            logger.debug('Ignoring generated ref: %s', ref)
+            return None
+    elif ref == 'ostree-metadata':
+        # Even if this came from a remote we don't know what type it
+        # was, so just ignore it.
+        logger.debug('Ignoring generated ref: %s', ref)
+        return None
+    else:
+        # XXX: Should this fail here?
```
pwithnall, line 159: I think it’s fair to warn and return `None` so the ref is kept in both new repos, as you are doing.

pwithnall (approved):

eos-split-flatpak-repo
dbnicholson: It shouldn't be the way they're constructed, but this function doesn't know that. I think an assert that they're disjoint sets would be useful. I'll add that.

dbnicholson:
Added a fixup to ensure the remotes don't overlap. I'll squash and merge if that looks OK.

pwithnall (approved):
:ship:

wjt (approved):

PR #345, opened 2021-06-10
eos-prune-printers: Add a service to prune printers

starnight:
EOS 4 will change printer mechanism to driverless. So, there will be
transition action when users' computer upgrade from eos3a to eos4.
Add eos-prune-printers.servive to prune the installed printers and
drivers.

https://phabricator.endlessm.com/T31862

### DIFF
--- a/eos-prune-printers
+++ b/eos-prune-printers
@@ -1,0 +1,45 @@
+#!/usr/bin/env sh
+
+# eos-prune-printers: Purge all installed printers and corresponding drivers
+#
+# Copyright (C) 2021 Endless OS Foundation LLC
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, see <http://www.gnu.org/licenses/>.
+
+# Exit if the stamp file exists. The service file checks for this, but also
+# guard against an accidental execution.
+stamp_file=/var/lib/eos4-prune-printers
+[ -e $stamp_file ] && exit 0
+
+# Purge and cancel all print jobs
+cancel -xa
+
+# Delete all printers with CUPS tool
+for printer in $(lpstat -p | awk '/^printer/{print $2}')
+do
+	echo Deleting "$printer"
+	lpadmin -x "$printer"
+done
+
+# Remove the printer drivers and residual print cache files
+systemctl stop cups
+
+# remove stamp file from eos-config-printer (from 4.0 use driverless printers)
+rm -rf /var/lib/eos-config-printer
+# clean up the residual print cache files
+rm -rf /var/cache/cups/*
+
+systemctl start cups
+
+touch $stamp_file
--- a/eos-prune-printers.service
+++ b/eos-prune-printers.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Endless prune printers
+After=cups.service
+Requires=cups.service
+ConditionPathExists=!/var/lib/eos4-printers-pruned
+ConditionKernelCommandLine=!endless.live_boot
+
+[Service]
+Type=oneshot
+ExecStart=/usr/sbin/eos-prune-printers
+StandardOutput=journal
+
+[Install]
+WantedBy=multi-user.target
--- a/Makefile.am
+++ b/Makefile.am
@@ -23,6 +23,7 @@ dist_systemdunit_DATA = \
 	eos-firstboot.service \
 	eos-image-boot-dm-setup.service \
 	eos-live-boot-overlayfs-setup.service \
+	eos-prune-printers.service \
 	eos-split-flatpak-repo.service \
 	eos-transient-setup.service \
 	eos-update-flatpak-repos.service \
@@ -99,6 +100,7 @@ dist_sbin_SCRIPTS = \
 	eos-firstboot \
 	eos-image-boot-dm-setup \
 	eos-live-boot-overlayfs-setup \
+	eos-prune-printers \
 	eos-repartition-mbr \
 	eos-split-flatpak-repo \
 	eos-transient-setup \


eos-prune-printers
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env sh
+
+# eos-prune-printers: Purge all installed printers and corresponding drivers
+# Copyright (C) 2021 Endless OS Foundation LLC
+# Licensed under the GPLv2
```
wjt, line 5: I know all the other files in this repo just have this line, but it *should* be:

```suggestion
# This program is free software; you can redistribute it and/or
# modify it under the terms of the GNU General Public License as
# published by the Free Software Foundation; either version 2 of the
# License, or (at your option) any later version.
#
# This program is distributed in the hope that it will be useful, but
# WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
# General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with this program; if not, see <http://www.gnu.org/licenses/>.
```

wjt (changes requested):
I have not reviewed the branch as a whole

eos-prune-printers.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Endless prune printers
+After=cups.service
+ConditionPathExists=!/var/eos-printers-pruned
```
dbnicholson, line 4: Let's make this friendlier to typical path usage and namespaced to eos4 in case we need to do this again for the eos4 -> eos5 transition.

How about `/var/lib/eos4-printers-pruned` (with the script adjusted, too)?

eos-prune-printers
```diff
@@ -0,0 +1,27 @@
...
+[ -e $stamp_file ] && exit 0
+
+# Purge and cancel all print jobs
+cancel -xa
+
+# Delete all printers with CUPS tool
+for printer in `lpstat -p | awk '{print $2}'`
+do
+	echo Deleting $printer
+	lpadmin -x $printer
```
dbnicholson, line 19: I think you should quote the variable since we don't know what's in it (spaces, shell special characters, etc.).

tmpfiles.d/eos-prune-printers.conf
```diff
@@ -0,0 +1,6 @@
+# remove stamp file from eos-config-printer
+# (from 4.0 use driverless printers)
+R /var/lib/eos-config-printer
+
+# clean up the residual print cache files
+R /var/cache/cups/*
```
dbnicholson, line 6: Wait, this will wipe the CUPS cache files on every boot. I think you want to only do this once in the script. Really both could probably be done in the script. `eos-config-printer` will never run again since it's not included in the OS anymore, so a one time cleanup is probably all that's needed.

eos-prune-printers.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Endless prune printers
+After=cups.service
```
dbnicholson, line 3: Since this depends on CUPS running (or `lpstat` and `lpadmin` will fail), I think adding `Requires=cups.service` is a good idea.

eos-prune-printers.service
```diff
@@ -0,0 +1,13 @@
...
+ConditionPathExists=!/var/eos-printers-pruned
+ConditionKernelCommandLine=!endless.live_boot
+
+[Service]
+Type=oneshot
+ExecStart=/usr/sbin/eos-prune-printers
+StandardOutput=journal+console
```
dbnicholson, line 10: Since this isn't an early boot service, I think just the default of `journal` is fine.

dbnicholson (changes requested):
A couple small suggestions, but overall this looks like it should work.

eos-prune-printers
starnight: sure

eos-prune-printers.service
starnight: Sure

eos-prune-printers.service
starnight: yap

eos-prune-printers
starnight: Sure

eos-prune-printers.service
starnight: okay

tmpfiles.d/eos-prune-printers.conf
starnight: I make `tmpfiles.d/eos-prune-printers.conf ` installed in [`/usr/share/eos-boot-helper/tmpfiles.d/`](https://github.com/endlessm/eos-boot-helper/blob/895fe92e2967ff08d85ae106f4702d0c77b67f77/Makefile.am#L127) originally.  Will this be taken by systemd-tmpfiles on every boot?

> I think you want to only do this once in the script.
Do you suggest that just the simple `rm` to replace the `systemd-tmpfiles --remove ...` in the script?

starnight: Okay!  After quick test, I believe the the simple `rm` works as well.

eos-prune-printers
```diff
@@ -0,0 +1,45 @@
...
+# Exit if the stamp file exists. The service file checks for this, but also
+# guard against an accidental execution.
+stamp_file=/var/lib/eos4-prune-printers
+[ -e $stamp_file ] && exit 0
+
+# Purge and cancel all print jobs
+cancel -xa
+
+# Delete all printers with CUPS tool
+for printer in `lpstat -p | awk '{print $2}'`
```
wjt, line 29: I ran [shellcheck](https://www.shellcheck.net/), an excellent shell script linter, over this script. It has only one suggestion:

```
In eos-prune-printers line 29:
for printer in `lpstat -p | awk '{print $2}'`
               ^----------------------------^ SC2006: Use $(...) notation instead of legacy backticked `...`.

Did you mean: 
for printer in $(lpstat -p | awk '{print $2}')

For more information:
  https://www.shellcheck.net/wiki/SC2006 -- Use $(...) notation instead of le...
```

```suggestion
for printer in $(lpstat -p | awk '{print $2}')
```

wjt (changes requested):

eos-prune-printers
starnight: Wow!  The shell script linter looks so cool!!!
Fixed!

eos-prune-printers
```diff
@@ -0,0 +1,45 @@
...
+# Exit if the stamp file exists. The service file checks for this, but also
+# guard against an accidental execution.
+stamp_file=/var/lib/eos4-prune-printers
+[ -e $stamp_file ] && exit 0
+
+# Purge and cancel all print jobs
+cancel -xa
+
+# Delete all printers with CUPS tool
+for printer in $(lpstat -p | awk '{print $2}')
```
dbnicholson, line 29: Sorry, one more nitpicky thing I noticed when I had a look at the eos2 -> eos3 upgrade script. I don't know if anything else actually comes out of `lpstat`, but there it's matching only lines that start with `printer`. That seems like a safer Idea, so I think using `awk '/^printer/{print $2}'` is slightly safer.

dbnicholson (changes requested):

eos-prune-printers
starnight: sure

dbnicholson (approved):
:+1:

PR #346, opened 2021-06-11
Use repo locking when splitting flatpak repo

dbnicholson:
Wire up the OSTree repo locking now that the API is available. That immediately deadlocked because the `.lock` file was being hardlinked to the new repo, so the hardlinking is reduced to just the objects and (likely nonexistent) deltas. Everything else is copied.

https://phabricator.endlessm.com/T32090


pwithnall (approved):
LGTM

PR #347, opened 2021-06-14
Add eos-test-mode

GeorgesStavracas:
This small script is imported from GNOME Initial Setup. The
Polkit action, however, was kept there, to preserve this nice
property of eos-boot-helper only providing scripts, and the
policies around these scripts are kept in where they're used.

https://phabricator.endlessm.com/T32045


Makefile.am
```diff
@@ -128,3 +129,14 @@ dist_extratmpfiles_DATA = \
 ldconfdir = $(sysconfdir)/ld.so.conf.d
 dist_ldconf_DATA = \
 	ld.so.conf.d/flatpak-extension-runtime-fdo-openh264.conf
+
+com.endlessm.TestMode.policy: com.endlessm.TestMode.policy.in
+	$(AM_V_GEN) $(SED) -e 's|[@]sbindir@|$(sbindir)|g' < "$<" > "$@"
+
+policydir = $(POLKIT_POLICYDIR)
+dist_policy_DATA = \
```
dbnicholson, line 137: Since the policy file is generated at build time, it shouldn't be distributed. Otherwise it will have the `sbindir` path from one the tarball was made. In other words, just `policy_DATA`.

dbnicholson (changes requested):
One nitpick but otherwise looks good.

Makefile.am
GeorgesStavracas: Good catch. Fixed.

dbnicholson (approved):
Jenkins likes you now

PR #348, opened 2021-06-16
eos-prune-printers: Fix the service's condition & requirements

starnight:
1. Fix the conflicted cups.service requirement by removing it. Because,
   the script needs to stop it temporarily.
2. Sync the service's condition guard stamp filename with the script's.
3. Check the directories exist before remove them and their content.

https://phabricator.endlessm.com/T31862

### DIFF
--- a/eos-prune-printers
+++ b/eos-prune-printers
@@ -36,9 +36,22 @@ done
 systemctl stop cups
 
 # remove stamp file from eos-config-printer (from 4.0 use driverless printers)
-rm -rf /var/lib/eos-config-printer
+LIB_CONFIG_PRINTER=/var/lib/eos-config-printer
+if [ -d $LIB_CONFIG_PRINTER ]; then
+	echo Removing $LIB_CONFIG_PRINTER
+	rm -rf $LIB_CONFIG_PRINTER
+else
+	echo $LIB_CONFIG_PRINTER does not exist
+fi
+
 # clean up the residual print cache files
-rm -rf /var/cache/cups/*
+CACHE_CUPS=/var/cache/cups
+if [ -d $CACHE_CUPS ]; then
+	echo Removing "$CACHE_CUPS/*"
+	rm -rf ${CACHE_CUPS:?}/*
+else
+	echo $CACHE_CUPS does not exist
+fi
 
 systemctl start cups
 
--- a/eos-prune-printers.service
+++ b/eos-prune-printers.service
@@ -1,8 +1,8 @@
 [Unit]
 Description=Endless prune printers
 After=cups.service
-Requires=cups.service
-ConditionPathExists=!/var/lib/eos4-printers-pruned
+Wants=cups.service
+ConditionPathExists=!/var/lib/eos4-prune-printers
 ConditionKernelCommandLine=!endless.live_boot
 
 [Service]


eos-prune-printers
```diff
@@ -36,9 +36,22 @@ done
...
+if [ -d $LIB_CONFIG_PRINTER ]; then
+	echo Removing $LIB_CONFIG_PRINTER
+	rm -rf $LIB_CONFIG_PRINTER
+else
+	echo $LIB_CONFIG_PRINTER does not exist
+fi
+
 # clean up the residual print cache files
-rm -rf /var/cache/cups/*
+CAHCE_CUPS=/var/cache/cups
```
wjt, line 48:
```suggestion
CACHE_CUPS=/var/cache/cups
```

eos-prune-printers
```diff
@@ -36,9 +36,22 @@ done
...
+if [ -d $CAHCE_CUPS ]; then
+	echo Removing $CAHCE_CUPS/*
```
wjt, line 50: What do you hope that this will print? Do you hope for:

```
Removing /var/cache/cups/*
```

or do you hope for

```
Removing /var/cache/cups/foo /var/cache/cups/bar [… and so on for each file in that directory]
```
?

eos-prune-printers
starnight: UH!  Thanks find out the typo

eos-prune-printers
```diff
@@ -36,9 +36,22 @@ done
...
+	rm -rf $CAHCE_CUPS/*
```
wjt, line 51:
```
$ shellcheck eos-prune-printers
In eos-prune-printers line 51:
	rm -rf $CAHCE_CUPS/*
               ^-----------^ SC2115: Use "${var:?}" to ensure this never expands to /* .

For more information:
  https://www.shellcheck.net/wiki/SC2115 -- Use "${var:?}" to ensure this nev...
```

wjt (changes requested):

eos-prune-printers
wjt: (As written, it will output the former if `/var/cache/cups` is empty, and the latter if it contains one or more files. This is related to the shellcheck warning below.)

starnight: I'd like simple `/var/cache/cups/*`

eos-prune-printers
starnight: sure

eos-prune-printers
```diff
@@ -36,9 +36,22 @@ done
...
+	echo Removing $LIB_CONFIG_PRINTER
+	rm -rf $LIB_CONFIG_PRINTER
+else
+	echo $LIB_CONFIG_PRINTER does not exist
+fi
+
 # clean up the residual print cache files
-rm -rf /var/cache/cups/*
+CACHE_CUPS=/var/cache/cups
+if [ -d $CAHCE_CUPS ]; then
```
wjt, line 49: I don't think you tested this new version :-)

```suggestion
if [ -d $CACHE_CUPS ]; then
```

eos-prune-printers
```diff
@@ -36,9 +36,22 @@ done
...
+	echo Removing ${CACHE_CUPS:?}/*
```
wjt, line 50: You followed the shellcheck warning but I don't think this does what you hope for.

```
       ${parameter:?word}
              Display Error if Null or Unset.  If parameter is null or unset, the expansion of word (or a message to that effect
              if  word  is not present) is written to the standard error and the shell, if it is not interactive, exits.  Other‐
              wise, the value of parameter is substituted.
```

CACHE_CUPS is always set so I don't think this or the line below do what you hope.

```
wjt@camille:~$ mkdir /tmp/foo
wjt@camille:~$ X=/tmp/foo
wjt@camille:~$ echo Removing $X/*
Removing /tmp/foo/*
wjt@camille:~$ echo Removing ${X:?}/*
Removing /tmp/foo/*
wjt@camille:~$ touch /tmp/foo/{a,b,c}
wjt@camille:~$ echo Removing $X/*                                                                                               
Removing /tmp/foo/a /tmp/foo/b /tmp/foo/c
wjt@camille:~$ echo Removing ${X:?}/*                                                                                               
Removing /tmp/foo/a /tmp/foo/b /tmp/foo/c
```

wjt (changes requested):

eos-prune-printers
starnight: Seems I missed some meaning in the description.  Thanks

eos-prune-printers
starnight: I tested the new script.  But, it is an obvious error here.  Maybe I edited it accidentally when I did the triple check.  :/
Going to fix it again.  Thanks.

wjt (approved):
LGTM but I haven't really thought about the systemd interaction here – @jprvita any thoughts?

eos-prune-printers.service
```diff
@@ -1,8 +1,7 @@
 [Unit]
 Description=Endless prune printers
 After=cups.service
-Requires=cups.service
```
dbnicholson, line 4: Instead of removing this entirely, can we just change it to `Wants`? That will still pull in cups so that `lpadmin` works, but it won't actually stop this service when cups is stopped.

dbnicholson (changes requested):

eos-prune-printers.service
starnight: Tested with Wants.  It works.
And, place the reference https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Wants=

dbnicholson (approved):

PR #349, opened 2021-07-09
Remove merged flatpak/ostree repo overlay setup

dbnicholson:
Except for an EOS 3 upgrade gone wrong, the system flatpak repo should now be split from the system ostree repo. In that case we don't need any special handling for that configuration. Remove that and fix some other minor bugs while here.

I tested this in an installed VM and it worked correctly, but I haven't tested the live boot setup yet.

https://phabricator.endlessm.com/T32011


dbnicholson:
After testing it, I also folded in the change to not remount `/run` with `user_xattr` since it's no longer needed.

jprvita (approved):

PR #350, opened 2021-07-09
Remove split disk support

dbnicholson:
We're dropping support for split disks in EOS 4, so we don't need any of this anymore. While here, fix a distcheck issue.

This is likely to conflict with #349 in eos-test-mode.

https://phabricator.endlessm.com/T32109


jprvita (approved):
Changes look good, feel free to merge this once you fix the conflicts.

GeorgesStavracas (approved):
LGTM as well

PR #352, opened 2021-08-24
eos-vm-generator: Enable virtualbox-guest-utils via generator

jprvita:
We need to enable virtualbox-guest-utils dynamically via a generator,
since ConditionVirtualization=oracle on the unit is not enought because
it will prevent conflicted units (systemd-timesyncd.service) from
starting.

https://phabricator.endlessm.com/T32330

### DIFF
--- a/eos-vm-generator
+++ b/eos-vm-generator
@@ -1,0 +1,20 @@
+#!/bin/sh
+#
+# Usage: eos-vm-generator normal-dir [...]
+#
+# Conditionally adds additional boot dependencies for vm boots.
+# This script implements systemd.generator(7).
+
+dest_dir="${1:?normal-dir argument missing}"
+system_dir=/usr/lib/systemd/system
+multi_user_target_wants_dir="$dest_dir/multi-user.target.wants"
+
+# Only run on virtualbox
+vm_string=$(/usr/bin/systemd-detect-virt)
+if [ "$vm_string" != "oracle" ] ; then
+  exit 0
+fi
+
+unit_path="$system_dir/virtualbox-guest-utils.service"
+mkdir -p "$multi_user_target_wants_dir"
+ln -sf "$unit_path" "$multi_user_target_wants_dir/"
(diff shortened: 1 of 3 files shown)


eos-vm-generator
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+#
+# Usage: eos-live-boot-generator normal-dir [...]
```
wjt, line 3: copypasta error, should be `eos-vm-generator`

wjt (changes requested):
Looks fine except for a typo. It seems quite sad to have to do this. Is it worth one or more of:

- a bug report to Debian
- a bug report to systemd

?

eos-vm-generator
jprvita: Yep, I thought I had changed it but maybe I accidentally reverted it back. Good eye!

jprvita:
> Looks fine except for a typo. It seems quite sad to have to do this. Is it worth one or more of:
> 
>     * a bug report to Debian
> 
>     * a bug report to systemd
> 
> 
> ?

Well, I don't think there is really any bug to report. I read one comment about a similar situation arguing that this is really a bug in the configuration, since we were trying to enable conflicting units. While that (enabling conflicting units) is not really what we are trying to achieve, that is how we are expressing it. So there is no way for us to express what we want via static configuration, but generators give us a way to dynamically define our configuration in runtime. I agree it is a bit less readable than having it all expressed in the unit files, but maybe this new approach is more aligned with what we actually want: enable virtualbox-guest-utils.service when we are running on virtualbox (but I guess it can be argued both ways).

It might be worth writing a "case study" email and asking if someone has ideas on how to improve it (either in our config or enhancements in systemd itself).

jprvita:
New push to address a `shellcheck` suggestion:
```
In eos-vm-generator line 14:
if [ $vm_string != "oracle" ] ; then
     ^-- SC2086: Double quote to prevent globbing and word splitting.
```

wjt:
Does Debian's systemd package enable `systemd-timesyncd.service` by default? Does its `virtualbox-guest-utils` package enable `virtualbox-guest-utils.service` by default? If so there is a bug in Debian, no?

PR #355, opened 2021-09-23
Remove dead code from eos-transient-setup

wjt:
This script has many vestiges of a time when demo mode existed, but that time is long behind us. Remove the dead code and command-line option. Spotted while reviewing #354.

Then while I was here I decided to run `pylint` over the file and fix most of the warnings. We are left with:

```
************* Module eos-transient-setup
eos-transient-setup:1:0: C0103: Module name "eos-transient-setup" doesn't conform to snake_case naming style (invalid-name)
eos-transient-setup:1:0: C0114: Missing module docstring (missing-module-docstring)
eos-transient-setup:59:0: C0115: Missing class docstring (missing-class-docstring)
eos-transient-setup:161:15: W0703: Catching too general exception Exception (broad-except)
```

Legit (except the first one) but ones for another day.

I haven't tested it "for real" but I have tested that `--dry-run` succeeds and that running without `--dry-run` as my own user fails because it can't write to `/var`, and that `flake8` is still happy.

### DIFF
--- a/eos-transient-setup
+++ b/eos-transient-setup
@@ -17,7 +17,6 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 import argparse
-import enum
 import glob
 import json
 import logging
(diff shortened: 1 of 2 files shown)


pwithnall (approved):
:100:

pwithnall:
I tested this on a real system at https://phabricator.endlessm.com/T32408#909228 and it seems fine.

PR #361, opened 2021-12-03
Add stats metric for splitting flatpak repo

dbnicholson:
The flatpak repo splitting has been really slow for many people, so record some statistics and send them in a new metric. Suggestions on the event name welcome.

https://phabricator.endlessm.com/T32790

### DIFF
--- a/eos-split-flatpak-repo
+++ b/eos-split-flatpak-repo
@@ -43,6 +43,7 @@ from gi.repository import EosMetrics  # noqa: E402
 logger = logging.getLogger(os.path.basename(__file__))
 
 EOS_UPDATER_METRIC_FAILURE = "927d0f61-4890-4912-a513-b2cb0205908f"
+SPLIT_FLATPAK_REPO_STATS = "880fd091-2e68-4bd2-baa1-f6810fabcc1c"
 
 
 class SplitError(Exception):
@@ -515,20 +516,36 @@ def main():
 
     logging.basicConfig(level=args.log_level)
 
+    recorder = EosMetrics.EventRecorder.get_default()
+
     try:
         with SplitStats() as stats:
             split_repo(prune=args.prune, stats=stats)
     except Exception as e:
-        recorder = EosMetrics.EventRecorder.get_default()
         recorder.record_event_sync(
             EOS_UPDATER_METRIC_FAILURE,
             GLib.Variant("(ss)", ("eos-split-flatpak-repo", str(e))),
         )
         raise
 
-    # If anything changed, show the statistics.
+    # If anything changed, show the statistics and send metrics.
     if stats.changed:
         stats.log_results()
+        recorder.record_event_sync(
+            SPLIT_FLATPAK_REPO_STATS,
+            GLib.Variant('(uuuuuuutut)', (
+                int(stats.elapsed),
+                stats.num_os_refs,
+                stats.num_flatpak_refs,
+                stats.num_other_refs,
+                stats.num_objects,
+                stats.num_deltas,
+                stats.num_apps,
+                stats.size_apps,
+                stats.num_runtimes,
+                stats.size_runtimes,
+            ))
+        )
 
 
 if __name__ == '__main__':
--- a/eos-split-flatpak-repo
+++ b/eos-split-flatpak-repo
@@ -29,6 +29,10 @@ import os
 import shutil
 import subprocess
 import tempfile
+import time
+
+gi.require_version('Flatpak', '1.0')
+from gi.repository import Flatpak  # noqa: E402
 
 gi.require_version('OSTree', '1.0')
 from gi.repository import OSTree  # noqa: E402
@@ -43,6 +47,53 @@ EOS_UPDATER_METRIC_FAILURE = "927d0f61-4890-4912-a513-b2cb0205908f"
 
 class SplitError(Exception):
     pass
+
+
+class SplitStats:
+    """Statistics for repo splitting"""
+    def __init__(self):
+        self.changed = False
+        self.num_os_refs = 0
+        self.num_flatpak_refs = 0
+        self.num_other_refs = 0
+        self.num_objects = 0
+        self.num_deltas = 0
+        self.num_apps = 0
+        self.size_apps = 0
+        self.num_runtimes = 0
+        self.size_runtimes = 0
+
+        self._start_time = 0
+        self._end_time = 0
+
+    @property
+    def elapsed(self):
+        return self._end_time - self._start_time
+
+    def start_timing(self):
+        self._start_time = time.monotonic()
+
+    def end_timing(self):
+        self._end_time = time.monotonic()
+
+    def log_results(self):
+        logger.info('Time elapsed: %.1f seconds', self.elapsed)
+        logger.info('Number of OS refs: %i', self.num_os_refs)
+        logger.info('Number of Flatpak refs: %i', self.num_flatpak_refs)
+        logger.info('Number of other refs: %i', self.num_other_refs)
+        logger.info('Number of objects: %i', self.num_objects)
+        logger.info('Number of deltas: %i', self.num_deltas)
+        logger.info('Number of Flatpak apps: %i', self.num_apps)
+        logger.info('Size of Flatpak apps: %i', self.size_apps)
+        logger.info('Number of Flatpak runtimes: %i', self.num_runtimes)
+        logger.info('Size of Flatpak runtimes: %i', self.size_runtimes)
+
+    def __enter__(self):
+        self.start_timing()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.end_timing()
 
 
 class RepoType(enum.Enum):
@@ -190,7 +241,7 @@ def get_refspec_repo_type(refspec, os_remotes, flatpak_remotes):
         return None
 
 
-def gather_refs(repo, os_remotes, flatpak_remotes):
+def gather_refs(repo, stats, os_remotes, flatpak_remotes):
     """Gather refs into OS and Flatpak lists"""
     # Put the remotes into sets for get_refspec_repo_type and ensure
     # they're disjoint.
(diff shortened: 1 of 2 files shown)


eos-split-flatpak-repo
```diff
@@ -219,6 +257,20 @@ def gather_refs(repo, os_remotes, flatpak_remotes):
...
 
+def record_repo_stats(repo, stats):
+    """Record statistics for repository data"""
+    logger.info('Recording repository %s stats', repo.path)
+
+    _, objects = repo.list_objects(
+        OSTree.RepoListObjectsFlags.ALL |
+        OSTree.RepoListObjectsFlags.NO_PARENTS
+    )
+    stats.num_objects = len(objects)
```
dbnicholson, line 268: Just to be up front about it, getting the number of objects here is a bit expensive. OSTree enumerates all the object files and then serializes each name into a GVariant. I'm not sure we really need this information, so I just want to be clear it's not free.

wjt: I timed this on my Mission One (5400 rpm disk):

```
>>> x = time.monotonic(); print(len(repo.list_objects( OSTree.RepoListObjectsFlags.ALL | OSTree.RepoListObjectsFlags.NO_PARENTS )[1])); y = time.monotonic(); print(x, y, y-x)
327813
832.414858406 900.593506899 68.17864849299997
```

(This is after the split had completed but I think before the repo has been pruned.)

So that's 68 seconds to enumerate all the objects. I waited a few minutes (during which time the computer was otherwise unused but I could hear its disk buzzing continuously, looks like it was the updater) and re-ran the same invocation, 66 seconds.

Maybe there would be some cache-warming effect where the subsequent "hardlink all the things" operation will be slightly quicker, but still I fear this will actually make the problem we are trying to quantify and solve worse! I'd leave it out.

eos-split-flatpak-repo
```diff
@@ -205,6 +242,7 @@ def gather_refs(repo, os_remotes, flatpak_remotes):
     flatpak_refs = []
     other_refs = []
     _, all_refs = repo.list_refs_ext(None, OSTree.RepoListRefsExtFlags.NONE)
+    stats.num_refs = len(all_refs)
```
wjt, line 245: Perhaps we can break this down and report the number of OS refs, flatpak refs, and other refs?

If we were feeling fancy we could also report the number of Flatpak apps installed – but I suppose you can't just assume that flatpak refs = 2 * flatpak apps (normal ref and deploy/ ref) because apps can have 1 or more extensions.

eos-split-flatpak-repo
wjt: OK, here's an idea: Run `cp` with `--verbose` and count the number of lines of output. Of course this includes directories too but on my system there are 514 directories and 928,349 non-directories across the ostree & flatpak repos so it's a reasonable approximation. Of course this would also come with a performance cost…

eos-split-flatpak-repo
```diff
@@ -403,16 +460,31 @@ def main():
...
+    recorder = EosMetrics.EventRecorder.get_default()
+
     try:
-        split_repo(prune=args.prune)
+        with SplitStats() as stats:
+            split_repo(prune=args.prune, stats=stats)
     except Exception as e:
-        recorder = EosMetrics.EventRecorder.get_default()
         recorder.record_event_sync(
             EOS_UPDATER_METRIC_FAILURE,
```
wjt, line 470: I was thinking, hey, we should report a event on error, too. But we do! But… there are no events in this table with component `eos-split-flatpak-repo`. Again I am suspicious of https://phabricator.endlessm.com/T31191…

dbnicholson: Yes, I see what you mean. This service does start early. Certainly before the event recorder and without any strict ordering w.r.t. to system dbus socket.

eos-split-flatpak-repo
dbnicholson: Thanks for testing, that's too painful.

The simple way to do this with counting lines from `--verbose` would be to run with `stdout=PIPE` and then split it. However, that would buffer all of the output into memory, which could be a lot. Another way would be to write to a file then open and read lines in a more controlled way. Or use `stdout.PIPE` without `communicate`. Or `cp -v | wc -l` with `shell=True`, ensuring that shell was `bash` and you were checking the return value of `cp` correctly. I found https://stackoverflow.com/questions/845058/how-to-get-line-count-of-a-large-file-cheaply-in-python, which was fascinating. On my system none of the python approaches even got close to `wc`.

dbnicholson: Yet another option is `find repo/objects -type f` or some python equivalent. It's slightly cheaper than `ostree_repo_list_objects` because the names aren't being serialized into variants, but you still to enumerate all the paths. However, this is a little faster because it avoids `stat`ing all the objects since we don't care about that:

```
def listcount(repo):
    objsdir = os.path.join(repo, 'objects')
    nobjs = 0
    for entry in os.scandir(objsdir):
        if not entry.is_dir():
            continue
        nobjs += len(os.listdir(entry.path))
    return nobjs
```

This is basically using the knowledge of how `objects` is broken up into a two level content addressed storage. It was faster than `glob`, `walk` and `find | wc -l` on my system.

dbnicholson: Yeah, wow, `ostree_repo_list_objects` is like 500x slower than my fastest `listdir` approach. From https://phabricator.endlessm.com/P868:

```
$ ./countobjs.py /var/lib/flatpak/repo
globcount: 308969: 0.5490682479576208
walkcount: 308961: 0.2801525180111639
listdircount: 308969: 0.17220832197926939
listobjscount: 308815: 5.436755720991641
findcount: 300130: 0.46448501001577824
```

Would be useful if you could try that out on your spinning disk. I have a feeling it would still be faster than having `cp` write a ton of output to stdout and then read it all. Or maybe we just scrap all of this as too expensive...

eos-split-flatpak-repo
dbnicholson: Sure, we could do that. I can also just open the `Flatpak.Installation` on get the info as interpreted by flatpak itself. That would also allow us to see the number of pulled but not installed refs.

eos-split-flatpak-repo
dbnicholson: My debugging in https://phabricator.endlessm.com/T31191 shows that it's an ordering cycle to try to pull in `dbus.socket`. Since in it's current form this service has `eos-updater-flatpak-installer.service`, then it effectively has the same ordering and can't rely on sending the metrics over DBus.

One option is to reorder it after `eos-updater-flatpak-installer.service` and any of the other early boot ostree/flatpak services. The other (I believe) is to write the statistics/errors to disk and have eos-metrics-instrumentation pick them up when it gets started.

dbnicholson: I opened #362 for the service ordering since I'm haven't gotten to addressing the actual metrics comments.

eos-split-flatpak-repo
wjt: OK, I'll test this tonight/tomorrow but sadly I'm miles from my Mission One today and can't SSH to it… :)

dbnicholson: I tracked down my Mission spinning disk, so I'll try it out myself today.

dbnicholson: I tested this out and the `listdir` method took less than 1 second while the `list_objects` method took 54 seconds. I think I'll go with the `listdir` method...

dbnicholson:
Ready again. I don't think the metrics collection is going to be too much of a slowdown, but I haven't tested this on a real system yet.

tests/test_split_flatpak_repo.py
```diff
@@ -471,11 +560,11 @@ def test_duplicate_repo(self):
         for root, dirs, files in os.walk(dup_path):
             repo_dir = os.path.relpath(root, dup_path).split(os.sep)[0]
             if repo_dir in ('objects', 'deltas'):
-                expected_links = 2
+                test = self.assertGreater
```
wjt, line 569: In what cases are the expected_links greater than 2? Is it roughly, if part of a Flatpak, then 3 (1 in system repo, 1 in flatpak repo, 1 deployed) and otherwise, 2?

tests/test_split_flatpak_repo.py
```diff
@@ -221,18 +243,52 @@ def create_flatpak_commits(self):
         flatpak_refs.update(self.random_commit(
             self.os_repo,
             ('flathub:app/org.gnome.Totem/x86_64/stable',
-             'deploy/app/org.gnome.Totem/x86_64/stable')
+             'deploy/app/org.gnome.Totem/x86_64/stable'),
+            flatpak_metadata=dedent("""\
```
wjt, line 247: I always forget this trailing-slash syntax and instead stick a `.strip()` on the end. Nice.

eos-split-flatpak-repo
```diff
@@ -379,7 +494,7 @@ def split_repo(prune=False, root='/'):
     if not os.path.islink(flatpak_dir_path):
         logger.info('Flatpak dir %s is not a symlink', flatpak_dir_path)
     else:
-        changed = True
+        stats.changed = True
```
wjt, line 497: Hmm. So in the very unlikely case that the block above succeeds and then the script is interrupted or fails in this block, we would end up not submitting the stats on that run, and on a subsequent run we would submit a stub event with most fields being `0`. I think that's OK though.

eos-split-flatpak-repo
```diff
@@ -216,9 +268,57 @@ def gather_refs(repo, os_remotes, flatpak_remotes):
...
+def get_num_repo_objects(repo):
+    """Get the count of objects in a repository
+
+    Normally you'd use OSTree.Repo.list_objects, but that's too slow as
+    it tries each potential object subdirectory and then serializes each
+    found object name into a GVariant.
+
+    All we care about is the count, so this function uses knowledge of
+    the 2 level content addressed object storage and just looks at
+    directory entries.
```
wjt, line 287: This is a nice solution!

wjt (approved):
LGTM. Please merge as-is if you choose to handle the uint64 point mentioned on https://github.com/endlessm/azafea/pull/172/ in azafea (which I think is fine).

tests/test_split_flatpak_repo.py
dbnicholson: Exactly. I thought about trying to keep this exact, but really I just want to make sure some things are hard linked and others aren't. I don't think the exact count matters that much.

eos-split-flatpak-repo
dbnicholson: Yes, this was really bugging me but I don't know exactly what to do about it. I guess you could treat this part as a separate result or just leave it out of the results completely.

wjt: I think "do nothing" because the chances of hitting this path are so tiny

wjt: (and if I'm wrong about that we'll know from the data – it will be easy to distinguish)

PR #362, opened 2021-12-07
split-flatpak-repo: Run service in background

dbnicholson:
Since `eos-update-flatpak-repos.service` and
`eos-updater-flatpak-installer.service` block boot via
`systemd-update-done.service`, ordering before them means that
`eos-split-flatpak-repos.service` blocks boot for a potentially very
long time.

Instead, order after those services and before any others that open the
system OSTree repo or the system Flatpak installation. By also reverting
to `DefaultDependencies=yes`, this service can run in the background
while the desktop is up. Of course, any system OSTree or Flatpak
operations will fail, but the system is otherwise usable.

This also allows the metrics to work since otherwise the service runs
befor the system DBus socket exists.

https://phabricator.endlessm.com/T32768

### DIFF
--- a/eos-split-flatpak-repo.service
+++ b/eos-split-flatpak-repo.service
@@ -1,20 +1,23 @@
 [Unit]
 Description=Split Flatpak repo from system OSTree repo
-DefaultDependencies=no
-Conflicts=shutdown.target
-Wants=local-fs.target
-After=local-fs.target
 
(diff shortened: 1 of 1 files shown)


wjt (approved):
LGTM. I did wonder whether we should ask systemd to retry this service but:

a. It'll be retried on the next startup
b. Thanks to this change we'll get metrics events if/when it fails.

PR #365, opened 2022-02-11
eos-user-preset: Remove user presets to disable pipewire service

GeorgesStavracas:
We're enabling PipeWire for audio now, therefore it is nice not
to disable the core daemon!

This effectively reverts commit 3f7ccdef69a86ccb757dcfdfe1def16fafc9992e.

https://phabricator.endlessm.com/T33042

### DIFF
--- a/50-eos-user.preset
+++ b/50-eos-user.preset
@@ -1,5 +1,0 @@
-# Systemd user presets for EOS.
-
-# Disable pipewire service by default - the service can still be activated
-# via socket activation or manually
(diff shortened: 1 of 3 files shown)


jprvita:
Do we actually need to do this? Pipewire can still be started via socket activation, even if its unit is disabled.

GeorgesStavracas:
Converting to a draft for now - let's see how it goes without this change first, and re-evaluate if necessary

PR #366, opened 2022-04-21
Remove Endless Key hooks

wjt:
The first version of Endless Key had:

- Windows and macOS apps to consume Kolibri & Kiwix content
- Kolibri & Kiwix content
- A bootable Endless OS image

We used magic to make the Kolibri & Kiwix content from the exFAT
partition be used by the Kolibri and Encyclopedia apps in Endless OS,
respectively.

We no longer include an Endless OS image on Endless Key; and the
encyclopedia content has been moved so even if we did still have an OS
image on the device, that part would be broken.

Remove these hooks from the OS.

I did look at reverting the patches, but there were surprisingly many of them and I think it is easier to read this way.

https://phabricator.endlessm.com/T33159


jprvita (approved):
There are two  parts being removed here, one is the generated unit files and tmpfile that setup Kolibri + Encyclopedia to use the data stored on USB; the other is the generated mount unit that makes the USB storage from a live USB available to the OS. 

AFAIR the latter is only used by the former, but would it make sense to keep it around for future use? Probably not, as we can easily bring it back if needed, but I just wanted to raise the question before we nuke it away.

Other than that, LGTM, so feel free to merge it if you agree there is no benefit in keeping the storage setup.

wjt:
Fair question – in fact I think it is useful because it is a partial fix for https://phabricator.endlessm.com/T22735 so I will remove that part of this patch.

jprvita:
> Fair question – in fact I think it is useful because it is a partial fix for https://phabricator.endlessm.com/T22735 so I will remove that part of this patch.

Oh right, good memory! I wasn't sure from the top of my head if that partition was named `eoslive` or `eosimages`.

LGTM now!

wjt:
> Oh right, good memory! I wasn't sure from the top of my head if that partition was named `eoslive` or `eosimages`.

Well… it's the former for live images and the latter for eosinstaller/"mass reformatter" images.

PR #367, opened 2022-09-02
nvidia-graphics-setup: Make egl-wayland use PRIME render offload

starnight:
The egl-wayland failed to choose the NVIDIA device for rendering and
chooses the internal GPU as the fallback since 1.1.10. This patch sets
the environment variable __NV_PRIME_RENDER_OFFLOAD to tell egl-wayland
choosing the NVIDIA GPU as the EGL device for rendering as past when the
nvidia driver is loaded for the NVIDIA device.

https://phabricator.endlessm.com/T33814


starnight:
We have found the better way for the dual GPU usage

PR #373, opened 2022-11-21
Revert "eos-user-preset: Add user presets to disable pipewire service"

wjt:
This reverts commit 3f7ccdef69a86ccb757dcfdfe1def16fafc9992e.

Pipewire is now the sound server on Endless OS.

https://phabricator.endlessm.com/T34142

### DIFF
--- a/50-eos-user.preset
+++ b/50-eos-user.preset
@@ -1,5 +1,0 @@
-# Systemd user presets for EOS.
-
-# Disable pipewire service by default - the service can still be activated
-# via socket activation or manually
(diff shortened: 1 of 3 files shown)


wjt:
Actually I don't think we should do this; I think we should blocklist *more* services, and rely on socket activation!

jprvita:
> Actually I don't think we should do this; I think we should blocklist _more_ services, and rely on socket activation!

Yes, we want sounds servers (and maybe other services too) to be started on demand via socket activation.

PR #374, opened 2022-11-21
Start more user services via socket/D-Bus activation

wjt:
This reduces the number of things that get launched when the desktop
session is coming up, and avoids some unnecessary services being run in
some contexts.

https://phabricator.endlessm.com/T34144

### DIFF
--- a/50-eos-user.preset
+++ b/50-eos-user.preset
@@ -1,5 +1,15 @@
 # Systemd user presets for EOS.
 
-# Disable pipewire service by default - the service can still be activated
-# via socket activation or manually
+# Start this SSH agent on-demand via socket activation. (It is believed that
+# this socket is never actually used because gnome-keyring provides its own
+# SSH_AUTH_SOCK.)
+disable gcr-ssh-agent.service
+
+# Start Pipewire (and its PulseAudio emulation) on-demand via socket activation.
+disable pipewire-pulse.service
 disable pipewire.service
+
+# Start Tracker components on-demand via D-Bus activation. In particular, this
+# avoids running Tracker in the GDM greeter or Initial Setup session.
+disable tracker-extract-3.service
+disable tracker-miner-fs-3.service


50-eos-user.preset
```diff
@@ -1,5 +1,15 @@
...
+disable gcr-ssh-agent.service
+
+# Start Pipewire (and its PulseAudio emulation) on-demand via socket activation.
+disable pipewire-pulse.service
 disable pipewire.service
+
+# Start Tracker components on-demand via D-Bus activation. In particular, this
+# avoids running Tracker in the GDM greeter or Initial Setup session.
+disable tracker-extract-3.service
+disable tracker-miner-fs-3.service
```
dbnicholson, line 15: In principle I agree with this, but the miner is something that benefits from running in the background all the time, IMO. If it doesn't start doing anything until someone clicks in a search box, then the results will be pretty useless. Is there anything else that might trigger it earlier on real user sessions?

wjt: I will do some further testing having dropped this file in locally… I guess I can run a dbus monitor that watches for the first attempt to talk to tracker and logs it to the journal!

jprvita: It would be nice to not have Tracker running in the greeter session + started automatically in the user session, as that combination usually means two instances running at the same time when the user session gets started [1], or whenever the user presses ctrl+alt+f1 to go back to the greeter session. So if there is no other early trigger in the real user session, perhaps we should implement one and still disable the unit by default.

[1] in general, greeter units are not stopped by the time the user units start -- we had some problems with the user's PulseAudio trying to register profile support with BlueZ, because the greeter one had registered first and was still around.

jprvita (approved):
LGTM. Left a comment in the conversation about Tracker.

wjt:
This change is not sufficient to avoid `tracker-extract-3.service` being run in GDM & Initial Setup because the ostree includes a symlink at /etc/systemd/user/default.target.wants/tracker-extract-3.service. Same for `pipewire-pulse.service`.

50-eos-user.preset
wjt: > If it doesn't start doing anything until someone clicks in a search box, then the results will be pretty useless. Is there anything else that might trigger it earlier on real user sessions?

Well, on Endless it also gets explicitly enabled at package install time, so even with this PR it will get eagerly launched by systemd. :)

Testing on my Fedora machine, you have to actually start a search in Nautilus, or in Shell (where one of the enabled search providers uses it). There are probably other things that will D-Bus-activate it but I haven't dug any further. So in that context, yes, it seems that Tracker won't start indexing your files in any given session until you first do something to cause an app to talk to Tracker. I see your point that the results will not reflect any files created/changed/deleted between the start of that session and some time after Tracker gets started & reindexes the files.

PR #377, opened 2023-01-16
Update system CA store when /etc is updated

starnight:
Users, or Endless may add some CAs which do not belong debian packages. Those CAs will be updated into /etc/ssl/certs/ca-certificates.crt and would be likely to be mismatched from the shipped version in /usr/etc.

Add a helper service that runs update-ca-certificates after update and boot to sync up when it is needed.

https://phabricator.endlessm.com/T34308

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -28,6 +28,7 @@ dist_systemdunit_DATA = \
 	eos-prune-printers.service \
 	eos-transient-setup.service \
 	eos-update-flatpak-repos.service \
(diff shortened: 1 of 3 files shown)


dbnicholson (changes requested):
This looks good, but can you add `ca-certificates` to `Depends` in `debian/control` to ensure the script is available?

starnight:
sure!  make sense.

dbnicholson (approved):

PR #379, opened 2023-02-14
eos-update-flatpak-repos: Migrate eos-runtimes to new URL

dbnicholson:
Historically the legacy runtimes were committed to the OS repo because they were built from the same packages and therefore were entirely deduplicated on the server. Fast forward many years and now the mixture of ostree and flatpak commits is starting to cause problems as flatpak is managing the repo metadata differently than flatpak.

The legacy runtimes have been duplicated into a new eos-runtimes repo that will contain only flatpaks. In order to use it, we need to migrate users to the new URL.

https://phabricator.endlessm.com/T33445

### DIFF
--- a/eos-update-flatpak-repos
+++ b/eos-update-flatpak-repos
@@ -130,6 +130,42 @@ def _remove_remotes():
                                    '--system', '--force', name])
 
 
+# dict of remote -> URLs and collection ID to migrate. If a collection
+# ID is not needed, None should be used.
+REMOTES_TO_MIGRATE = {
+    "eos-runtimes": {
+        "old-url": "https://ostree.endlessm.com/ostree/eos",
+        "new-url": "https://ostree.endlessm.com/ostree/eos-runtimes",
+        "new-collection-id": "com.endlessm.Runtimes",
+    },
+}
+
+
+def _migrate_remotes():
+    insts = Flatpak.get_system_installations()
+    inst = insts[0]
+
+    for name, config in REMOTES_TO_MIGRATE.items():
+        remote = _flatpak_inst_get_remote(inst, name)
+        if remote and remote.get_url() == config["old-url"]:
+            logging.info("Migrating remote {} from {} to {} ({})"
+                         .format(name, config["old-url"], config["new-url"],
+                                 config["new-collection-id"]))
+
+            remote.set_url(config["new-url"])
+            remote.set_collection_id(config["new-collection-id"])
+            inst.modify_remote(remote)
+
+            # Try to update the metadata from the remote configuration,
+            # but ignore failures in case we're offline.
+            logging.info("Updating remote metadata for {}".format(name))
+            try:
+                inst.update_remote_sync(name)
+            except GLib.Error as err:
+                logging.warning("Could not update remote metadata for {}: {}"
+                                .format(name, err))
+
+
 # dict of remote -> list of runtimes
 RUNTIMES_TO_REMOVE = {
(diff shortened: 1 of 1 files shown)


eos-update-flatpak-repos
```diff
@@ -130,6 +130,33 @@ def _remove_remotes():
...
+        if remote and remote.get_url() == urls["old-url"]:
+            logging.info("Migrating remote {} from {} to {}"
+                         .format(name, urls["old-url"], urls["new-url"]))
+
+            subprocess.check_call(["flatpak", "remote-modify",
+                                   "--system", "--url", urls["new-url"],
+                                   name])
+            subprocess.check_call(["flatpak", "remote-modify",
+                                   "--system", "--update-metadata",
+                                   name])
```
wjt, line 157: Does this require a network connection? What happens if the computer is offline?

eos-update-flatpak-repos
```diff
@@ -130,6 +130,33 @@ def _remove_remotes():
...
+    insts = Flatpak.get_system_installations()
+    inst = insts[0]
+
+    for name, urls in REMOTES_TO_MIGRATE.items():
+        remote = _flatpak_inst_get_remote(inst, name)
+        if remote and remote.get_url() == urls["old-url"]:
+            logging.info("Migrating remote {} from {} to {}"
+                         .format(name, urls["old-url"], urls["new-url"]))
+
+            subprocess.check_call(["flatpak", "remote-modify",
```
wjt, line 152: Why use the CLI? `_fix_remote_urls` (lower down this file) uses the libflatpak API to do something rather similar.

wjt:
It's a shame we never managed to switch to declarative repo configuration… https://phabricator.endlessm.com/T23145

eos-update-flatpak-repos
dbnicholson: Fine by me. Most of the other code uses the CLI, but I prefer the API.

eos-update-flatpak-repos
dbnicholson: Good catch. It does fetch the remote repo config and syncs to it. The important thing you'd lose is not getting the collection ID, so I think I should update the configuration dict so that can be specified here. The only other part would be the title. It appears that flatpak transaction does that automatically, so you'd probably get it later.

eos-update-flatpak-repos
```diff
@@ -130,6 +130,42 @@ def _remove_remotes():
...
 
+# dict of remote -> URLs and collection ID to migrate. If a collection
+# ID is not needed, None should be used.
+REMOTES_TO_MIGRATE = {
+    "eos-runtimes": {
+        "old-url": "https://ostree.endlessm.com/ostree/eos",
+        "new-url": "https://ostree.endlessm.com/ostree/eos-runtimes",
+        "new-collection-id": "com.endlessm.Runtimes",
+    },
+}
```
dbnicholson, line 141: It would be kind of nice to also specify `old-collection-id` and include that in the match so you could migrate the URL and/or the collection ID, but I didn't need that right now.

dbnicholson:
> It's a shame we never managed to switch to declarative repo configuration… https://phabricator.endlessm.com/T23145

True. Every time I've thought about that I feel like there's a whole level of "can I update this because it's out of date or because the user modified it locally?" that would be hard to handle. Not to mention the complex migration of existing systems. Maybe someday, but Philip's list is accurate and daunting for something that (IMO) is not particularly critical.

PR #381, opened 2023-03-16
dracut: Disable hostonly to install more modules into initramfs

starnight:
As an Linux distribution, Endless OS tries to support as much machines as possible. So, install more modules into initramfs generated by dracut to support more storage/block devices during the boot time by disabling "hostonly".

If anyone needs to generate a host specific initramfs, they can pass the "--hostonly" option to dracut.

https://phabricator.endlessm.com/T34575


dbnicholson (approved):
Looks good!

PR #382, opened 2023-03-18
ESP mount generator

dbnicholson:
This adds a systemd generator to mount the ESP. The intention is to supplant the ESP mounting currently handled by systemd-gpt-auto-generator since it does not contain the policy to handle all the systems EOS runs on. I've tested this (in a VM) on the following configurations:

* Standard x86_64 EOS with GPT partitioning
* Standard x86_64 EOS with MBR partitioning (using `eos-repartition-mbr` from the original GPT)
* PAYG x86_64

I have not tested this on:

* Windows dual boot
* arm64 EOS

As we previously discussed, I'm sure this will get the dual boot case wrong. However, with this generator we can add the policy and tests for it. Trying to cram the policy into systemd-gpt-auto-generator would probably never work.

There's pretty extensive testing since I really don't want to regress any systems. You can test this on a real device by dropping it into `/etc/systemd/system-generators/`. Without the `/efi` directory it only does anything useful on PAYG. You can do a full test by creating a temporary deployment with an `/efi` directory:

```
sudo mkdir -p /etc/systemd/system-generators
sudo install -m755 eos-esp-generator /etc/systemd/system-generators/
# Adjust the refspec as shown in ostree admin status
eos_refspec=eos:os/eos/amd64/master
sudo mkdir esproot
sudo mkdir -m700 esproot/efi
sudo ostree commit -b esp -s "ESP testing" --tree=ref=$eos_refspec --tree=dir=esproot
sudo ostree deploy esp
reboot
```

When done, redeploy the normal ostree:
```
sudo ostree deploy "$eos_refspec"
sudo rm /etc/systemd/system-generators/eos-esp-generator
```

https://phabricator.endlessm.com/T29930


tests/run-tests
```diff
@@ -2,8 +2,4 @@
 set -e
 
 cd "$(dirname "$0")"
-if pytest=$(which py.test-3); then
-    exec "$pytest" "$@"
-else
-    exec python3 -m unittest discover --top-level-directory .. "$@"
-fi
+exec pytest "$@"
```
wjt, line 5: I think you need to spell this as [`pytest-3` or `py.test-3`](https://packages.debian.org/bullseye/all/python3-pytest/filelist):

```
../../../tests/run-tests: line 5: exec: pytest: not found
```

dbnicholson: Ugh, I thought I checked that. I switched it to `python3 -m pytest` since that will work universally.

dbnicholson:
OK, so this is nearly a rewrite of the generator that was primarily driven by the need to handle the Windows dual boot case. The significant changes are:

* For dual boot, look for the `endless.image.device` kernel command line argument and try to use the ESP from there.
* Use `blkid` for gathering partition data instead of `sfdisk`. I think `sfdisk` could still work, but the nested partition within disk JSON format made that awkward. I really wanted to use `lsblk`, but that depends on `udev`, which is blocked from starting until the generators complete.
* While trying to debug issues, I found out that generator stdout/stderr goes nowhere. As noted in [systemd.generator(7)](https://www.freedesktop.org/software/systemd/man/latest/systemd.generator.html#Notes%20about%20writing%20generators), the only way to get output is to write to `/dev/kmsg`. So, I wrote a logging handler to do that, which seems to work nicely.
* Added a "gather mode" that I used to get test data.

I tested this on 4 different systems that are represented in the test data:

* Regular EOS with shim and GPT partitions
* Regular EOS with shim and MBR partitions
* PAYG with sd-boot
* Windows dual boot with shim

Earlier I wrote some instructions on how to test this by making a temporary ostree commit, but you can do it even easier. Since directories in the checkout are made on the fly, you can just make the `/efi` directory after temporarily making `/` mutable:

```
chattr -i /
mkdir /efi
chattr +i /
mkdir -p /etc/systemd/system-generators
install eos-esp-generator /etc/systemd/system-generators/
```

The `/efi` directory will persist until a new ostree deployment is made.

eos-esp-generator
```diff
@@ -0,0 +1,968 @@
...
+            # kmsg recognizes a syslog style <priority>identifier[pid]: prefix.
+            priority = self.PRIORITY_MAP.get(record.levelno, SysLogHandler.LOG_INFO)
+            prefix = f'<{priority:d}>{progname}[{self.pid:d}]: '
+
+            # kmsg allows a maximum of 1024 bytes per message, so split the
+            # message up if needed.
+            size = 1024 - len(prefix)
+            for pos in range(0, len(message), size):
+                buf = f'{prefix}{message[pos:pos + size]}'.encode('utf-8')
+                self.kmsg.write(buf)
```
wjt, line 118: 1024 bytes or 1024 characters (which is >= 1024 bytes when encoded as UTF-8)? I assume it's bytes. In that case this will not do the right thing with non-ASCII characters. Consider the message:

```
"☃" * 512
```

Asssuming `prefix` is less than 512 characters, this code will assume that just a single chunk is needed. But:

```
>>> len(("☃" * 512).encode("utf-8"))
1536
```

Unfortunately encoding as UTF-8 before naïvely chunking is also wrong because you can't split the UTF-8-encoded bytes at any arbitrary byte.

I think you need to do something like this:

```python
            message = self.format(record).encode('utf-8')

            # kmsg recognizes a syslog style <priority>identifier[pid]: prefix.
            priority = self.PRIORITY_MAP.get(record.levelno, SysLogHandler.LOG_INFO)
            prefix = f'<{priority:d}>{progname}[{self.pid:d}]: '.encode('utf-8')

            # kmsg allows a maximum of 1024 bytes per message, so split the
            # message up if needed, without splitting multibyte-encoded characters
            start = 0
            length = len(message)
            size = 1024 - len(prefix)
            while start < length:
                end = start + size
                while end < length and (message[end] & 0b1100_0000) == 0b1000_0000:
                    end -= 1

                buf = prefix + message[start:end]
                self.kmsg.write(buf)
```

(The UTF-8 encoding is such that if a byte is of the form `10xxxxxx` it is a continuation byte.)

eos-esp-generator
```diff
@@ -0,0 +1,968 @@
...
+    If the EFI variable doesn't exist, None is returned. Any nul
+    terminating bytes will be removed.
+    """
+    value = read_efivar(name)
+    if value is None:
+        return None
+
+    logger.debug(f'EFI variable {name} contents: {value}')
+
+    # Systemd appends 3 nul bytes for some reason. If there are an odd
```
wjt, line 293: I'm curious to know what the reason is!

(From a less careful codebase I might wonder if this is the UTF-16-encoding of U+0000 (two nul bytes) followed by a C byte string terminator (1 nul byte).)

tests/test_esp_generator.py
```diff
@@ -0,0 +1,414 @@
...
+    # Windows dual boot. The Endless disk (including an ESP partition) is a
+    # file on the NTFS partition that gets loop mounted. The real disk's ESP
+    # should be mounted instead of the loop disk's ESP partition. The real disk
+    # is found from the endless.image.device kernel command line parameter.
+    'windows': espgen.EspMount(
+        source='/dev/sda1',
+        target='/efi',
+        type='vfat',
+        umask='0077',
+    ),
```
wjt, line 201: There are 2 Windows dual-boot cases: efi and bios. In the bios case we don't expect to find an mbr. Perhaps we should rename the test case here as `windows-efi` so that in future some brave explorer can find a BIOS windows system in a cave, install EOS on it and check that in that case the generator doesn't break.

tests/test_esp_generator.py
```diff
@@ -0,0 +1,414 @@
...
+        term = b'\0' * n
+        with open(varpath, 'wb') as f:
+            f.write(attr)
+            f.write('hello'.encode(utf_16))
+            f.write(term)
+        value = espgen.read_efivar_utf16_string(var)
+        assert value == 'hello'
+
+    # Unicode.
+    uface = '\N{UPSIDE-DOWN FACE}'  # 🙃
```
wjt, line 407: 🙃

eos-esp-generator
dbnicholson: Indeed you are correct. As Aqua once said, I'm an ASCII girl, in an ASCII world. I never understood why UTF-8 was encoded that way, but now I get it! Your suggestion makes perfect sense. Thanks!

eos-esp-generator
```diff
@@ -0,0 +1,968 @@
...
+            if err.returncode == 2:
+                logger.warning(f'Ignoring blkid --probe exit code 2 for {dev}')
+                continue
+            else:
+                raise
+
+        entry = {}
+        for line in proc.stdout.splitlines():
+            tag, _, value = line.partition('=')
+            if not tag:
```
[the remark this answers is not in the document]
dbnicholson: It's unlikely but better safe than sorry. Usually the way I like to do this is check if the separator field is empty. Will fix.

eos-esp-generator
dbnicholson: I found the spot in the systemd code where it happens, but it was quite a while ago. There weren't really comments, but I think it was sort of what you're suggesting. Whether you try to parse as a C string or UTF-16, you'll find a trailing NUL. But it just has easily could have been a mistake that's essentially irrelevant the way systemd reads the values. The Unicode handling was entirely bespoke. I'll try to find it again when I get a chance.

tests/test_esp_generator.py
dbnicholson: Good point, I agree with the rename. I think I was trying to support the idea that the ESP should be mounted whether you're booted with UEFI or not. However, thinking about it again, I don't think that's a great idea. Especially since part of the method depends on reading EFI variables, which obviously won't be there. I think I'll add a short circuit if you're not in UEFI mode.

wjt: On dual-boot systems (unlike standalone ones) we only ever install the corresponding bootloader for the firmware type, so if a dual-boot system is booted in BIOS mode then it is certain that there is no EFI bootloader.

In theory I guess it's possible that a standalone install may be booted in CSM/legacy BIOS mode one day and real EFI mode another day if someone is doing weird stuff with the firmware settings. In that case I think it's still fine to short-circuit if booted in BIOS mode and not mount the ESP (which ultimately we're doing in order to update the bootloader) even if there is one. If the system is later booted in EFI mode we'll have a chance to update it then; and if not then it's a moot point whether the EFI bootloaders are out of date.

eos-esp-generator
wjt: It's a really clever encoding: the best thing to come out of Plan-9!

tests/test_esp_generator.py
dbnicholson: Did the rename and added a `/sys/firmware/efi` check (which is also how `systemd-gpt-auto-generator` decides if you're EFI booted or not).

eos-esp-generator
dbnicholson: You algorithm seems to work correctly, although it was missing incrementing `start`. In order to prove it, I added a couple tests. I also added some explicit flushes like `logging.FileHandler` and changed the UTF-8 encoding to use `errors='replace'` so it doesn't bomb on malformed messages.

dbnicholson:
I added the new changes as fixups so it would be easier to see the differences from the previous mega-commit. I'll squash them if you approve.

wjt (approved):
Looks great. Squash away!

PR #385, opened 2023-04-12
Add service to ensure systemd-boot installed

dbnicholson:
On systemd-boot systems, the bootloader is updated with `bootctl update` on every boot. Unfortunately, it doesn't currently work because `bootctl` considers systemd-boot to be installed if `/EFI/systemd` is non-empty in the ESP. The image builder only populates `/EFI/BOOT`, so systemd-boot is not considered installed.

If `bootctl` does not consider systemd-boot to be installed, `eos-ensure-sd-boot` will check the `LoaderInfo` EFI variable to see if the current bootloader is systemd-boot. If so, it populates `/EFI/systemd` so that the next time `bootctl update` runs, it will update the bootloader.

https://phabricator.endlessm.com/T34703


dbnicholson:
I tested this on both grub and systemd-boot VMs. For the grub system it correctly did nothing. For the systemd-boot VM, it correctly populated `/EFI/systemd`. That allowed `bootctl update` to run and update the EFI binaries. Unfortunately, it also tried to update `/boot/loader`, and that doesn't exist since we have some hacks to use a fake symlink `/boot/loader.sln`.

Marking this as a draft for now. One solution I can think of is to change systemd downstream so that `bootctl` only tries to install or update the EFI binaries and doesn't touch the rest of `/boot`. We've been getting along fine without any of the things it wants to put in `/boot/loader.

PR #387, opened 2023-11-30
transient-setup: Only add reformatter to favourites

wjt:
Previously, on live sessions, com.endlessm.Installer.desktop was both
prepended to the favourites, and to the app grid. This made sense in
Endless OS 4 and older, because an app could be shown in both places.
However, starting from Endless OS 5 (or rather, GNOME 40) an app that is
in the favourites is not shown in the grid. Prepending it to the grid is
not harmless because the logic that reads the grid JSON still reserves
space for it, leaving an empty space on the first page of the grid.

Stop prepending it to the grid.

https://phabricator.endlessm.com/T35104

### DIFF
--- a/eos-transient-setup
+++ b/eos-transient-setup
@@ -32,10 +32,6 @@ from gi.repository import GLib, Gio, OSTree  # noqa: E402
 
 log = logging.getLogger(sys.argv[0])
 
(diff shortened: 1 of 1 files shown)


dbnicholson (approved):
Seems reasonable.

PR #391, opened 2024-01-08
T35041 add user manager config

dylanmccall:
To achieve this, we will copy the same configuration file that is added
to system.conf.d.

This change is necessary to prevent systemd from stopping the user
session when a child process is killed by the OOM killer. It should be
possible to revert this once Endless OS is using systemd >= v253, where
OOMPolicy defaults to "continue" for login session scopes:
https://github.com/systemd/systemd/commit/98b6c94b577205d31b019286c2a84cc9af244ea0

https://phabricator.endlessm.com/T35041

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -56,6 +56,10 @@ dist_systemdsystemmanagerconf_DATA = \
 	10-eos-oom-policy.conf \
 	$(NULL)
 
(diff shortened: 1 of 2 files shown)


starnight (approved):
Thanks!

PR #393, opened 2024-03-01
Add suspend-mode-setup to force system use preferred suspend mode

starnight:
Add suspend-mode-setup.service executing suspend-mode-setup when system boots up. It forces system use preferred suspend mode. For example, GIGABYTE Mission One should use s2idle.

Since Mission One is the only one in the list, put it into the array s2idle_model_list in suspend-mode-setup script directly now. It can be split into another data file, if there are more models in the future.

https://phabricator.endlessm.com/T35070#988997


starnight:
Use another fix

PR #394, opened 2024-03-01
udev: Force Mission one use s2idle suspend mode

starnight:
Mission one cannot resume from S3 suspend by any key press, nor other outside trigger ways. Because the USB is dead. Use s2idle mode to avoid turn off USB when suspend.

https://phabricator.endlessm.com/T35070

### DIFF
--- a/60-suspend-mode.rules
+++ b/60-suspend-mode.rules
@@ -1,0 +1,2 @@
+SUBSYSTEM=="dmi", ENV{ID_VENDOR}="GIGABYTE", ENV{ID_MODEL}="Mission one", RUN{program}+="/bin/sh -c 'echo s2idle > /sys/power/mem_sleep'"
+SUBSYSTEM=="dmi", ENV{ID_MODEL}="GB-BXBT-2807", RUN{program}+="/bin/sh -c 'echo s2idle > /sys/power/mem_sleep'"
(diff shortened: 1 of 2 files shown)


starnight:
I forgot install it.

60-suspend-mode.rules
```diff
@@ -0,0 +1 @@
+SUBSYSTEM=="dmi", ENV{ID_VENDOR}="GIGABYTE", ENV{ID_MODEL}="Mission one", RUN{program}+="/bin/sh -c 'echo s2idle > /sys/power/mem_sleep'"
```
wjt, line 1: From https://phabricator.endlessm.com/w/hardware/endless/mission-one/ I would have guessed that it is `GB-BXBT-2807` but it's true that https://github.com/endlessm/systemd/commit/1ecc9fe237a750a65249b948f4b28834a6c1fc16 uses `Mission one` (as well as `GB-BXBT-2807`).

I updated https://metabase.endlessm.com/question/267-pings-from-mission-one to look at both names, and `GB-BXBT-2807` way outnumbers `Mission one`. I think we should treat both the same, just as the old patch does.

wjt (changes requested):

60-suspend-mode.rules
starnight: Do they have the same sys_vendor "GIGABYTE"?

starnight: Pushed a new commit.  Ready for review again.

PR #398, opened 2024-04-12
image-boot: use loopback mounts for all filesystems

dsd:
https://phabricator.endlessm.com/T35126


dracut/image-boot/eos-image-boot-setup
```diff
@@ -55,6 +54,9 @@ if [ $? != 0 ]; then
   echo "image-boot: failed to detect filesystem type"
   exit 1
 fi
+if [ "${fstype}" = "ntfs" ]; then
+  fstype=ntfs3
+fi
 
 # $squashfs is used both as a suffix for the device-mapper device (if needed),
 # and as a flag indicating that we're dealing with a squashfs.
```
wjt, line 56: This comment is no longer true: `$squashfs` is now only tested for emptiness versus non-emptiness. I suggest updating the comment and also adjusting the value to e.g. `yes` or `true` or `1`

wjt (approved):
LGTM except for a now-misleading comment and associated potential simplification/clarification.

PR #400, opened 2024-04-16
More psi-monitor info

dbnicholson:
Get some more information out of `psi-monitor` so we can see what's happening.

https://phabricator.endlessm.com/T35318


dsd:
Could we also have a cmdline arg to adjust the 10% default threshold? Would be handy for experimenting with other values.

dbnicholson:
> Could we also have a cmdline arg to adjust the 10% default threshold? Would be handy for experimenting with other values.

Indeed, that's half the reason I added the option parsing, but then I forgot to do it!

PR #402, opened 2024-04-18
psi-monitor: Raise default memory pressure threshold to 30%

starnight:
Current PSI monitor's default memory pressure threshold is 10%. However, according to the application usage, the memory pressure (full avg10) exceeds the threshold 10% easily, then PSI monitor triggers OOM killer to kill a process for taking memory back. But, there is still some available memory space at the moment. That means it is too sensitive on some users' side.

Thanks to the statistics of application usage and the memory pressure measured by Daniel. This commit raises the memory pressure threshold to 30% for trying more widely. Hope it get to the balance more closely.

Besides, add a comment about the memory pressure interface provided by kernel.

https://phabricator.endlessm.com/T35318

### DIFF
--- a/psi-monitor/psi-monitor.c
+++ b/psi-monitor/psi-monitor.c
@@ -15,9 +15,14 @@
 /* Daemon parameters */
 static unsigned int poll_interval = 5;
 static unsigned int recovery_interval = 15;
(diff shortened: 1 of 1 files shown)


dsd:
merged my own version at 40%. let's try!

PR #408, opened 2024-06-04
Run eos-firstboot a second time on dual-boot systems

wjt:
In new dual-boot installs of 6.0.0, eos-firstboot did not correctly grow the filesystem due to the bug fixed in commit
56aea35e497df9d035e8ccf391d9b720de264530.

We now want to run it a second time, only on dual-boot systems. Ideally we would like to only trigger this on dual-boot systems that were affected by this bug but it's harder to detect that case than to run eos-firstboot again and have resize2fs gracefully do nothing.

When booting a non-live image-booted system which has not previously run this unit, remove the flag file (if present) that prevents eos-firstboot.service from running more than once, then touch another flag file to prevent running this unit again. The service is ordered Before=eos-firstboot.service and is Type=oneshot so we are guaranteed that its ExecStart= line has completed before eos-firstboot.service's ConditionPathExists check is evaluated. From systemd.unit(5):

> The conditions and asserts are checked at the time the queued start
> job is to be executed. The ordering dependencies are still respected,
> so other units are still pulled in and ordered as if this unit was
> successfully activated, and the conditions and asserts are executed
> the precise moment the unit would normally start and thus can validate
> system state after the units ordered before completed initialization.

On a fresh installation of an image containing this unit, /var/eos-booted does not exist; so this unit has no effect but is harmless.

On a system originally installed with 6.0.0, eos-firstboot.service will be retriggered once and will successfully resize the filesystem to fill the partition.

On a system originally installed with a release prior to 6.0.0, eos-firstboot.service will be retriggered once and will successfully take no action.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,6 +25,7 @@ dist_systemdunit_DATA = \
 	eos-live-boot-overlayfs-setup.service \
 	eos-ostree-bootloader-setup.service \
 	eos-prune-printers.service \
+	eos-retrigger-firstboot.service \
 	eos-transient-setup.service \
 	eos-update-flatpak-repos.service \
 	eos-update-system-ca.service \
--- a/eos-retrigger-firstboot.service
+++ b/eos-retrigger-firstboot.service
@@ -1,0 +1,19 @@
+# On dual-boot systems, (re)run eos-firstboot to grow the root filesystem, which
+# was not grown correctly on first boot into Endless OS 6.0.0.
+[Unit]
+Description=Fix dual-boot filesystem size
+DefaultDependencies=no
+After=sysinit.target local-fs.target
+Before=basic.target eos-firstboot.service
+ConditionPathExists=!/var/%N
+ConditionKernelCommandLine=endless.image.device
+ConditionKernelCommandLine=!endless.live_boot
+
+[Service]
+Type=oneshot
+ExecStart=rm -f /var/eos-booted
+ExecStartPost=touch /var/%N
+StandardOutput=journal+console
+
+[Install]
+WantedBy=basic.target


Makefile.am
```diff
@@ -25,6 +25,7 @@ dist_systemdunit_DATA = \
 	eos-live-boot-overlayfs-setup.service \
 	eos-ostree-bootloader-setup.service \
 	eos-prune-printers.service \
+	eos-secondboot.service \
```
wjt, line 28: This is a draft only because I don't like the name of the unit!

wjt: Suggestions welcome for a better name that is accurate and not incredibly long...

eos-retrigger-firstboot.service
```diff
@@ -0,0 +1,19 @@
+# On dual-boot systems, (re)run eos-firstboot to grow the root filesystem, which
+# was not grown correctly on first boot into Endless OS 6.0.0.
+[Unit]
+Description=Fix dual-boot filesystem size
+DefaultDependencies=no
+After=sysinit.target local-fs.target
+Before=basic.target eos-firstboot.service
+ConditionPathExists=!/var/%N
```
dbnicholson, line 8: I think Rob has previously expressed regret about dropping these files directly in `/var`. I.e., maybe this should be `/var/lib/%N`.

dbnicholson (commented):
If we're going to unconditionally re-run `eos-firstboot.service`, why not just change the stamp file it uses? I.e., `s,/var/eos-booted,/var/lib/%N,` in `eos-firstboot` and `eos-firstboot.service`? You could even have it delete the old `/var/eos-booted`, which is a bit of an eyesore.

eos-retrigger-firstboot.service
wjt: > If we're going to unconditionally re-run eos-firstboot.service, why not just change the stamp file it uses?

Because I want to only re-run it on dual-boot systems.

But yes, that would be easier. It's harmless to rerun (except in terms of time wasted). It would also be an excuse to clean up the ugly top-level-of-/var stamp file. Nice idea.

wjt:
Converting this to draft, I'll rework it tomorrow as described above.

eos-retrigger-firstboot.service
dbnicholson: Ah, I missed the implications of `ConditionKernelCommandLine=endless.image.device`.

wjt: I implemented that alternative at https://github.com/endlessm/eos-boot-helper/pull/410.

wjt:
#410 was merged instead.

PR #410, opened 2024-06-04
Run eos-firstboot a second time on existing installs

wjt:
An alternative approach to https://github.com/endlessm/eos-boot-helper/pull/408.

----

In new dual-boot installs of 6.0.0, eos-firstboot did not correctly grow the filesystem due to the bug fixed in commit
56aea35e497df9d035e8ccf391d9b720de264530.

We now want to run it a second time. Ideally we would only trigger it on affected dual-boot systems; but it is harder to detect that case than to run eos-firstboot again and have resize2fs gracefully do nothing.

While we could easily run a unit only on dual-boot systems with:

    ConditionKernelCommandLine=endless.image.device
    ConditionKernelCommandLine=!endless.live_boot

to do this only once requires a second unit with its own new flag file. So rather than introducing a second flag file, let's harmlessly re-run this unit once on all systems, and take the opportunity to move the flag file out of the top-level of /var to /var/lib/eos-firstboot.

### DIFF
--- a/eos-firstboot
+++ b/eos-firstboot
@@ -6,5 +6,5 @@ root_part=$(findmnt -rvnf -o SOURCE /)
 
 resize2fs "${root_part}"
 
-> /var/eos-booted
+> /var/lib/eos-firstboot
 exit 0
--- a/eos-firstboot.service
+++ b/eos-firstboot.service
@@ -5,11 +5,12 @@ Description=Endless Boot Helper
 DefaultDependencies=no
 After=sysinit.target local-fs.target
 Before=basic.target
-ConditionPathExists=!/var/eos-booted
+ConditionPathExists=!/var/lib/%N
 ConditionKernelCommandLine=!endless.live_boot
 
 [Service]
 ExecStart=/usr/sbin/eos-firstboot
+ExecStartPost=-rm -f /var/eos-booted
 StandardOutput=journal+console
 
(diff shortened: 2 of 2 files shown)


wjt:
Draft because untested.

I also have a mild concern that there may exist other devices out there where the filesystem and partition don't match, and so unexpectedly running `resize2fs` may not in fact be harmful. But I'm struggling to think of a non-contrived situation where this would happen & be a problem.

wjt:
I tested this on a dual-boot system in the state where the partition had been resized but not the filesystem; as expected, the job re-ran once and grew the filesystem.

dbnicholson:
> I also have a mild concern that there may exist other devices out there where the filesystem and partition don't match, and so unexpectedly running `resize2fs` may not in fact be harmful. But I'm struggling to think of a non-contrived situation where this would happen & be a problem.

I paused a bit on this. If we do run into a situation where old systems can't run the `resize2fs` sequence again, we can change the script like:

```
need_resize=false
[ -f /var/lib/eos-firstboot ] || need_resize=true
[ -f /var/eos-booted ]] && relevant_dual_boot_setup && need_resize=true
if "$need_resize"; then
    resize2fs "${root_part}"
fi

> /var/lib/eos-firstboot
```

dbnicholson (approved):

PR #411, opened 2024-06-06
repartition: Use partx rather than busybox partprobe

wjt:
Since da7b937b252bfe5776d3d6ef8902d8fd93453bfd, partprobe in the
initramfs is busybox, not the tool from parted.

For unknown reasons, busybox partprobe fails to probe /dev/loop0 after
it has been repartitioned, failing with:

    partprobe: /dev/loop0: Invalid argument

partx --update performs a similar function, except that it works
correctly at this stage of the boot process. Using partx also has the
advantage that running this tool manually from the booted system (where
partprobe is the real thing) will actually run the same code as in the
initramfs.

### DIFF
--- a/dracut/repartition/endless-repartition.sh
+++ b/dracut/repartition/endless-repartition.sh
@@ -201,7 +201,7 @@ udevadm settle
 
 # Loop devices need a prod
 if [ -n "$using_loop" ]; then
-  partprobe $root_disk
+  partx --update --verbose $root_disk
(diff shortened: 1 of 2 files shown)


dbnicholson (approved):
Shocking that a busybox implementation is not an actual replacement for the real thing...

The alternative is to make the busybox module not include partprobe, but that requires changing dracut. This is better. Until busybox ships a half assed partx...

PR #417, opened 2024-10-16
overlayfs-setup: Fallback to read only mount when overlay not supported

dbnicholson:
Overlayfs doesn't support using case insensitive filesystems such as vfat and exfat as a lower directory. Change the mount to read only so changes can't be made there.

In theory this would simply be `mount -o remount,ro`. Unfortunately, if systemd has setup an automount, autofs is not our friend. In that case, the mount options need to be persisted into the mount unit so that a subsequent remounting retains the ro option.

https://phabricator.endlessm.com/T35641

### DIFF
--- a/eos-live-boot-overlayfs-setup
+++ b/eos-live-boot-overlayfs-setup
@@ -88,6 +88,58 @@ setup_and_mount_storage() {
 	mount "${device}" /run/eos-live
 }
 
+setup_ro_mount() {
+    local path=$1
+
+    # If the directory isn't a mount point, make a read only bind mount
+    # on top of the directory and be done.
+    if ! findmnt "$path" >/dev/null; then
+        mount -o bind,ro "$path" "$path"
+        return 0
+    fi
+
+    # In case of an automount, trigger the real mount so its options can
+    # be inspected.
+    ls "$path" >/dev/null
+
+    # If it's already mounted read only, assume we're done.
+    local mounted_options
+    mounted_options=$(findmnt -n -f -d backward -o OPTIONS "$path")
+    [[ $mounted_options =~ (,|^)ro(,|$) ]] && return 0
+
+    # If there's an automount, adjust the mount options in the mount
+    # unit so the read only option persists across mountings.
+    local escaped_path automount_unit
+    escaped_path=$(systemd-escape -p "$path")
+    automount_unit="${escaped_path}.automount"
+    if systemctl -q is-active "$automount_unit"; then
+        # Add a drop-in to append ro to the mount unit's Options.
+        # mount(8) will use the last rw or ro found in the options.
+        local mount_unit mount_options new_mount_options
+        mount_unit="${escaped_path}.mount"
+        mount_options=$(systemctl show -P Options "$mount_unit")
+        new_mount_options="${mount_options:+${mount_options},}ro"
+
+        local dropin_dir="/run/systemd/system/${mount_unit}.d"
+        local dropin_conf="${dropin_dir}/readonly.conf"
+        mkdir -p "$dropin_dir"
+        echo "Changing ${mount_unit} Options from \"${mount_options}\"" \
+            "to \"${new_mount_options}\"" >&2
+        cat > "$dropin_conf" <<EOF
+[Mount]
+Options=${new_mount_options}
+EOF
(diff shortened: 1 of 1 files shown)


eos-live-boot-overlayfs-setup
```diff
@@ -88,6 +88,64 @@ setup_and_mount_storage() {
...
+    # If the directory isn't a mount point, make a read only bind mount
+    # on top of the directory and be done.
+    if ! findmnt "$path" >/dev/null; then
+        mount -o bind,ro "$path" "$path"
+        return 0
+    fi
+
+    # In case of an automount, trigger the real mount so its options can
+    # be inspected.
+    ls "$path" >/dev/null
```
wjt, line 103: Is there a timeout on the automount for `/boot`?

wjt (approved):
LGTM.

I wonder what happens if you have both `rw` and `ro` in the mount options. Which one wins? Is it the last one specified? If so the mount flag change could be simplified to unconditionally appending `,ro`.

eos-live-boot-overlayfs-setup
dbnicholson: Systemd uses 120 seconds (2 minutes). I think that should be more than enough time, but if that's not deemed sufficient, a simple way to keep the mount from going idle is to change to the directory. I.e., `pushd $path; ...; popd` would do it.

An alternative I considered was changing the timeout to 0 so that the automounts never consider idleness. Unfortunately, you'd either have to restart the automount systemd unit after changing its parameters or issue an `ioctl` to the existing autofs mount. In either case you'd have to unmount a potentially mounted real mount first. autofs in all its wisdom uses an open fd on the root directory _of the autofs mount, not the real mount_ to issue `ioctl`s. Unless you've stashed an open fd from before the real mount shadows the autofs mount, there's no way to make an `ioctl` on an autofs mount.

dbnicholson:
> I wonder what happens if you have both `rw` and `ro` in the mount options. Which one wins? Is it the last one specified? If so the mount flag change could be simplified to unconditionally appending `,ro`.

I'm not sure. I think probably last one wins and just appending was my first thought. Lemme try that out...

eos-live-boot-overlayfs-setup
wjt: I think it's fine as-is – realistically the script will not take 2 minutes.

dbnicholson:
> > I wonder what happens if you have both `rw` and `ro` in the mount options. Which one wins? Is it the last one specified? If so the mount flag change could be simplified to unconditionally appending `,ro`.
> 
> I'm not sure. I think probably last one wins and just appending was my first thought. Lemme try that out...

Yeah, that seems to work. I added an override to a test mount unit that set `Options=umask=0077,noauto,rw,ro`. After mounting the options were displayed as `ro,...` with no `rw` anywhere. I don't know if mount(8) or the kernel is doing the canonicalizing. Similary, I tried `mount -o remount,rw,ro` and it similary became `ro,...`. I suspect it's the kernel that does this since it always lists `ro` or `rw` as the first option regardless of what order they were specified.

I'll change that since the mount option editing was definitely my least favorite part of this patch.

dbnicholson:
> I suspect it's the kernel that does this since it always lists `ro` or `rw` as the first option regardless of what order they were specified.

Oh, actually it is mount(8). I ran `strace mount -o umask=0077,rw,ro /dev/vda1 /mnt` and the resulting syscall was `mount("/dev/vda1", "/mnt", "vfat", MS_RDONLY, "umask=0077")`. Since read only is a mount(2) flag, it has to extract it from the options. Running it again with `ro,rw` or neither `ro` or `rw` resulted in `mount("/dev/vda1", "/mnt", "vfat", 0, "umask=0077")`. In other words, `rw` is the default of no flags specified.

dbnicholson:
Updated now to unconditionally append `ro`. I tested in a VM and it seems to work correctly.

eos-live-boot-overlayfs-setup
dbnicholson: It also just needs to survive the subsequent `mounted_options=$(findmnt -n -f -d backward -o OPTIONS "$path")` line. The rest of the function behaves correctly whether the mount has been triggered or not.

PR #418, opened 2024-10-18
Validate shell scripts with Shellcheck

wjt:
Currently all shell scripts found in the tree are at least syntax-checked by `bash -n` or `dash -n` based on their shebang.

But as I reviewed #417 I was suddenly gripped with the desire to run them all through `shellcheck`. This devolved into over-engineering the runner emitting TAP-formatted output and wiring up a TAP parser in Automake.

Having done all this, I think it may be unnecessary, but here's the PR anyway.


starnight:
Tried to build it on OBS.  The shellcheck alarms much more syntax issues.

wjt:
Interesting – it should be the same version of shellcheck and I added some logic to make the shellcheck check not fatal for certain files.

wjt:
Thanks for the OBS test. I don't understand why the offending scripts are not being skipped in OBS.

I think for now I'll pause this and occasionally nudge it forwards by fixing the warnings so that the list of scripts that are expected to fail can be removed.

PR #419, opened 2025-02-05
Partially revert "Remove obsolete cleanup of obsolete stamp files"

wjt:
This reverts part of commit 1338aae7721a5bfddc8b121fb1b761bd779d4313.

Sadly I had not realised that this stamp file was also created by a hook in the image builder. So on every new system installed between 3.5.0 and 6.0.1 inclusive, this stamp file was preinstalled and deleted on first boot; while on all systems installed with 6.0.2 or newer, the stamp file was preinstalled but never removed.

https://github.com/endlessm/eos-image-builder/pull/173 removes the image builder hook, but for the sake of those systems installed with 6.0.2 or newer, reinstate the tmpfiles snippet that deletes the stamp file.


dbnicholson (approved):
:roll_eyes:

PR #422, opened 2025-07-09
Migrate Chrome to Flathub

starnight:
Some notes:
- Install Flathub's Chrome if it is not installed and eos-apps' Chrome is installed
- Some services credentials (e.g. gmail) are lost during migration (requiring re-login)
- The migration always runs if '~/.var/app/com.google.Chrome/.migrated' (created at the end of the migration) doesn't exist and eos-apps' Chrome is installed (`/var/lib/flatpak/app/com.google.Chrome/x86_64/eos3' exists), even if '~/.var/app/com.google.Chrome' exists:
  - This is such that we can update '~/.config/mimeapps.list' to point the new Chrome in case the old Chrome is specified as default there but the user never ran it
  - And also to remove old google-Chrome.desktop shortcut on '~/.local/share/applications' that'd refer to the old Chrome as the executable
  - Note that the actual user profile migration is skipped if '~/.var/app/com.google.Chrome' already exists or if '~/.config/google-chrome' doesn't exist by the time the migration runs

https://phabricator.endlessm.com/T26944


starnight:
The new version commit uses `eos-update-flatpak-repos`, instead of uninstalling eos-apps Chrome & installing Flathub Chrome.

eos-migrate-chrome-profile
```diff
@@ -0,0 +1,133 @@
...
+NEW_DESKTOP_FILE = "com.google.Chrome.desktop"
+
+
+def update_mimeapps_list(path):
+    """Update file associations, which in particular includes x-scheme-handler/http and
+    friends to specify the default web browser.
+
+    We cannot use GLib's own API to query what mime types the old Chromium desktop file
+    is a handler for, because we can't construct a GDesktopAppInfo for it, because its
+    desktop file no longer exists.
```
wjt, line 53: The built-in desktop file has to still exist, because otherwise when the user boots their system, Chrome will disappear until it happens to be upgraded.

That's because `eos-update-flatpak-repos` doesn't actually install the new version from Flathub. It can't do so because it runs early in the boot process, when the network may not be available and where we don't want to block on a large download. All it does is adjust the **already installed app** to appear to have been installed from `flathub`, so that future updates come from that remote.

So you have to handle the case where the old Flatpak is installed (with its deploy file rewritten). The eos-apps version can't be launched directly, only via `/usr/share/applications/google-chrome.desktop`. So that desktop file has to remain on the system. We can only remove it following a checkpoint (perhaps the eos6 → eos7 transition); and even then only if we force `com.google.Chrome` to be updated (if installed) across the upgrade.

And then this script can indeed use the normal GLib mime API.

eos-migrate-chrome-profile
```diff
@@ -0,0 +1,133 @@
...
+
+
+def remove_old_desktop_shortcut(path=DESKTOP_SHORTCUTS_DIR):
+    try:
+        shortcuts = os.listdir(path)
+    except FileNotFoundError:
+        return
+
+    if OLD_DESKTOP_FILE in shortcuts:
+        os.remove(os.path.join(path, OLD_DESKTOP_FILE))
```
wjt, line 96: Why would `google-chrome.desktop` be in `.local/share/applications`? It lives in `/usr/share/applications`.

wjt (changes requested):
This isn't going to work as written, for a number of reasons, outlined below (and discussed further in the 6 years of comments in https://phabricator.endlessm.com/T26944).

I think what will work is the following strategy (developing https://phabricator.endlessm.com/T26944#903813):

1. Update the `eos-apps` version's desktop file to be launchable directly, in such a way that its desktop file is hidden on current Endless OS versions but will become visible in a future release, which you can do using `TryExec=/usr/share/eos-google-chrome-helper/magic-flag-executable`
2. Now in an OS update:
    - Use eos-application-tools to force `eos-apps` `com.google.Chrome` to be updated as part of the OS upgrade if it is installed
    - Put a stub executable at that path
    - Remove `/usr/share/applications/google-chrome.desktop` from the OS (or set it to NoDisplay=true so you can use the normal GLib MIME API to update the MIME associations to `com.google.Chrome.desktop`)
    - Add this migration script that moves the config files but adds a symlink from the old directory to the new
    - And use `eos-update-flatpak-repos` to, on first boot into the new OS version, rewrite the deploy file so that future updates come from Flathub
3. Then in a **later** OS update, after a checkpoint (so we can guarantee that the switcheroo has happened), and also with an `eos-application-tools` entry to ensure the `flathub` version of `com.google.Chrome` has been updated if installed, we can remove all this stuff.

It's a really awkward and messy transition.

eos-migrate-chrome-profile
```diff
@@ -0,0 +1,133 @@
...
+
+    if (
+        os.path.isdir(OLD_CHROME_CONFIG_DIR) and
+        not os.path.isdir(NEW_CHROME_CONFIG_DIR)
+    ):
+        update_old_config_references(
+            os.path.join(OLD_CHROME_CONFIG_DIR,
+                         "WidevineCdm", "latest-component-updated-widevine-cdm"))
+        os.makedirs(os.path.dirname(NEW_CHROME_CONFIG_DIR), exist_ok=True)
+        os.rename(OLD_CHROME_CONFIG_DIR, NEW_CHROME_CONFIG_DIR)
```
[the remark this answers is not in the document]
starnight: The new commit installs Flathub's Chrome and uninstalls eos-apps Chrome.  So, this avoids the symbolic link.

eos-update-flatpak-repos
```diff
@@ -1305,6 +1305,15 @@ FLATPAKS_TO_MIGRATE = [
...
     },
+    # migrate Chrome to flathub (T26944)
+    {
+        'name': 'com.google.Chrome',
+        'kind': Flatpak.RefKind.APP,
+        'old-branch': 'eos3',
+        'new-branch': 'stable',
+        'old-origin': 'eos-apps',
+        'new-origin': 'flathub'
+    },
```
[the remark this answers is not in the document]
starnight: Thanks for the eye.  Sorry for my messy brain.

eos-migrate-chrome-profile
```diff
@@ -0,0 +1,132 @@
...
+
+
+def main():
+    migrated_file = os.path.join(NEW_CHROME_DATA_DIR, ".migrated")
+    if os.path.exists(migrated_file):
+        return
+
+    res = subprocess.run(["flatpak", "info", "-o", "com.google.Chrome//stable"])
+    if res.returncode == 1:
+        res = subprocess.run(["flatpak", "install", "flathub", "com.google.Chrome", "-y"])
```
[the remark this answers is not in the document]
starnight: I had an idea before, but did not do like it now.  Two services:
* Uninstall eos-apps Chrome and install Flathub Chrome before gdm.service and having NetworkManager.service
* Migrate Chrome profile service in gnome-session.target

However, according to the questions, seems like using the two services is a better solution.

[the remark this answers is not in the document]
starnight: Updated in the new commit.

eos-migrate-chrome.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Migrate eos-apps Chrome to Flathub Chrome
```
wjt, line 2: I would call this:


```suggestion
Description=Replace eos-apps Chrome with Flathub Chrome
```

starnight: Thanks!

eos-migrate-chrome-profile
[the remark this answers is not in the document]
wjt: But there is no guarantee that the user will **ever** boot their system and have the network be available prior to reaching GDM. Perhaps they always connect via a Wi-Fi hotspot on their phone, and they are in the habit of doing this:

1. Boot their laptop
2. Log in
3. Oops, I forgot! Turn on the Wi-Fi hotspot on the phone
4. Now the laptop is online, and can receive OS updates, but this script will not have run.

eos-migrate-chrome.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Replace eos-apps Chrome with Flathub Chrome
+
+ConditionPathExists=/var/lib/flatpak/app/com.google.Chrome/x86_64/eos3
+After=network-online.target
+Wants=network-online.target
+Before=gdm.service
```
wjt, line 7: `systemd.special(7)` tells me (emphasis mine):

> **network-online.target**
> Units that strictly require a configured network connection should pull in network-online.target (via a Wants= type dependency) and order themselves after it. This target unit is intended to pull in a service that **delays further execution until the network is sufficiently set up**. What precisely this requires is left to the implementation of the network managing service.

In the case where no network is available during boot, I believe this will cause the boot process to block for 60 seconds; then `NetworkManager-wait-online.service` will time out; then this script will run (and fail); and only then will GDM start.

I don't think it's acceptable to block the boot for a minute waiting for a network that may never come up.

And `ConditionPathExists` doesn't save us here. From `systemd.unit(5)`, in the section *Conditions and Asserts* (emphasis mine):

> The ordering dependencies are still respected, so other units are still pulled in and ordered as if this unit was successfully activated, and the conditions and asserts are executed the precise moment the unit would normally start and thus can validate system state after the units ordered before completed initialization.

I believe this means that even if `/var/lib/flatpak/app/com.google.Chrome/x86_64/eos3` does not exist, the `Wants` and `After` relationship on `network-online.target` will still be respected, and so this change would require the network to come up before GDM regardless of whether the migration needs to be performed.

wjt (changes requested):
We cannot rely on the network being available during the boot process to run this migration. If we are insisting on using an ad-hoc migration script, rather than using some combination of `eos-update-flatpak-repos` + making the `eos-apps` Chrome Flatpak be directly launchable + `eos-updater-flatpak-installer` to do the migration, then we will need to make the migration script be triggered by the network going online; or I guess another option is to have `eos-updater` run it directly. Either way, this means we need to handle the case where the migration runs while a user is logged in and has `eos-apps` Chrome installed.

----

I think it's right to do this in EOS 6, but we will separately need to add logic in `eos-updater` to prevent crossing the checkpoint and upgrading to `EOS 7` if the migration is necessary and has not been completed. This will be similar to some of the previous ad-hoc checks we've used to block upgrades, such as "is nvme-remap in use".

eos-migrate-chrome-profile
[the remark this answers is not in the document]
starnight: Okay, I agree this is a problem now.

eos-migrate-chrome.service
starnight: Uhm, found [For those who use systemd services that rely on a “network” connection](https://forum.manjaro.org/t/for-those-who-use-systemd-services-that-rely-on-a-network-connection/83626).

```
After=network-online.target
Wants=network-online.target
```
Seems waiting on `network-online.target` is not as expected, nor reliable anymore.

So, I check system log and notice when system connects to available internet, NetworkManager shows message: `manager: NetworkManager state is now CONNECTED_GLOBAL`.  Then, I wondering is there any NetworkManager's dbus signal relating to `CONNECTED_GLOBAL`.  However, after checked with DSpy, I did not find any NetworkManager's signal relating to `CONNECTED_GLOBAL`.

wjt: I was sure I had mentioned this but I can't find it in this thread.

Look at /etc/NetworkManager/dispatcher.d/

starnight: Have script `50-eos-migrate-chrome` and link it as `/etc/NetworkManager/dispatcher.d/50-eos-migrate-chrome`.  It will trigger `eos-migrate-chrome.service` when NetworkManager changes state to `CONNECTED_GLOBAL`.

starnight:
Although EOS master's environment is (going to be) based on GNOME OS now and this should be applied to EOS 6, it is still good be in master branch, too.

starnight:
Note: If one tries to build eos-boot-helper locally during this transition time, will hit the issue mentioned by commit [eos-update-efi-uuid: Fix warning treated as error](https://github.com/endlessm/eos-boot-helper/commit/1ac27713c6881fb1858c076271f3a70d2e3efc8d).  Revert it as workaround if the environment is based on old EOS master.

eos-migrate-chrome.service
```diff
@@ -0,0 +1,11 @@
...
+Description=Replace eos-apps Chrome with Flathub Chrome
+ConditionPathExists=/var/lib/flatpak/app/com.google.Chrome/x86_64/eos3
+After=NetworkManager.service
+Wants=NetworkManager.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/eos-boot-helper/eos-migrate-chrome
+Restart=no
+RemainAfterExit=yes
```
wjt, line 11: I can never quite remember the semantics here, but I believe this means that, once the service is started once, then it will always be considered active, and so `systemctl start eos-migrate-chrome.service` will do nothing. If I'm right then I believe this is wrong. Consider the following sequence of events:

1. Computer becomes connected;
2. NM hook starts this service;
3. Computer becomes disconnected while downloading Flathub Chrome;
4. This service exits, but I believe it will still be considered to be running;
5. Now the computer goes offline and online again.

I believe in this case we would want the script to be started again, but due to `RemainAfterExit=yes` it won't be.

eos-migrate-chrome.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Replace eos-apps Chrome with Flathub Chrome
+ConditionPathExists=/var/lib/flatpak/app/com.google.Chrome/x86_64/eos3
+After=NetworkManager.service
+Wants=NetworkManager.service
```
wjt, line 5: I don't think this is necessary.

eos-migrate-chrome-profile.service
```diff
@@ -0,0 +1,15 @@
...
+ConditionPathExists=/var/lib/flatpak/app/com.google.Chrome/x86_64/stable
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/eos-boot-helper/eos-migrate-chrome-profile
+Restart=no
+RemainAfterExit=yes
+
+[Install]
+WantedBy=gnome-session.target
```
wjt, line 15: Now that `eos-migrate-chrome.service` can run at any time – not just during boot – then it's not sufficient to run this at login.  You need to trigger this user service for any logged-in users at the time when flathub Chrome gets installed.

I think you probably want to do this with a `.path` unit. Something like this (untested!) which will trigger the corresponding `.service` in 2 cases:

- On login, if Flathub Chrome is installed
- Post-login, if Flathub Chrome becomes installed

```ini
# /lib/systemd/user/eos-migrate-chrome-profile.path
[Unit]
Description=...

[Path]
PathExists=/var/lib/flatpak/app/com.google.Chrome/x86_64/stable

[Install]
WantedBy=gnome-session.target
```

and then remove the `[Install]` section from this file:


```suggestion
```

wjt (changes requested):

eos-migrate-chrome-profile.service
starnight: > I think you probably want to do this with a `.path` unit. Something like this (untested!) which will trigger the corresponding `.service` in 2 cases:
> 
> * On login, if Flathub Chrome is installed

Installing `eos-migrate-chrome-profile.service` into `gnome-session.target` takes care of it.

> * Post-login, if Flathub Chrome becomes installed

Installing `eos-migrate-chrome-profile.path` into `gnome-session.target` takes care of it.  `eos-migrate-chrome-profile.path` keeps watch the path `/var/lib/flatpak/app/com.google.Chrome/x86_64/stable` is changed, or not.  If Flathub's Chrome becomes installed, path `/var/lib/flatpak/app/com.google.Chrome/x86_64/stable` becomes existed from not existed.  Then, it triggers `eos-migrate-chrome-profile.service`.

wjt (changes requested):

eos-migrate-chrome-profile.service
starnight: > I don't quite understand this. I think `PathExists` does what we want:
> 
> * When the `.path` unit starts, if the target path exists, the corresponding service is activated
> * Otherwise, if the target path is later created, the corresponding service is activated

If uses `PathExists=/var/lib/flatpak/app/com.google.Chrome/x86_64/stable` as the condition in `eos-migrate-chrome-profile.path` and install Flathub's Chrome,  `eos-migrate-chrome-profile.path` will match the condition multiple times, then triggers `eos-migrate-chrome-profile.service` multiple times:
```
× eos-migrate-chrome-profile.path - Monitor eos3 Chrome replacement and trigger profile migration
     Loaded: loaded (/usr/lib/systemd/user/eos-migrate-chrome-profile.path; disabled; preset: enabled)
     Active: failed (Result: trigger-limit-hit) since Wed 2025-09-24 18:10:53 CST; 2min 17s ago
   Duration: 5min 24.330s
   Triggers: ● eos-migrate-chrome-profile.service

Sep 24 18:05:28 endless systemd[1552]: Started eos-migrate-chrome-profile.path - Monitor eos3 Chrome replacement and trigger profile migration.
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.path: Trigger limit hit, refusing further activation.
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.path: Failed with result 'trigger-limit-hit'.

○ eos-migrate-chrome-profile.service - Migrate Chrome profile to new path
     Loaded: loaded (/usr/lib/systemd/user/eos-migrate-chrome-profile.service; static)
     Active: inactive (dead)
TriggeredBy: × eos-migrate-chrome-profile.path
  Condition: start condition unmet at Wed 2025-09-24 18:10:53 CST; 7min ago
             └─ ConditionPathExists=/sysroot/home/dev/.config/google-chrome was not met

Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.service - Migrate Chrome profile to new path was skipped because of an unmet condition check (ConditionPathExists=/sysroot/home/dev/.config/google-chrome).
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.service - Migrate Chrome profile to new path was skipped because of an unmet condition check (ConditionPathExists=/sysroot/home/dev/.config/google-chrome).
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.service - Migrate Chrome profile to new path was skipped because of an unmet condition check (ConditionPathExists=/sysroot/home/dev/.config/google-chrome).
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.service - Migrate Chrome profile to new path was skipped because of an unmet condition check (ConditionPathExists=/sysroot/home/dev/.config/google-chrome).
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.service - Migrate Chrome profile to new path was skipped because of an unmet condition check (ConditionPathExists=/sysroot/home/dev/.config/google-chrome).
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.service - Migrate Chrome profile to new path was skipped because of an unmet condition check (ConditionPathExists=/sysroot/home/dev/.config/google-chrome).
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.service - Migrate Chrome profile to new path was skipped because of an unmet condition check (ConditionPathExists=/sysroot/home/dev/.config/google-chrome).
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.service - Migrate Chrome profile to new path was skipped because of an unmet condition check (ConditionPathExists=/sysroot/home/dev/.config/google-chrome).
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.service - Migrate Chrome profile to new path was skipped because of an unmet condition check (ConditionPathExists=/sysroot/home/dev/.config/google-chrome).
Sep 24 18:10:53 endless systemd[1552]: eos-migrate-chrome-profile.service - Migrate Chrome profile to new path was skipped because of an unmet condition check (ConditionPathExists=/sysroot/home/dev/.config/google-chrome).
```

Well, triggering `eos-migrate-chrome-profile.service` multiple times might be okay, because it has the "condition check" as the guarding gate.  But, `eos-migrate-chrome-profile.path` hits `Trigger limit hit, refusing further activation.` after install Flathub's Chrome.

> We don't want to re-run the migration script every time Flathub Chrome is updated.

Same as `eos-migrate-chrome-profile.service` has the "condition check" as the guarding gate.

starnight: From [systemd.path](https://www.freedesktop.org/software/systemd/man/latest/systemd.path.html):
> When a service unit triggered by a path unit terminates (regardless whether it exited successfully or failed), monitored paths are checked immediately again, and the service accordingly restarted instantly. As protection against busy looping in this trigger/start cycle, a start rate limit is enforced on the service unit, see StartLimitIntervalSec= and StartLimitBurst= in [systemd.unit(5)](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#). Unlike other service failures, the error condition that the start rate limit is hit is propagated from the service unit to the path unit and causes the path unit to fail as well, thus ending the loop.

So, if it matches `PathExists=/var/lib/flatpak/app/com.google.Chrome/x86_64/stable`, it will trigger `eos-migrate-chrome-profile.service`.  And, because `eos-migrate-chrome-profile.service` is not activated due to mismatched condition, `eos-migrate-chrome-profile.service` keeps triggered by `eos-migrate-chrome-profile.path` until the limit.

eos-migrate-chrome-profile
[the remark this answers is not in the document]
starnight: I recorded a video to prove that the symlink is not necessary.  (phew, it is hard to do the emulation.)
[Screencast from 2025-09-25 16-54-59.webm](https://github.com/user-attachments/assets/2ad37ef4-4bd6-4d90-b791-1eed4ddf8f20)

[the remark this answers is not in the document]
wjt: I don't think this video proves much. By 2:31 you have shown that moving ~/.config/google-chrome while eos-apps Chrome is running doesn't crash it, but then you close the browser and move on. I am interested in what happens if you continue to use eos-apps Chrome after its config directory has been moved from underneath it. What happens if you try to make a new bookmark? Save a new password? Install a new theme? Change some settings? etc. Does ~/.config/google-chrome get recreated? Or does it save these to the new directory (which might happen if it has a dirfd for its config directory or files within it).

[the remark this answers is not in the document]
starnight: Okay, I tried to keep use eos-apps Chrome after the migration.  And, also modified eos-apps Chrome's settings.  The `~/.config/google-chrome` shows up again.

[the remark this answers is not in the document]
starnight: I think we can delete the symlink when eos-apps Chrome is not running by checking `ps -x | grep chrome-real`.  

But, as the idea "Personally I wouldn't delete the symlink. It's not hurting anyone." => Ya, that's true!  This makes things simpler.

eos-migrate-chrome-helper
```diff
@@ -0,0 +1,133 @@
...
+
+    if (
+        os.path.isdir(OLD_CHROME_CONFIG_DIR) and
+        not os.path.isdir(NEW_CHROME_CONFIG_DIR)
+    ):
+        update_old_config_references(
+            os.path.join(OLD_CHROME_CONFIG_DIR,
+                         "WidevineCdm", "latest-component-updated-widevine-cdm"))
+        os.makedirs(os.path.dirname(NEW_CHROME_CONFIG_DIR), exist_ok=True)
+        os.rename(OLD_CHROME_CONFIG_DIR, NEW_CHROME_CONFIG_DIR)
```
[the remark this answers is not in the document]
starnight: Updated the new implementation into the new commit.

eos-migrate-chrome
```diff
@@ -0,0 +1,18 @@
...
+  echo "eos-apps Chrome is not installed"
+  exit
+fi
+
+if ! [[ $(flatpak info -o com.google.Chrome//stable) == "flathub" ]]; then
+  echo "Install Flathub Chrome"
+  flatpak install flathub com.google.Chrome -y
+fi
+
+/usr/lib/eos-boot-helper/eos-migrate-chrome-helper --mimeapps
```
wjt, line 15: I only just noticed this. It's not correct to run this from the system service: the mimeapps settings are per-user.

I think you should remove the `--mimeapps` and `--config` flags for `eos-migrate-chrome-helper` and have it always do both.

wjt: Yes, of course; otherwise you'd change the default browser to a browser that's not installed.

starnight: ok, goes back to https://github.com/endlessm/eos-boot-helper/pull/422#discussion_r2378604189 and add the symlink as https://github.com/endlessm/eos-boot-helper/pull/422#discussion_r2378749403 simply.

Updated as the new commit.

PR #423, opened 2025-09-01
Adapt for use in eos7

ssssam:
This PR adapts eos-boot-helper to work with EOS7, based on GNOME OS. See commits for details, but the breaking changes are:

  * Deal with explicit `unsigned char` parameter in efivar (https://github.com/rhboot/efivar/commit/ba165d1482558005a16f4b44abaa1df5169727d1)
  * Enable systemd-resolved service
  * Remove some initramfs modules that don't exist in EOS7
  * Drop systemd preset states for services that don't exist in EOS7

When merged, this will effectively drop support for EOS6 in the 'master' branch.

### DIFF
--- a/50-eos.preset
+++ b/50-eos.preset
@@ -5,7 +5,6 @@
 # Reverse systemd defaults
 disable systemd-networkd.service
 disable systemd-networkd.socket
-disable systemd-resolved.service
 disable systemd-networkd-wait-online.service
 disable systemd-sysupdate.service
 disable systemd-sysupdate.timer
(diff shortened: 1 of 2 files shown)


ssssam:
You will probably want to either update eos6 to build from a new 'eos6' branch; or merge this to a new 'eos7' branch instead of 'master'.

wjt:
We use BusyBox in the initramfs not for size reasons but for licence reasons on PAYG systems.

ssssam:
> I'm a bit concerned that we'll set ourselves up for some traps if we swap out the shell and busybox, as we may encounter behavioural differences with these replacements, or may then encounter behavioural differences if we tweak other stuff and move back later.
> 
> But trying to be practical and keep things moving - I suggest seeing if dash can be added quickly at this stage, and doing that if so. I know busybox is a bit more painful, and maybe we push that til later.

I'll have a go today. It may not be that hard.

